PR #2, opened 2025-02-13
Bottom Tab with initial screens

CassioMG:
This PR creates a `folder structure` along with a initial rough version for the `Bottom Tab` with 4 tabs: `Home`, `Swap`, `History` and `Settings`. It also includes `Login` and `Sign out` buttons along with a network detector which displays a `No internet connection` banner.

This PR also ports the `Button` component from `SDS` to `react-native` along with its corresponding `Button/theme.ts` file also ported from `SDS`.

Also included in this PR:
- Rename project from `FreighterMobile` to `freighter-mobile` for naming consistency across other stellar projects
- Add `styled-components` for UI styling
- Add initial `redux store` and `reducer`
- Add initial `navigators` and `routes`
- Add `BaseLayout` component that should be used to wrap all screens
- Add a `dimensions` helper to automatically resize components, fonts and spacing based on device screen size
- Add a `debug` helper for better dev experience
- Port global `theme` values from SDS

https://github.com/user-attachments/assets/e57e8254-92b3-4bd9-b026-0146ed880a07

Screenshot 2025-02-12 at 17 20 38


src/ducks/networkInfo.ts
```diff
@@ -0,0 +1,25 @@
...
+const initialState: NetworkState = {
+  isConnected: true,
+  isInternetReachable: true,
+};
+
+const networkInfoSlice = createSlice({
+  name: 'networkInfo',
+  initialState,
+  reducers: {
+    setNetworkInfo: (state, action) => {
```
aristidesstaffieri, line 17: I'm curious about the reason for keeping the network connection in a store. It seems like every view needs a fresh subscription to the connection state.

If one view or component sets this in the store, would any other view get benefit from using this state before subscribing to the network info?

Does this introduce more of a possibility of showing the network connection warning from a stale previous network state?

src/config/theme.ts
```diff
@@ -0,0 +1,19 @@
+export const COLORS = {
```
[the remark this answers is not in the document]
CassioMG: great idea, I'll port over this theme file to the project

src/ducks/networkInfo.ts
CassioMG: I'm not sure I'm following your rationale completely but I'll highlight here how I thought it would work.

The `OfflineDetection` component is wrapping the `RootNavigator`, like this:
```
export const App = (): React.JSX.Element => {
  return (
    <Provider store={store}>
      <NavigationContainer>
        <OfflineDetection>
          <RootNavigator />
        </OfflineDetection>
      </NavigationContainer>
    </Provider>
  );
};
```

This means the `OfflineDetection` is the only component that would be subscribing to `NetInfo.addEventListener()` and also the only place calling the `setNetworkInfo` method.

Then any other component or screen in the app would have access to the `networkInfo` object and react to it accordingly. This could be useful in case we'd like to leave some portion of the app still functional even while in offline mode, but I also don't have a strong opinion on it! This was really just a first reducer sample to kick off the folder structure.

Does that answer your questions? Would you have some alternative suggestion?

Thanks!

aristidesstaffieri: ah I see, yeah I missed that the OfflineDetection wraps all of the other views. This makes sense, thanks!

src/config/store.ts
```diff
@@ -0,0 +1,31 @@
...
+});
+
+export type RootState = ReturnType<typeof store.getState>;
+export type Dispatch = typeof store.dispatch;
+export type Store = typeof store;
+
+// Use throughout your app instead of plain `useDispatch` and `useSelector`
+export const useDispatch = () => useReduxDispatch<Dispatch>();
+export const useSelector: TypedUseSelectorHook<RootState> = useReduxSelector;
+export const useStore = () => useReduxStore<Store>();
```
piyalbasu, line 31: Cassio and I talked about this briefly over DM, but we may want to consider sharing reducers between extension and mobile. It would be nice to be dealing with the the same actions and similarly shaped data when retrieving account info, for example

src/config/theme.ts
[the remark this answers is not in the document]
CassioMG: https://github.com/stellar/freighter-mobile/pull/2/commits/2ff34b20815a1295cf1d7bec8c89b585a59919d4

src/config/store.ts
CassioMG: Thanks for bringing this up. I think having similarly shaped data would be nice and could be helpful with code/logic understanding between the 2 products. But I'm also a bit worried that it could get in the way sometimes, since web and mobile have their differences as well. I'd say we could try mimicking each other whenever possible as a suggested approach but also have the flexibility to diverge from each other. Wdyt?

> when retrieving account info, for example

@piyalbasu would you mind linking here the extension actions for retrieving account info that you have in mind so we could take a look as an example?

Thanks!

CassioMG: I think we can keep discussing it offline and also during the next PRs, I'll merge this one now so we can keep moving! :)

aristidesstaffieri: yeah I'd lean to not do this so we don't inherit the store patterns we have in Freighter today, we've had trouble getting the desired UX out of the usage of our store in the past and are currently refactoring some big parts of the store away from it and into hooks(history/balances).

PR #5, opened 2025-02-27
Add SVG support

CassioMG:
This PR creates a `src/assets/icons` folder and adds `SVG support` so we can start rendering icons on Freighter mobile.

We can easily import them like `import ClipboardIcon from "assets/icons/clipboard.svg";`

This PR also moves the `src/config/sds/theme.ts` file to `src/config/theme.ts` for simplicity.

Screenshot 2025-02-27 at 10 42 54


Brunonascdev (approved):
LGTM!

PR #8, opened 2025-02-28
Add Link variant to Typography component

Brunonascdev:
Update the Text component to include an onPress handler that allows it to open URLs when clicked. 

The Linking API is utilized to check if the URL can be opened and to perform the action. 
Additionally, the test suite for the Text component has been updated to verify this new functionality.

https://github.com/user-attachments/assets/9e9b866b-3f55-4fb8-a4c0-83821665e652

[link-typography.webm](https://github.com/user-attachments/assets/d6740653-037d-4e7a-85c4-c3a404cc2eae)


src/components/sds/Typography/index.tsx
```diff
@@ -309,15 +315,29 @@ export const Text: React.FC<TextProps> = ({
...
-  </StyledText>
-);
+}) => {
+  const handlePress = () => {
+    if (url) {
+      Linking.canOpenURL(url).then((supported) => {
+        if (supported) {
+          Linking.openURL(url);
+        }
+      });
```
CassioMG, line 330: Since we don't have a tracking error tool installed it (like Sentry), could you please add a TODO here for us to catch and track the error once we have a tracking error mechanism in place? Wdyt?

Gemfile.lock
```diff
@@ -5,26 +5,23 @@ GEM
       base64
       nkf
       rexml
-    activesupport (7.2.2.1)
```
CassioMG, line 8: The Gemfile being changed seems a little displaced since this PR is not adding any new package

CassioMG: 🤔

CassioMG (approved):

PR #9, opened 2025-02-28
Add intl with i18next

Brunonascdev:
This PR introduces internationalization (i18n) using the i18next library, enabling multi-language support in the application.

Currently, translations are available for Portuguese and English, with the flexibility to add more languages in the future.

[intl.webm](https://github.com/user-attachments/assets/880a0680-237b-49f2-bd6d-07af7402d060)

https://github.com/user-attachments/assets/4f37085e-7b3d-4dcc-a88d-23c8be4444ec


CassioMG:
> I can see how it would be valuable for testing, but I don’t think we should maintain Portuguese translations unless we think it’s a product requirement. Right now we only need to have English. Can we have i18n support with only 1 language to start?

The Portuguese translations were added just for this initial testing, we don't plan on maintain them. We'll be using only English moving forward.

src/i18n/locales/en/translations.json
```diff
@@ -0,0 +1,25 @@
+{
+  "noInternetConnection": "No internet connection",
```
CassioMG, line 2: @Brunonascdev can we make the json keys be the english sentence?
E.g.:
"No internet connection": "No internet connection",

So that we don't need to worry about complex json structures now. I understand how this could be valuable as we could apply different translations based on the context of the phrase, but I believe this is something we could potentially tackle in the future. Wdyt?

src/helpers/testing.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { RenderAPI, render } from "@testing-library/react-native";
+import i18n from "i18n";
+import React from "react";
+import { I18nextProvider } from "react-i18next";
+
+type RenderWithProviderType = (component: React.ReactElement) => RenderAPI;
+
+jest.mock("helpers/get-os-language");
+
+export const renderWithProviders: RenderWithProviderType = (component) =>
```
CassioMG, line 11: how do we feel about renaming this file to `testUtils.tsx`?

src/helpers/get-os-language.ts
```diff
@@ -0,0 +1,25 @@
+import { I18nManager, Platform, Settings } from "react-native";
+
+function getOSLanguage(): string | undefined {
```
CassioMG, line 3: could we rename the file to `getOsLanguage` instead of `get-os-language` to keep the camelCase pattern?

CassioMG (approved):
Lgtm - left a few comments regarding file names

Brunonascdev:
> Lgtm - left a few comments regarding file names

Done!

Issue #10, opened 2025-03-04
[Onboarding] Create Icon component


CassioMG:
This task is about porting the [SDS Icon component](https://github.com/stellar/stellar-design-system/blob/main/%40stellar/design-system/src/icons.ts#L1176) along with its SVG files to `freighter-mobile` as a RN component and enhance it so that devs can render icons using an interface like `<Icon.IconName size={20} circle />`.

The `size` property would be `optional` and defaults to 24px.

The `circle` property would be `optional` and should add a rounded border around it.

It'll be used on the [Onboarding flow](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2218-7181&m=dev).

<img width="284" alt="Image" src="https://github.com/user-attachments/assets/28f36914-a22a-43f8-a02f-c202b40df06e" />

PR #26, opened 2025-03-06
Create Icon component

Brunonascdev:
Addresses issue #10 

This PR add a Icon component that can be used like this:

```typescript
<Icon.Activity size={32} color="#fff" />
```

The icons were imported from the [SDS](https://github.com/stellar/stellar-design-system/tree/main/%40stellar/design-system/src/assets/icons).

Samples:


src/components/sds/Icon/index.tsx
```diff
@@ -0,0 +1,2421 @@
...
+  name?: keyof typeof Icons;
+  circle?: boolean;
+  size?: number;
+  color?: string;
+  testID?: string;
+}
+
+const CircleContainer = styled.View`
+  width: ${({ size }: { size: number }) => size * 1.5}px;
+  height: ${({ size }: { size: number }) => size * 1.5}px;
```
CassioMG, line 2366: I think we should wrap those with `px()` instead of the inline px string so it resizes accordingly.

src/components/sds/Icon/index.tsx
```diff
@@ -0,0 +1,2421 @@
...
+  testID,
+}) => {
+  const IconSvg = name ? Icons[name] : null;
+
+  if (!IconSvg) {
+    return null;
+  }
+
+  const iconElement = (
+    <IconSvg width={size} height={size} color={color} testID={testID} />
```
CassioMG, line 2390: same thing here, I think we should do `pxValue(size)` to resize accordingly

Brunonascdev: my bad! I forgot about this function

CassioMG (approved):
Aside from the missing px() and pxValue() wrappers - LGTM!

Issue #15, opened 2025-03-04
[Onboarding] Create ChoosePassword UI


CassioMG:
*Depends on [[Onboarding] Create OnboardLayout component](https://github.com/stellar/freighter-mobile/issues/13)

This task is about creating just the UI for the [choose a password](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2232-21068&m=dev) screen which should be reused on "Create a new wallet" and "I already have a wallet" flows since on both flows users need to create a new password.

<img width="839" alt="Image" src="https://github.com/user-attachments/assets/95dd53a9-9f45-4a9b-bf1f-a78c20c231cf" />

The logic for actually handling/storing the password should be handled in a separate task.

Issue #14, opened 2025-03-04
[Onboarding] Create WelcomeScreen UI


CassioMG:
This task is about creating just the UI for the [welcome screen](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2232-21057&m=dev) which is the very first screen after the splash screen.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/379aa5da-b7e8-41cc-819a-65de868d52d3" />

The buttons behavior should be implemented in a different task.

Issue #13, opened 2025-03-04
[Onboarding] Create OnboardLayout component


CassioMG:
This task is intended to create a base `OnboardLayout` component which should be reused in all Onboarding screens.

We can check on [Figma designs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2218-7181&m=dev) (image below) that all Onboarding screens have the below consistent similarities:
- Background color (fixed)
- Screen padding (fixed)
- Avatar on top left corner with custom icon
- Display title with custom copy
- Continue/Submit button with custom title
- (optional) Footnote custom copy
- (optional) Back/Cancel button

<img width="1010" alt="Image" src="https://github.com/user-attachments/assets/36a08ee0-53dd-4287-bd49-f14608636132" />

The `OnboardLayout` component interface should comply to those UI properties. It should also implement the [KeyboardAvoidingView](https://reactnative.dev/docs/keyboardavoidingview) (or some other applicable solution) so that the content reacts accordingly when the keyboard shows up.

Suggested interface:

```
<OnboardLayout
     icon={<Icon.PasscodeLock />}
     title="Choose a Password"
     submitButtonTitle="Continue"
     onSubmit={goToNextScreen}
     footnote="We cannot recover your password for you. Make sure you keep this in a safe place."
     cancelButtonTitle="Go back"
     onCancel={goToPreviousScreen}
>
     {children}
<OnboardLayout />
```

PR #33, opened 2025-03-12
Create OnboardLayout and adapt route structure

Brunonascdev:
This PR addresses issue #13, #14 and #15 by introducing key enhancements to our onboarding flows and navigation structure:

- **OnboardLayout Component:**

A new OnboardLayout component has been added to streamline the onboarding screens. Since these screens share common components and design patterns, this layout ensures a consistent and maintainable UI across all onboarding flows.

- **Navigation Structure Update:**

The React Navigation routes have been restructured to prepare for an Authorized/Unauthorized state flow. Although the app currently lacks the logic to differentiate these states, a mocked variable is in place to simulate the behavior until the actual implementation is completed.

- **Welcome Screen:**

Added the Welcome Screen, with a basic logic to redirect to the next screens on the flow.

- **ChoosePassword Screen:**

Added the ChoosePassword Screen, with a basic logic to handle the password min. size (8) and next screens redirects.

The accompanying video illustrates the entire process:

1. The initial onboarding screen (Welcome screen) is displayed.
2. The next screen is the ChoosePassword, built using the new OnboardLayout component.
3. When the mocked authentication variable is updated, the app automatically redirects the user to the TabScreen, representing the authorized state.

https://github.com/user-attachments/assets/12ae3a73-7898-4e03-a2bd-d5327095a413


src/components/App.tsx
```diff
@@ -1,16 +1,25 @@
...
 
 export const App = (): React.JSX.Element => (
-  <I18nextProvider i18n={i18n}>
-    <NavigationContainer>
-      <OfflineDetection>
-        <RootNavigator />
-      </OfflineDetection>
-    </NavigationContainer>
-  </I18nextProvider>
+  <SafeAreaProvider>
```
CassioMG, line 12: I've read that using the `SafeAreaProvider` is a bit less performant than handling the safe area in a per-screen basis as _usually_ not all screens need to worry about safe area. But I agree that it simplifies a lot the implementation having it here. Maybe we could leave it like this and if we have some performance issues in the future we could revisit it. Since Freighter does not have a whole lot of screens maybe we will never notice the performance hit.

src/config/routes.ts
```diff
@@ -1,23 +1,41 @@
...
-  // Auth Stack
-  LOGIN: "Login",
-  MAIN_TABS: "MainTabs",
+export const ROOT_NAVIGATOR_ROUTES = {
+  AUTH_STACK: "AuthStack",
+  MAIN_TAB_STACK: "MainTabStack",
+} as const;
 
-  // Tab Stack
+export const MAIN_TAB_ROUTES = {
```
CassioMG, line 6: I think we could leave only `History, Home and Discovery` tabs as we have on Figma right now.

src/config/routes.ts
```diff
@@ -1,23 +1,41 @@
...
+export type AuthStackParamList = {
+  [AUTH_STACK_ROUTES.WELCOME_SCREEN]: undefined;
+  [AUTH_STACK_ROUTES.CHOOSE_PASSWORD_SCREEN]: undefined;
+  [AUTH_STACK_ROUTES.CONFIRM_PASSWORD_SCREEN]: undefined;
+  [AUTH_STACK_ROUTES.RECOVERY_PHRASE_ALERT_SCREEN]: undefined;
+  [AUTH_STACK_ROUTES.RECOVERY_PHRASE_SCREEN]: undefined;
+  [AUTH_STACK_ROUTES.IMPORT_WALLET_SCREEN]: undefined;
+};
+
 export type RootStackParamList = {
```
CassioMG, line 31: Could we move `RootStackParamList` above `AuthStackParamList` to keep the same sequence pattern of the _ROUTES consts?

CassioMG: Humm I think we could also move MAIN_TAB_ROUTES below AUTH_STACK_ROUTES, so we have it like this:

ROOT_NAVIGATOR_ROUTES

AUTH_STACK_ROUTES

MAIN_TAB_ROUTES

RootStackParamList

AuthStackParamList

MainTabStackParamList

src/components/sds/Icon/index.tsx
```diff
@@ -2388,11 +2390,21 @@ const IconComponent: React.FC<IconProps> = ({
...
+      color={color}
+      testID={testID}
+      onPress={onPress}
+    />
   );
 
   if (circle) {
-    return <CircleContainer size={size}>{iconElement}</CircleContainer>;
+    return (
+      <CircleContainer size={size} onPress={onPress}>
```
CassioMG, line 2404: Where would we be using a pressable `Icon`? Should we create an `IconButton` component to separate the concepts?

src/components/layout/OnboardLayout.tsx
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * @param {DefaultFooterProps} props - The component props.
+ * @param {() => void} [props.onPressDefaultActionButton] - Callback when the action button is pressed.
+ * @param {boolean} [props.isDefaultActionButtonDisabled] - Flag to disable the action button.
+ * @param {string} [props.defaultActionButtonText="Continue"] - Text to display on the action button.
+ */
+const DefaultFooter: React.FC<DefaultFooterProps> = ({
+  onPressDefaultActionButton,
+  isDefaultActionButtonDisabled,
+  defaultActionButtonText = "Continue",
```
CassioMG, line 106: should wrap with `t()` here

src/components/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+  };
+
+  return (
+    <BaseLayout useSafeArea>
+      <Container>
+        <StyledIconContainer>
+          <FreighterLogo width={px(32)} height={px(32)} />
+        </StyledIconContainer>
+        <View>
+          <StyledDisplay>Freighter Wallet</StyledDisplay>
```
CassioMG, line 70: Although it doesn't necessarily needs a translation, I think it could be useful to still wrap it with something like `t("freighterWallet")` so if we need to change the app name in the future we could simply change it in the translations files. Wdyt?

src/components/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <StyledSpacer $size={32} />
+          <Button
+            variant={ButtonVariants.TERTIARY}
+            size={ButtonSizes.LARGE}
```
CassioMG, line 74:
```suggestion
            tertiary
            lg
```

src/components/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+            onPress={handleCreateNewWallet}
+          >
+            {t("welcomeScreen.createNewWallet")}
+          </Button>
+          <StyledSpacer $size={12} />
+          <Button
+            variant={ButtonVariants.SECONDARY}
+            size={ButtonSizes.LARGE}
```
CassioMG, line 82:
```suggestion
            secondary
            lg
```

src/components/screens/WelcomeScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+            onPress={handleIAlreadyHaveWallet}
+          >
+            {t("welcomeScreen.iAlreadyHaveWallet")}
+          </Button>
+        </View>
+        <StyledTermsTextContainer>
+          <StyledTermsText md secondary weight="medium">
+            {t("welcomeScreen.terms.byProceeding")}
+            <Text md weight="medium" url="https://stellar.org/terms-of-service">
```
CassioMG, line 91:
```suggestion
            <Text md medium url="https://stellar.org/terms-of-service">
```

CassioMG (approved):
Great job! 👏 
Left some suggestions

src/components/sds/Icon/index.tsx
Brunonascdev: I added this prop because of the CustomHeader component, which has a ArrowLeft icon that can be pressed to return to the previous screen

Brunonascdev: I don't really see an issue with having a prop like this on the icon, but it's okay to change too

src/components/App.tsx
CassioMG: I'm not finding the link where I've read about it, but looking at the [official documentation](https://reactnavigation.org/docs/handling-safe-area/) they do recommend using `SafeAreaProvider` + `useSafeAreaInsets()` instead of `SafeAreaView` to avoid UI glitches so looks like it's the best approach :)

src/components/screens/WelcomeScreen.tsx
```diff
@@ -50,7 +50,7 @@ const StyledSpacer = styled.View<StyledProps>`
 `;
 
 export const WelcomeScreen: React.FC<WelcomeScreenProps> = ({ navigation }) => {
-  const { t } = useTranslation();
+  const { t } = useTranslation("translations");
```
CassioMG, line 53: do we need to pass a "translations" argument here?

CassioMG (approved):

src/components/screens/WelcomeScreen.tsx
Brunonascdev: not necessarily, but it helps to map the keys that we have on the translation files

src/components/App.tsx
Brunonascdev: yeah that's true, but in that case I'm wrapping the whole App with the **Provider** itself and **not the SafeAreaView**; 

I was using the `SafeAreaView` in the BaseLayout, but I'll change that too.

src/helpers/useAppTranslation.ts
```diff
@@ -0,0 +1,5 @@
+import { useTranslation as useTranslationOriginalHook } from "react-i18next";
+
+const useAppTranslation = () => useTranslationOriginalHook("translations");
+
+export default useAppTranslation;
```
CassioMG, line 5: thanks for making this change!

could we move it to a /hooks folder?

src/components/OfflineMessage.tsx
```diff
@@ -24,10 +32,11 @@ const Message = styled.Text`
 `;
 
 export const OfflineMessage = () => {
+  const insets = useSafeAreaInsets();
   const { t } = useAppTranslation();
 
   return (
-    <SafeContainer>
+    <SafeContainer $insets={insets}>
```
CassioMG, line 39: 🚀

src/components/layout/BaseLayout.tsx
```diff
@@ -11,27 +10,37 @@ interface BaseLayoutProps {
...
+  background-color: ${({ $backgroundColor }: StyledViewProps) =>
+    $backgroundColor};
 `;
 
 export const BaseLayout = ({
   children,
   useSafeArea = true,
   backgroundColor = THEME.colors.background.default,
 }: BaseLayoutProps) => {
+  const insets = useSafeAreaInsets();
```
CassioMG, line 38: 🚀

CassioMG (approved):

PR #37, opened 2025-03-14
Create ImportWallet screen

### LINKED ISSUE #18: [Onboarding] Create ImportWallet UI
*Depends on [[Onboarding] Create OnboardLayout component](https://github.com/stellar/freighter-mobile/issues/13)

This task is about creating the UI for the [import wallet](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2232-21088&m=dev) screen.

Image

The logic for actually importing the wallet should be handled in a separate task.

Brunonascdev:
This PR introduces the ImportWallet screen, which enables users to securely import their seed phrase and log into the app.

closes #18 

[importwallet.webm](https://github.com/user-attachments/assets/9760f3c8-e209-4f89-bcf0-fc01164bfb27)

https://github.com/user-attachments/assets/fa59f297-8188-4de6-adb4-71a3a4b2dae3

### DIFF
--- a/src/components/screens/ImportWalletScreen.tsx
+++ b/src/components/screens/ImportWalletScreen.tsx
@@ -1,0 +1,52 @@
+import Clipboard from "@react-native-clipboard/clipboard";
+import { NativeStackScreenProps } from "@react-navigation/native-stack";
+import { OnboardLayout } from "components/layout/OnboardLayout";
+import Icon from "components/sds/Icon";
+import { Textarea } from "components/sds/Textarea";
+import { AUTH_STACK_ROUTES, AuthStackParamList } from "config/routes";
+import useAppTranslation from "hooks/useAppTranslation";
+import React, { useState } from "react";
+
+type ImportWalletScreenProps = NativeStackScreenProps<
+  AuthStackParamList,
+  typeof AUTH_STACK_ROUTES.IMPORT_WALLET_SCREEN
+>;
+
+export const ImportWalletScreen: React.FC<ImportWalletScreenProps> = ({
+  navigation,
+}) => {
+  const [recoveryPhrase, setRecoveryPhrase] = useState("");
+  const { t } = useAppTranslation();
+
+  const handleContinue = () => {
+    // TODO: Navigate to the next screen after authentication is implemented
+    navigation.replace(AUTH_STACK_ROUTES.WELCOME_SCREEN);
+  };
+
+  const onPressPasteFromClipboard = async () => {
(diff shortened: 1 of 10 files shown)


src/components/screens/ImportWalletScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <OnboardLayout
+      icon={<Icon.Download01 circle />}
+      title={t("importWalletScreen.title")}
+      isDefaultActionButtonDisabled={!canContinue}
+      defaultActionButtonText={t("importWalletScreen.defaultActionButtonText")}
+      onPressDefaultActionButton={handleContinue}
+      hasClipboardButton
+      onPressClipboardButton={onPressPasteFromClipboard}
```
CassioMG, line 41: Do we have a good reason for adding the clipboard button inside the `OnboardLayout` component instead of handling it externally as a child together with Textarea? Like we are doing with the `RecoveryPhraseScreen` component

src/components/sds/Textarea/index.tsx
```diff
@@ -81,6 +81,7 @@ const StyledTextInput = styled.TextInput<Pick<StyledProps, "$fieldSize">>`
     px(TEXTAREA_SIZES[$fieldSize].paddingHorizontal)};
   padding-vertical: ${({ $fieldSize }: Pick<StyledProps, "$fieldSize">) =>
     px(TEXTAREA_SIZES[$fieldSize].paddingVertical)};
+  font-family: "Inter-Regular";
```
CassioMG, line 84: Nice catch!

But I believe we'll need to do something like the below here so we mimic what's being done in Typography for consistency:
```
  font-family: ${Platform.select({
    ios: "Inter-Variable",
    android: "Inter-Regular",
  })};
  font-weight: ${Platform.select({
    ios: "400",
    android: "normal",
  })};
```

The reason being `iOS` is using the `Inter-VariableFont_opsz,wght.ttf` file and `Android` is using the other font files.

I believe we'll actually need to apply this same fix for the `Input` component, would you mind double checking it? Thanks!

src/components/screens/ImportWalletScreen.tsx
Brunonascdev: It would require more changes on the screen and on the layout to make it work; I thought it was a good idea to keep there to keep it simple

CassioMG (approved):

Issue #28, opened 2025-03-08
[Home] View Balances Classic


CassioMG:
Implement the Tokens balance list below for `classic assets`.

~In this phase we can ignore the price (USD conversion) and percentage values as they will be added separately in a future task.~
We can also ignore all the other elements which are not highlighted in orange as they will be implemented in separate tasks.

We should use `freighter-backend`'s [`GET /account-balances`](https://github.com/stellar/freighter-backend/blob/main/src/route/index.ts#L287C15-L287C33) endpoint to fetch the balances.

We should create a reusable `BalancesList` component since the `Send` and `Swap` flows will also use the exact same list.

Figma: https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2547-36692&t=2HEx1KbdDRlwQbNh-1

<img width="368" alt="Image" src="https://github.com/user-attachments/assets/75e47f6a-6533-4f57-9bc2-1b041bcbfbaa" />

PR #38, opened 2025-03-17
Add `BalancesList` component functionality

CassioMG:
Related issue: https://github.com/stellar/freighter-mobile/issues/28

This PR adds a `BalancesList` component which is a `self-contained` component that displays a user's token balances in a scrollable list.

Please note that this PR is only about it's `FUNCTIONALITY`, the `actual DESIGN` will be implemented in a separate PR.

This `BalancesList` component should be used for `Home`, `Send` and `Swap` features. For the `Swap` feature we may need to add a _hidePercentagePriceChange_ flag since the current Design does not show the percentage label for swaps.

**The features in `BalancesList` component include:**
- **Displays regular tokens and liquidity pool tokens**
- **Shows** token balances with corresponding USD values
- **Displays 24h price changes with color indicators**
- **Supports pull-to-refresh to update balances and prices**
- **Shows loading, error, and empty states**
- **Auto-refreshes when the screen comes into focus**

**It can be rendered as easily as `<BalancesList />`.**

This PR also includes several `helpers`, `ducks` and a new `services` folder which contains an `API Factory` module for creating API service instances with consistent behavior.

This `apiFactory` module uses `axios` and provides a clean interface for making HTTP requests with features like:
 - Consistent error handling
 - Automatic retry with exponential backoff
 - Authorization token management

The first service using this factory is our `backend` API service which currently has the `fetchBalances` and `fetchTokenPrices` functions.

*** NOTE: the prices are currently [being mocked here](https://github.com/stellar/freighter-mobile/pull/38/files#diff-764fd5abbef6b6142d50d01205d9a2ec9b1e97b850607c0563eef4081d787d04R126-R139) with random values each time it's fetched. That's why you'll see different values on each refresh on the videos (which also makes it easier to detect the changes). We can start hitting the real endpoint once it's deployed to `/api/v1/token-prices`.

https://github.com/user-attachments/assets/4de192d9-8740-4cf4-8109-cfb0b8382020

https://github.com/user-attachments/assets/66b2abeb-1151-449f-b317-c739b369eb44


src/config/types.ts
```diff
@@ -0,0 +1,102 @@
...
+  issuer: Issuer;
+  anchorAsset?: string;
+  numAccounts?: BigNumber;
+  amount?: BigNumber;
+  bidCount?: BigNumber;
+  askCount?: BigNumber;
+  spread?: BigNumber;
+};
+
+export type BaseBalance = {
```
aristidesstaffieri, line 28: I happen to be re-doing these types in the extension now and would suggest to go with something that more accurately represents the keys across the distinct balance types, like a union of the different balance types. For example the `BaseBalance` here has keys like `contractId` but in reality that key only exists in `SorobanBalances` which makes the consumers of these types less aware of which balance type they are using.

Example - https://github.com/stellar/freighter/blob/de5ae6df911b3a0fe857eddedd3278c928f0e620/%40shared/api/types/account-balance.ts

src/components/BalancesList.tsx
```diff
@@ -0,0 +1,295 @@
...
+          <IconPlaceholder>
+            <Text md>{firstChar}</Text>
+          </IconPlaceholder>
+          <AssetTextContainer>
+            <Text md>{assetCode}</Text>
+            <AmountText sm>{formatAssetAmount(item.total)}</AmountText>
+          </AssetTextContainer>
+        </LeftSection>
+        <RightSection>
+          <Text md>{currentPrice ? formatFiatAmount(fiatValue) : "—"}</Text>
```
aristidesstaffieri, line 269: I don't think this is likely but in theory balances can be large and the number could be long enough to overflow the container here. Should this include some styling for values that are too long like the truncation/ellipsis?

Brunonascdev (approved):

src/components/BalancesList.tsx
CassioMG: Nice catch. So I was concerned about this too since it's easy to have long numbers with crypto. I was planning on trying some font resizing logic on my next PR where I tackle the styling and see how it goes. The truncation could also be an option.

How did you think about it? E.g. Would we display something like : "1.1234...." and then the user would have to tap on it to see the full number on the balance details screen?

src/config/types.ts
CassioMG: @aristidesstaffieri I see that you are using the suffix "_Asset" for the balance types, I'd suggest using "_Balance" instead. What do you think?

Personally, the word "Asset" makes me think more about just the "Token" than the "Balance". The word "Balance" implies that there is a token/asset and an amount, but the word "Asset" doesn't necessarily imply the same thing (at least on my head haha).

CassioMG: About the Liquidity Pool interface:
export interface LiquidityPoolShareAsset {
  available: BigNumber;
  total: BigNumber;
  liquidityPoolId: string;
  reserves: Horizon.HorizonApi.Reserve[];
  limit: string;
}

From my tests, it appear it doesn't have a `available` amount, but just a `total` amount. E.g. when I fetch balances for this Testnet account `GBNMQBDE2BPGG7QMNZTKA5VMKMSUNBQMMADANNMPS6VNRUYIVAU5TJRQ` it returns something like:
```
{
      total: "1472.6043561",
      limit: "100000",
      liquidityPoolId:
        "4ac86c65b9f7b175ae0493da0d36cc5bc88b72677ca69fce8fe374233983d8e7",
      reserves: [
        {
          asset: "native",
          amount: "5061.4450626",
        },
        {
          asset:
            "USDC:GBUNQWSNHUCOCUDRESGNY5SIS2CXILTWHZV5VARUP47G44NRUOOEYICX",
          amount: "44166.9752644",
        },
      ],
    };
```

src/components/BalancesList.tsx
aristidesstaffieri: yeah so today we actually just truncate it without any in line option to see the whole thing but I think copying the value or expanding it in some way could work. I also think another option could be to just leave it truncated and make sure the whole value can be viewed in the balance detail view when you click in.

src/config/types.ts
CassioMG: I've updated the balances typing here so it looks similar to what you have in your refactor: https://github.com/stellar/freighter-mobile/pull/38/commits/d5af607da3e7d4716608d7f1032596b2133add6b

aristidesstaffieri: ah nice catch I'll remove that key from our instance of the `LiquidityPoolShareAsset`.

aristidesstaffieri (approved):

Brunonascdev (approved):

Issue #20, opened 2025-03-04
[Onboarding] Implement "I already have a wallet" flow


CassioMG:
This task is about creating the business logic for the [I already have a wallet](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2232-21057&m=dev) flow. The actual UI components will be handled in separate tasks.

Included in this task:
- Navigation between all screens
- Handle password creation
- Handle importing accounts from a mnemonic phrase
- Error handling
- Handle onboarding flow states (skipped steps, resuming process)

CassioMG:
@Brunonascdev working well! The only thing I noticed is that when you switch accounts the balances from the previous account blinks for a second before it loads the balances from the current account.

PR #41, opened 2025-03-22
Create Wallet Flow

### LINKED ISSUE #19: [Onboarding] Implement "Create a new wallet" flow
This task is about creating the business logic for the [create a new wallet](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2232-21057&m=dev) flow. The actual UI components will be handled in separate tasks.

Included in this task:
- Navigation between all screens
- Handle password creation
- Handle mnemonic phrase generation
- Handle multiple account derivation from same mnemonic phrase
- Handle deriving distinct key pairs for `Testnet` and `Mainnet` from the same mnemonic phrase
- Setup a secure storage for password, mnemonic phrase and secret keys
- Error handling
- Handle onboarding flow states (skipped steps, resuming process)

Brunonascdev:
This PR implements the logic for generating a secure keypair (wallet) and storing it safely on the user’s device using the [React Native Keychain](https://github.com/oblador/react-native-keychain) library.

closes #19.

- Future Enhancements:
- This PR focuses exclusively on keypair creation and secure storage.
- The logic for handling hash key expiration and login flows will be introduced in the subsequent PR addressing issue #20.

The general auth flow implemented here follows the same standard used on the extension:

- Avoid Storing Plain Text Passwords in Memory
  - Instead of keeping the raw password, generate a hashed version (referred to as hashKey) using random salts, ensuring it cannot be reverse-engineered.
- hashKey Lifecycle
  - The hashKey remains active for 24 hours.
  - After expiration or when re-authentication occurs, a new hashKey is generated from the password and stored in memory. (TBD)
- Encrypt Sensitive User Data
  - Upon authentication, encrypt the user’s private keys and mnemonic phrase using the generated hashKey.
  - This step ensures that the sensitive data is protected and can only be decrypted using the unique hashKey.
- Secure Storage with React Native Keychain
  - Store the hashKey-encrypted private keys and mnemonic phrase securely using React Native Keychain instead of browser storage.
  - This separation keeps the decryption key (hashKey in memory) and the encrypted data (stored via Keychain) in different secure locations.
- Decryption Process on Demand
  - When the user initiates an action requiring access to their private keys or mnemonic phrase, decrypt the information directly from the secure Keychain store using the in-memory hashKey, without requiring the password again.
- Post-Expiration Handling
  - After 24 hours, clear both the hashKey and the data stored in the secure Keychain (temporary storage).
  - The KeyManager’s encrypted blobs remain intact, allowing for data recovery upon subsequent re-authentication.
- Re-authentication Workflow
  - When the user re-authenticates using their password, use the password to decrypt the KeyManager blobs.
  - Generate a new unique hashKey from the password, re-encrypt the private keys and mnemonic phrase, and store them again securely using React Native Keychain.

[create-wallet.webm](https://github.com/user-attachments/assets/6c07f872-0e17-4d16-a253-935b1468cee6)

https://github.com/user-attachments/assets/a515f813-0068-40cc-a92a-0106b3cf7a94


ios/freighter-mobile.xcodeproj/project.pbxproj
```diff
@@ -315,6 +315,7 @@
 					"-ObjC",
 					"-lc++",
 				);
+				OTHER_SWIFT_FLAGS = "$(inherited) -D EXPO_CONFIGURATION_DEBUG";
```
CassioMG, line 318: is this a leftover from testing the expo package?

src/config/logger.ts
```diff
@@ -0,0 +1,130 @@
...
+  error: (message: string, ...args: unknown[]) => void;
+}
+
+/**
+ * Default console adapter implementation
+ */
+const consoleAdapter: LoggerAdapter = {
+  debug: (message: string, ...args: unknown[]) => {
+    if (__DEV__) {
+      console.debug(`[DEBUG] ${message}`, ...args);
```
CassioMG, line 28: maybe we could use our existing `debug()` helper to pretty print debug logs here with timestamps, what do you think?

src/config/types.ts
```diff
@@ -1,6 +1,29 @@
 import { AssetType, Horizon } from "@stellar/stellar-sdk";
 import BigNumber from "bignumber.js";
 
+export type Account = {
+  id: string;
```
CassioMG, line 5: should we add a `network` prop here?

src/services/storage/reactNativeKeychainStorage.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Default service name used for the keychain
+ */
+const DEFAULT_SERVICE = "freighter_secure_storage";
+
+/**
+ * Sanitizes a key to be compatible with keychain
+ * Ensures that keys are valid for use with react-native-keychain
+ */
+function sanitizeKey(key: string): string {
```
CassioMG, line 14: I'm a little concerned on sanitizing keys inside the getItem and setItem functions as it would be changing the original string input without the caller realizing it. Would it be safer (or less bug prone) to instead throw an error here and have the caller sanitize the key if they need to? What do you think?

src/ducks/auth.ts
```diff
@@ -0,0 +1,355 @@
...
+  activeKeyPair: null,
+};
+
+// Constants
+const HASH_KEY_EXPIRATION_MS = 24 * 60 * 60 * 1000; // 24 hours
+
+/**
+ * Key manager instance for handling cryptographic operations
+ */
+const keyManager = createKeyManager(Networks.TESTNET);
```
CassioMG, line 74: Could we add a comment here explaining that the Testnet network will be used only as the default network passphrase but that we can actually use this same keyManager instance for any other network?

src/hooks/useGetActiveAccount.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    if (!timestampStr) {
+      return false;
+    }
+
+    const timestamp = parseInt(timestampStr, 10);
+    const isValid = Date.now() < timestamp;
+
+    return isValid;
+  } catch (error) {
```
CassioMG, line 141: should we log the error here?

src/hooks/useGetActiveAccount.ts
```diff
@@ -0,0 +1,243 @@
...
+interface TemporaryStore {
+  expiration: number;
+  privateKeys: Record<string, string>;
+  mnemonicPhrase: string;
+}
+
+/**
+ * Retrieves data from the temporary store
+ */
+const getFromTemporaryStore = async (): Promise<TemporaryStore | null> => {
```
CassioMG, line 25: should we rename it to `getDataFromTemporaryStore` ?

CassioMG: or maybe `getTemporaryStore` if it looks better :)

src/helpers/keyManager/ReactNativeKeychainFacade.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  private accessGroup?: string;
+
+  private securityLevel?: Keychain.SECURITY_LEVEL;
+
+  private indexKey: string;
+
+  constructor() {
+    this.service = DEFAULT_SERVICE;
+    this.indexKey = sanitizeKey("stellarkeys_index");
```
CassioMG, line 46: is `indexKey` this being used somewhere?

src/helpers/keyManager/ReactNativeKeychainFacade.ts
```diff
@@ -0,0 +1,235 @@
...
+
+      let keyIds: string[] = [];
+
+      if (result !== false) {
+        keyIds = JSON.parse(result.password);
+      }
+
+      if (!keyIds.includes(id)) {
+        keyIds.push(id);
+        await Keychain.setGenericPassword("index", JSON.stringify(keyIds), {
```
CassioMG, line 200: just double checking, is this hardcoded "index" string here correct?

src/helpers/keyManager/ReactNativeKeychainFacade.ts
```diff
@@ -0,0 +1,235 @@
...
+    try {
+      const result = await Keychain.getGenericPassword({
+        service: `${this.service}_index`,
+      });
+
+      if (result !== false) {
+        const keyIds: string[] = JSON.parse(result.password);
+        const newKeyIds = keyIds.filter((keyId) => keyId !== id);
+
+        await Keychain.setGenericPassword("index", JSON.stringify(newKeyIds), {
```
CassioMG, line 226: looks like we are using this "index" username in a few places, should we move it to a const?

src/components/screens/ConfirmPasswordScreen.tsx
```diff
@@ -20,13 +20,24 @@ export const ConfirmPasswordScreen: React.FC<ConfirmPasswordScreenProps> = ({
   const [confirmPasswordValue, setConfirmPasswordValue] = useState("");
   const { t } = useAppTranslation();
 
-  const handleContinue = () => {
-    navigation.navigate(AUTH_STACK_ROUTES.RECOVERY_PHRASE_ALERT_SCREEN);
-  };
+  const canContinue = useMemo(
+    () =>
+      confirmPasswordValue.length >= PASSWORD_MIN_LENGTH &&
```
piyalbasu, line 25: A future improvement might be to consider using a lib for form validation. I think this is fine for now, but it can get a little cumbersome when you have a lot of different form fields with different validation rules (which will likely happen in the future as you add more screens)

FWIW, we use [formik](https://formik.org/) on the extension side

src/navigators/RootNavigator.tsx
```diff
@@ -1,31 +1,71 @@
...
 
   useEffect(() => {
-    // We can bypass the eslint rule here because we need to hide the splash screen
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    RNBootSplash.hide({ fade: true });
-  }, []);
+    const validateAuth = async () => {
+      try {
+        // First check if hash key is valid - this might authenticate the user
+        await isHashKeyValid();
```
CassioMG, line 24: should we have an `if()` statement here? Like `if(await isHashKeyValid())`

ios/freighter-mobile.xcodeproj/project.pbxproj
Brunonascdev: nice catch!

src/navigators/RootNavigator.tsx
```diff
@@ -1,31 +1,71 @@
...
+
+        // Then get auth status
+        getIsAuthenticated();
```
CassioMG, line 27: should we wait for this `getIsAuthenticated()` to finish loading before proceeding to next steps?

src/navigators/RootNavigator.tsx
```diff
@@ -1,31 +1,71 @@
...
+      return ROOT_NAVIGATOR_ROUTES.MAIN_TAB_STACK;
+    }
+
+    // If we have an account but not authenticated, show login screen
+    if (hasAccount) {
+      return ROOT_NAVIGATOR_ROUTES.AUTH_STACK;
+    }
+
+    // Otherwise show the auth stack (welcome, signup, etc.)
+    return ROOT_NAVIGATOR_ROUTES.AUTH_STACK;
```
CassioMG, line 63: nit: they both are returning the same thing, in case you think it'd make sense to simplify this a little bit

src/navigators/RootNavigator.tsx
CassioMG: otherwise in case it takes a while to process it _looks like_ we could have a glitch in the UI transitioning from the welcome screen to the main tabs screen. Does that make sense?

src/navigators/RootNavigator.tsx
Brunonascdev: it's returning the same thing for now only; on the next PR I'll improve and add a proper login screen

src/helpers/keyManager/ReactNativeKeychainFacade.ts
Brunonascdev: no, just removed

src/helpers/keyManager/ReactNativeKeychainFacade.ts
Brunonascdev: yes, this is the key that's I'm using for storing all the keys. it's necessary since the React Native Keychain lib doesn't have a native getAllKeys method.

Brunonascdev: I moved to a new env though

src/components/screens/ConfirmPasswordScreen.tsx
Brunonascdev: Hey, apparently they're not maintaining formik anymore: https://github.com/jaredpalmer/formik/issues/3601

Should we move to alternate solutions, like [react-hook-form](https://github.com/react-hook-form/react-hook-form)?

src/navigators/RootNavigator.tsx
Brunonascdev: the logic is already being handled on the useAuthStore (using .then and .catch)

CassioMG (approved):
LGTM

src/components/screens/ConfirmPasswordScreen.tsx
piyalbasu: Oh wow - I didn't realize that. Yeah, react-hook-form looks like a good alternative! I should move the extension off of formik too

PR #47, opened 2025-03-27
View Balances Classic UI

CassioMG:


src/navigators/RootNavigator.tsx
danilosilvackl, line 1: PR #45  is already fixing this, i think you will need to revert these changes before merging

aristidesstaffieri:
One difference between the sorting here and the sorting in the extension is that we always keep xlm at the top regardless of value, should we do the same here?

src/navigators/RootNavigator.tsx
CassioMG: That's correct, I'll update it - thanks!

CassioMG:
> One difference between the sorting here and the sorting in the extension is that we always keep xlm at the top regardless of value, should we do the same here?

@aristidesstaffieri I initially copied that behavior, but then I realized it's different on Figma designs. It's always sorting using the USD amount regardless if it's XLM or not. @sdfcharles was that Figma change intentional? I can quickly update to always display XLM on top if that's our goal - thanks!

sdfcharles:
@CassioMG nope that change is correct, we shouldnt pin XLM at the top, just default sort from highest value yo lowest. if anything that would be a separate feature request.

aristidesstaffieri (approved):

src/navigators/RootNavigator.tsx
CassioMG: https://github.com/stellar/freighter-mobile/pull/47/commits/476d1dd60fd03051fda46f6e4313b552780d8bd9

PR #51, opened 2025-03-28
Create Validate Recovery Phrase Screen

### LINKED ISSUE #46: [Onboarding] Add recovery phrase validation
This task is about creating the new screen for the validation of the recovery phrases following the [design specs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2588-36554&p=f&t=0kpWaHB0VWaTEvOE-0)

## Implementation Details ⚙️
- The new screen should validate 3 randomly chosen words
- can't repeat the words already validated
- show error for the current phrase
- after all 3 are validated, redirect the user to the home screen

## Phase specific 🔍
- Only the input will be used now
- the word bubbles bellow will be implemented in the future

## Screenshots 📸
Image

danilosilvackl:
This PR introduces a new screen after the user saves his recovery words.
closes #46 

## How it works:
- It will split the recovery phrase into words
- select 3 of those
- after the user validates all 3 of them
- redirect the user to the home screen

## iOS :
https://github.com/user-attachments/assets/97563259-d898-49c4-be3d-540a6db9b15d

## Android 🤖:
[android.webm](https://github.com/user-attachments/assets/1d02e20e-36cf-4ff0-a349-6d9e32b0205f)


src/components/screens/ValidateRecoveryPhraseScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }, [canContinue, currentIndex, password, recoveryPhrase, signUp, t]);
+
+  const handleWordChange = useCallback((value: string) => {
+    setCurrentWord(value.trim());
+    setError(undefined);
+  }, []);
+
+  return (
+    <OnboardLayout
+      icon={<Icon.PasscodeLock circle />}
```
CassioMG, line 74: I think it's using a slightly different icon for this screen, it appears to be `Icon.Passcode`

src/hooks/useWordSelection.ts
```diff
@@ -0,0 +1,16 @@
+import { useMemo, useState } from "react";
+
+export const useWordSelection = (recoveryPhrase: string) => {
+  const words = useMemo(() => recoveryPhrase.split(" "), [recoveryPhrase]);
+
+  const [selectedIndices] = useState(() => {
+    const indices = Array.from({ length: words.length }, (_, i) => i);
```
CassioMG, line 7:
```suggestion
  const [selectedIndexes] = useState(() => {
    const indexes = Array.from({ length: words.length }, (_, i) => i);
```

src/components/screens/ValidateRecoveryPhraseScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+const DELAY_MS = 500;
+
+export const ValidateRecoveryPhraseScreen: React.FC<
+  ValidateRecoveryPhraseScreenProps
+> = ({ route }) => {
+  const { password, recoveryPhrase } = route.params;
+  const [currentWord, setCurrentWord] = useState("");
+  const [error, setError] = useState<string | undefined>();
+  const [isLoading, setIsLoading] = useState(false);
+  const [isSigningUp, setIsSigningUp] = useState(false);
```
CassioMG, line 26: have you tried using our auth store to track "isSigningUp" state like the below so we simplify the code a little bit?
`const isSigningUp = useAuthenticationStore(state => state.isLoading);`
The `signUp` call should immediately set that loading state.

src/components/screens/ValidateRecoveryPhraseScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+        setIsSigningUp(true);
+        InteractionManager.runAfterInteractions(() => {
+          signUp({
+            password,
+            mnemonicPhrase: recoveryPhrase,
+          });
+        });
+      }
+      setIsLoading(false);
+    }, DELAY_MS);
```
CassioMG, line 64: are we adding this timeout to  improve the UX a little bit? Like giving the user a sense of "data is being processed"
if so, should we also add this delay for the error case to keep the UX consistent?

src/hooks/useWordSelection.ts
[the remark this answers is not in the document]
CassioMG: this is great, I think we should always try to make the screens dumb components and extract the logic to somewhere else

CassioMG (approved):
Lgtm - left a few suggestions

src/components/screens/ValidateRecoveryPhraseScreen.tsx
danilosilvackl: OH, you're right, updated, thanks!

src/hooks/useWordSelection.ts
danilosilvackl: i agree, thanks for the suggestion, fixed!

src/components/screens/ValidateRecoveryPhraseScreen.tsx
danilosilvackl: i did not try that until this comment, thanks for the heads up! done!

src/components/screens/ValidateRecoveryPhraseScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }, [canContinue, currentIndex, password, recoveryPhrase, signUp, t]);
+
+  const handleWordChange = useCallback((value: string) => {
```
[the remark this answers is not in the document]
danilosilvackl: done!

Brunonascdev (approved):

PR #74, opened 2025-03-31
Add Friendbot button

CassioMG:
Closes https://github.com/stellar/freighter-mobile/issues/69

This PR adds the `FriendbotButton` component which should only be shown and used on `Testnet` (or Futurenet once we add support for it) to fund recently created accounts with `10000 XLM`.

This PR also fixes a small UI glitch that was causing the empty state to briefly blink for a fraction of second before displaying the balances spinner.

Screenshot 2025-03-31 at 16 33 01

### iOS
https://github.com/user-attachments/assets/d2b8ce0e-1fbb-4c77-8dfc-77016f485388

### Android
[android-friendbot.webm](https://github.com/user-attachments/assets/c9d5061c-4b6f-469f-9802-d758e8e9e920)

### DIFF
--- a/src/components/BalancesList.tsx
+++ b/src/components/BalancesList.tsx
@@ -1,4 +1,5 @@
 import { AssetIcon } from "components/AssetIcon";
+import { FriendbotButton } from "components/FriendbotButton";
 import { Text } from "components/sds/Typography";
 import { NETWORKS } from "config/constants";
 import { THEME } from "config/theme";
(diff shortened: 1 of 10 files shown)


src/components/BalancesList.tsx
```diff
@@ -106,6 +108,22 @@ export const BalancesList: React.FC<BalancesListProps> = ({
     fetchAccountBalances,
   } = useBalancesStore();
 
+  const noBalances = Object.keys(pricedBalances).length === 0;
+
+  const isTestNetwork = [NETWORKS.TESTNET, NETWORKS.FUTURENET].includes(
+    network,
+  );
+
+  // Set isMounting to false after the component mounts.
```
aristidesstaffieri, line 117: maybe you could use a lifecycle state machine like we are starting to do in the extension, something like `IDLE, PENDING, SUCCESS, ERROR`.  This could help stay away from having to use view sentinels values to step around the state.

aristidesstaffieri (approved):

Brunonascdev (approved):

PR #75, opened 2025-04-01
Import Wallet Flow

Brunonascdev:
This PR includes:

- Import Wallet flow implementation
- LockScreen UI with session timeout logic
- Auth store methods for managing auth state
- `useGetActiveAccount` hook for reliable account access
- Updated documentation for new features

closes issue #20 

# iOS Samples
## Create new wallet - verifying seed phrase
https://github.com/user-attachments/assets/38c6d1ed-77a4-41ca-9e43-e564d64023fe
## Create new wallet - skipping seed phrase verification
https://github.com/user-attachments/assets/f217b9a8-15c8-4617-a153-ad1c37cb0917
## Importing new wallet
https://github.com/user-attachments/assets/275f672f-904a-4e7d-b00b-cdaeee3acd04
## Session timeout
https://github.com/user-attachments/assets/11ebf815-2e8b-4a39-99d1-c568351fbbd7

# Android Samples
## Create new wallet - verifying seed phrase
https://github.com/user-attachments/assets/a07259af-1cd5-4959-aba4-66f7175c46a2
## Create new wallet - skipping seed phrase verification
https://github.com/user-attachments/assets/3b204c3e-6410-4104-a566-3cf6e56c7e4a
## Importing new wallet
https://github.com/user-attachments/assets/a8c88189-17f3-4def-8575-d641421a3402
## Session timeout
https://github.com/user-attachments/assets/103d182f-771b-4ef5-b67f-68602f170a60


src/components/screens/HistoryScreen.tsx
```diff
@@ -45,36 +76,21 @@ export const HistoryScreen = () => {
...
-        )}
-
         <ButtonContainer>
+          <Text>{account?.publicKey}</Text>
+
           <Button isFullWidth onPress={handleLogout}>
             <Text>Logout</Text>
           </Button>
+
+          <Button
```
danilosilvackl, line 86: do we want this button to visible at all times? it looks like a thing for testnet/futurenet

src/components/screens/LockScreen.tsx
```diff
@@ -0,0 +1,155 @@
...
+              onPress={handleUnlock}
+              disabled={!canContinue}
+              isLoading={isSigningIn}
+            >
+              {t("lockScreen.unlockButtonText")}
+            </Button>
+          </StyledInputContainer>
+        </StyledFormContainer>
+        <ForgotPasswordContainer>
+          <Button secondary lg onPress={() => {}}>
```
CassioMG, line 148: @Brunonascdev once this "Forgot Password" action is implemented will this take the user back to the Welcome Screen so that s/he can either import the wallet again using seed phrase or create a new one?

Is the user currently able to get back to the Welcome Screen if they logout and delete the app?

Asking just to understand how Teaser testers could get back to the Welcome Screen

Thanks!

src/hooks/useAuthCheck.ts
```diff
@@ -0,0 +1,209 @@
...
+  const lastInteractionRef = useRef(Date.now());
+  const lastCheckRef = useRef(Date.now());
+  const panResponderRef = useRef<PanResponderInstance | null>(null);
+
+  // Check auth status based on current state (background, foreground, active)
+  const checkAuth = useCallback(async () => {
+    try {
+      const now = Date.now();
+      // If it's been too soon since last check, skip
+      if (now - lastCheckRef.current < 1000) {
```
danilosilvackl, line 36: i think you could put the ms value for this check into a const too something like 
`const MIN_CHECK_INTERVAL = 1000;`

src/hooks/useAuthCheck.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      // Don't perform auth checks too frequently while on lock screen
+      if (authStatus === AUTH_STATUS.HASH_KEY_EXPIRED) {
+        const skipChecksWhenExpired = true;
+        if (skipChecksWhenExpired) {
+          // Skip frequent checks when already in expired state
+          return;
+        }
+      }
```
danilosilvackl, line 47: this could be simplified to this:
```suggestion
if (authStatus === AUTH_STATUS.HASH_KEY_EXPIRED) {
  return;
}
```

src/hooks/useAuthCheck.ts
```diff
@@ -0,0 +1,209 @@
...
+      if (
+        appState.current.match(/inactive|background/) &&
+        nextAppState === "active"
+      ) {
+        // App has come to the foreground - check auth with a delay to allow app to fully resume
+        setTimeout(() => {
+          checkAuth().catch((err) =>
+            logger.error("handleAppStateChange", "Error checking auth", err),
+          );
+        }, 300);
```
danilosilvackl, line 112: i feel this delay could be in a const as well to make it easier to manipulate since you are using it on line :127 as well

src/hooks/useAuthCheck.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const interactionCheck = setInterval(() => {
+      const now = Date.now();
+      // If user hasn't interacted in 30 seconds, consider them "inactive"
+      if (now - lastInteractionRef.current > 30000) {
+        setIsActive(false);
+      } else {
+        setIsActive(true);
+      }
+    }, 5000);
```
danilosilvackl, line 152: you can use the `ACTIVE_CHECK_INTERVAL` here

src/hooks/useAuthCheck.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    return () => {
+      InteractionManager.clearInteractionHandle(interactionStartListener);
+      clearInterval(interactionCheck);
+    };
+  }, []);
+
+  // Initialize PanResponder to track user interactions
+  useEffect(() => {
+    const updateLastInteraction = () => {
```
danilosilvackl, line 162: you could extract this function and use it on line :141 too

src/hooks/useAuthCheck.ts
```diff
@@ -0,0 +1,209 @@
...
+      subscription.remove();
+      if (checkIntervalRef.current) {
+        clearInterval(checkIntervalRef.current);
+      }
+    };
+  }, [setupCheckInterval, checkAuth]);
+
+  // Track user interaction
+  useEffect(() => {
+    const interactionStartListener =
```
CassioMG, line 139: Is this `interactionStartListener` being effectively used? It appears this is being created here and removed on the useEffect clean-up function but I don't see any other reference to it

danilosilvackl (approved):

src/components/layout/BaseLayout.tsx
```diff
@@ -1,20 +1,63 @@
...
+        flex: 1,
+        backgroundColor: THEME.colors.background.default,
+      },
+      ...props,
+    }) as ScrollViewProps,
+)`
+  flex-grow: 1;
+  background-color: ${THEME.colors.background.default};
+`;
+
```
CassioMG, line 60: It appears this is a copy/paste of what we have on `OnboardLayout`. Could we extract it to a reusable component?

src/config/constants.ts
```diff
@@ -2,8 +2,10 @@
 import { Networks } from "@stellar/stellar-sdk";
 
 const PASSWORD_MIN_LENGTH = 8;
+const PASSWORD_MAX_LENGTH = 32;
+const HASH_KEY_EXPIRATION_MS = 48 * 60 * 60 * 1000; // 48 hours
```
CassioMG, line 6: just checking, is the extension also using 48h?

src/config/constants.ts
```diff
@@ -2,8 +2,10 @@
...
 
-export { PASSWORD_MIN_LENGTH };
+export { PASSWORD_MIN_LENGTH, PASSWORD_MAX_LENGTH, HASH_KEY_EXPIRATION_MS };
```
CassioMG, line 8: just out of curiosity, why are we exporting those consts like this? 
```
export { PASSWORD_MIN_LENGTH, PASSWORD_MAX_LENGTH, HASH_KEY_EXPIRATION_MS };
```

instead of this:
```
export const PASSWORD_MIN_LENGTH = 8;
export const PASSWORD_MAX_LENGTH = 32;
export const HASH_KEY_EXPIRATION_MS = 48 * 60 * 60 * 1000; // 48 hours
```

src/config/constants.ts
```diff
@@ -88,30 +90,16 @@ export const DEFAULT_NETWORKS: Array<NetworkDetails> = [
...
-  HASH_KEY_EXPIRE_AT = "hashKeyExpireAt",
 }
 
 /**
  * Sensitive storage keys.
  *
  * TEMPORARY_STORE The temporary store contains encrypted private keys and mnemonic phrase.
  * HASH_KEY The hash key and salt in an JSON stryngified object. This is used to encrypt and decrypt the temporary store.
- * HASH_KEY format: { hashKey: string, salt: string }
+ * HASH_KEY format: { hashKey: string, salt: string, expiration: number }
```
CassioMG, line 100:
```suggestion
 * HASH_KEY format: { hashKey: string, salt: string, expiresAt: number }
```

src/navigators/RootNavigator.tsx
```diff
@@ -1,95 +1,71 @@
...
-    if (hasAccount) {
-      return ROOT_NAVIGATOR_ROUTES.AUTH_STACK;
+    if (authStatus === AUTH_STATUS.HASH_KEY_EXPIRED) {
+      return ROOT_NAVIGATOR_ROUTES.LOCK_SCREEN;
     }
 
-    // Otherwise show the auth stack (welcome, signup, etc.)
-    return ROOT_NAVIGATOR_ROUTES.AUTH_STACK;
-  };
+    return ROOT_NAVIGATOR_ROUTES.MAIN_TAB_STACK;
```
CassioMG, line 41: Should we consider `AUTH_STATUS.NOT_AUTHENTICATED` as default instead of `AUTH_STATUS.AUTHENTICATED` ?

The way it's implemented now works but I think it's a _little_ scary to consider "authenticated" as the default behavior here. Wdyt?

src/ducks/auth.ts
```diff
@@ -336,29 +947,253 @@ export const useAuthenticationStore = create<AuthStore>()((set) => ({
...
+
+  /**
+   * Wipes all user data for debugging purposes
+   *
+   * Removes all keys from key manager, clears all stored data,
+   * and resets the store state to initial.
+   *
+   * @returns {Promise<boolean>} True if successful, false otherwise
+   */
+  wipeAllDataForDebug: async () => {
```
CassioMG, line 1154: could we guard this function with a `if (!__DEV__) return;` just in case to be future proof?

src/components/screens/HistoryScreen.tsx
```diff
@@ -45,36 +76,21 @@ export const HistoryScreen = () => {
...
+            isFullWidth
+            onPress={handleWipeData}
+            isLoading={isWiping}
+            destructive
+          >
+            <Text>Debug: Wipe All Data</Text>
```
CassioMG, line 92: Could we remove the `Text` tags from the buttons on this screen and just use the string directly so they render the expected style?

src/components/screens/HomeScreen.tsx
```diff
@@ -52,8 +54,10 @@ const BorderLine = styled.View`
 `;
 
 export const HomeScreen = () => {
-  const publicKey = TEST_PUBLIC_KEY;
-  const networkDetails = TEST_NETWORK_DETAILS;
+  const { account } = useGetActiveAccount();
+  const { network } = useAuthenticationStore();
+  const publicKey = account?.publicKey;
+  const networkDetails = network ?? NETWORKS.PUBLIC;
```
CassioMG, line 60: could we please rename `networkDetails` to something else? If we look at `constants.ts` we'll see that `NetworkDetails` is an object type with several keys and here `networkDetails` is actually a `string` which is referring to a `NETWORKS` value.

src/components/screens/LoadingScreen.tsx
```diff
@@ -0,0 +1,18 @@
...
+const Container = styled.View`
+  flex: 1;
+  align-items: center;
+  justify-content: center;
+`;
+
+export const LoadingScreen: React.FC = () => (
+  <BaseLayout useSafeArea>
+    <Container>
+      <ActivityIndicator />
```
CassioMG, line 15:
```suggestion
      <ActivityIndicator size="large" color={THEME.colors.secondary}/>
```

src/components/screens/LoadingScreen.tsx
```diff
@@ -0,0 +1,18 @@
+import { BaseLayout } from "components/layout/BaseLayout";
```
CassioMG, line 1:
```suggestion
import { BaseLayout } from "components/layout/BaseLayout";
import { THEME } from "config/theme";
```

src/components/screens/ImportWalletScreen.tsx
```diff
@@ -47,6 +52,11 @@ export const ImportWalletScreen: React.FC<ImportWalletScreenProps> = ({
         value={recoveryPhrase}
         onChangeText={setRecoveryPhrase}
       />
+      {error && (
```
CassioMG, line 55: should we use the built-in `error` prop from `Textarea` to render the error here?

src/hooks/useGetActiveAccount.ts
```diff
@@ -1,249 +1,92 @@
...
 /**
- * Interface for temporary store data structure
+ * Checks if a hash key is expired
  */
-interface TemporaryStore {
-  expiration: number;
-  privateKeys: Record<string, string>;
-  mnemonicPhrase: string;
-}
+export const isHashKeyExpired = (hashKey: HashKey): boolean =>
```
CassioMG, line 14: should we move this `isHashKeyExpired`  to ducks/auth ?

src/hooks/useGetActiveAccount.ts
```diff
@@ -1,249 +1,92 @@
...
+  Date.now() > hashKey.expiresAt;
 
 /**
- * Retrieves data from the temporary store
+ * Gets the public key of the active account. Used on lock screen.
  */
-const getTemporaryStore = async (): Promise<TemporaryStore | null> => {
+export const getActiveAccountPublicKey = async (): Promise<string | null> => {
```
CassioMG, line 20: is this `getActiveAccountPublicKey` a duplicate from the `getActiveAccountPublicKey` in ducks/auth? should we leave it only in ducks/auth?

src/components/screens/LockScreen.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <Text secondary>{t("lockScreen.description")}</Text>
+          <StyledInputContainer>
+            <Input
+              isPassword
+              placeholder={t("lockScreen.passwordInputPlaceholder")}
+              fieldSize="lg"
+              value={passwordValue}
+              onChangeText={handlePasswordChange}
+            />
+            {error && <StyledErrorText>{error}</StyledErrorText>}
```
CassioMG, line 135:
```suggestion
            {error && <Text color={THEME.colors.status.error}>{error}</Text>}
```

I think we could probably also create an `error` Text prop in a separate PR so we could use it like `<Text error> </Text>`.  :)

src/components/screens/LockScreen.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  return (
+    <BaseLayout useSafeArea useKeyboardAvoidingView>
+      <Container>
+        <StyledIconContainer>
+          <FreighterLogo width={px(48)} height={px(48)} />
+        </StyledIconContainer>
+        <StyledFormContainer>
+          <Avatar size="lg" publicAddress={publicKey ?? ""} />
+          <StyledTitle>{t("lockScreen.title")}</StyledTitle>
```
CassioMG, line 125:
```suggestion
          <Display xs semiBold>{t("lockScreen.title")}</Display>
```

src/components/screens/LockScreen.tsx
CassioMG: Actually, should we use `Input`'s built-in `error` prop to render the error here?

src/components/screens/LockScreen.tsx
```diff
@@ -0,0 +1,155 @@
...
+    error,
+    authStatus,
+  } = useAuthenticationStore();
+  const [publicKey, setPublicKey] = useState<string | null>(null);
+  const [passwordValue, setPasswordValue] = useState("");
+  const { t } = useAppTranslation();
+
+  // Monitor auth status changes to navigate when unlocked
+  useEffect(() => {
+    if (authStatus === "AUTHENTICATED") {
```
CassioMG, line 73:
```suggestion
    if (authStatus === AUTH_STATUS.AUTHENTICATED) {
```

src/ducks/auth.ts
```diff
@@ -1,72 +1,198 @@
...
-import { isHashKeyValid } from "hooks/useGetActiveAccount";
+import { isHashKeyExpired } from "hooks/useGetActiveAccount";
 import { t } from "i18next";
+import { clearTemporaryData, getHashKey } from "services/storage/helpers";
 import {
   dataStorage,
   secureDataStorage,
 } from "services/storage/storageFactory";
-import { fromMnemonic } from "stellar-hd-wallet";
+import StellarHDWallet from "stellar-hd-wallet";
```
CassioMG, line 35: @Brunonascdev let's try doing it in a separate PR, but I think we should update this package.
It's been recently upgraded from `0.0.10` to `1.0.1` a few days ago after 6 years of inactivity 😅 
As a reference we have this [recent task](https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=104072709&issue=stellar%7Cfreighter%7C1933) for the extension.

CassioMG: I've created an issue for us to tackle this next sprint: https://github.com/stellar/freighter-mobile/issues/76

src/ducks/auth.ts
```diff
@@ -275,53 +631,308 @@ const signUp = async ({
...
+      await logoutInternal();
+      throw new Error(t("authStore.error.noKeyPairFound"));
+    }
+
+    const accountListRaw = await dataStorage.getItem(STORAGE_KEYS.ACCOUNT_LIST);
+    const accountList = JSON.parse(accountListRaw ?? "[]") as Account[];
+    let account = accountList.find((a) => a.id === activeAccount);
+
+    if (!account) {
+      logger.error("signIn", "Account not found in account list");
```
CassioMG, line 696: when do we expect this to happen?

src/ducks/auth.ts
```diff
@@ -275,53 +631,308 @@ const signUp = async ({
...
+}: ImportWalletParams): Promise<void> => {
+  try {
+    // Generate a key pair from the mnemonic
+    const wallet = StellarHDWallet.fromMnemonic(mnemonicPhrase);
+    const keyDerivationNumber = 0;
+
+    const keyPair = {
+      publicKey: wallet.getPublicKey(keyDerivationNumber),
+      privateKey: wallet.getSecret(keyDerivationNumber),
+    };
```
CassioMG, line 742: looks like we could extract this to a reusable function

src/ducks/auth.ts
```diff
@@ -1,72 +1,198 @@
...
 import { create } from "zustand";
 
-// Parameters for signUp function
+/**
+ * Parameters for signUp function
+ *
+ * @interface SignUpParams
+ * @property {string} mnemonicPhrase - The mnemonic phrase for the wallet
+ * @property {string} password - User's password for encrypting the wallet
+ * @property {boolean} [imported] - Whether the wallet was imported (optional)
```
CassioMG, line 44: it appears we are not calling `signUp` with this `imported` parameter anywhere. Should we remove it?

is this because we are using `importWallet` instead?

src/ducks/auth.ts
```diff
@@ -275,53 +631,308 @@ const signUp = async ({
...
+      ];
+
+      await Promise.all(promises);
+    }
+
+    // Store the account in the key manager and create the temporary store
+    await storeAccount({
+      mnemonicPhrase,
+      password,
+      keyPair,
```
CassioMG, line 760: should we pass `imported: true` here to `storeAccount` ?

will we use this `imported` flag to derive the first 5 accounts from the seed phrase like we're doing on the extension?

src/ducks/auth.ts
```diff
@@ -336,29 +947,253 @@ export const useAuthenticationStore = create<AuthStore>()((set) => ({
...
+    if (navigationRef && navigationRef.isReady()) {
+      // Check if we're already on the lock screen to prevent navigation loops
+      const currentRoute = navigationRef.getCurrentRoute();
+      if (currentRoute?.name === ROOT_NAVIGATOR_ROUTES.LOCK_SCREEN) {
+        // Already on lock screen, don't navigate again
+        return;
+      }
+
+      // Use resetRoot instead of navigate to avoid warnings with conditional navigators
+      navigationRef.resetRoot({
```
CassioMG, line 1088: Does it mean it'll erase everything that's in the stack and leave only the Lock Screen?

If so, I think we should take care about this approach as I believe in the future we'll only want to display the Lock Screen when users try to write something in the account. E.g. a user tries to make a Send operation but the hash key is expired, in this case I think we should probably push the Lock Screen on top of the current stack so that the user can input the password and proceed with the Send operation. What do you think?

CassioMG: I pictured that the Lock Screen would always be put on top of the navigation stack and once the user inputs the password we would just "pop" it from the stack so that users could continue from where they stopped. Have you tried this approach yet? Was it throwing those "warnings with conditional navigators" which is in the comment?

CassioMG (commented):
@Brunonascdev great job on this! 🚀🚀🚀 

I've left some questions and suggestions.
I'm mostly worried about the `LockScreen` navigation clearing all the current navigation stack.

src/components/screens/HistoryScreen.tsx
Brunonascdev: I leave it there so people can reset their auth state on the teaser

src/ducks/auth.ts
Brunonascdev: I tried to replicate the same structure from the extension, but I ended up creating another method for it. I will remove that parameter.

src/ducks/auth.ts
Brunonascdev: yes, that was my plan. since i'm not doing that yet, I'll just remove the parameter for now.

src/ducks/auth.ts
Brunonascdev: We discussed offline but:

We can leave like this for the moment, but ideally we should just pop the LockScreen, without actually resetting the state.

src/config/constants.ts
Brunonascdev: they're using 24hrs. I updated my version

src/components/screens/LockScreen.tsx
```diff
@@ -131,8 +124,10 @@ export const LockScreen: React.FC<LockScreenProps> = ({ navigation }) => {
               fieldSize="lg"
               value={passwordValue}
               onChangeText={handlePasswordChange}
+              error={
```
CassioMG, line 127: I think we should just pass it like `error={error}` and it should work. In case the error text needs some tweak I think we should change it inside the Input component so we keep consistency in all places that use the Input component. Wdyt?

CassioMG (approved):
LGTM! Just left another comment regarding using the `error` prop from `Input` component

src/components/screens/LockScreen.tsx
Brunonascdev: Makes sense! I changed it to send the error only.

PR #81, opened 2025-04-03
Copy public key

### LINKED ISSUE #61: Copy public key
Users should be able to copy their public key when tapping Home's copy button and also when using the "..." button on the Wallets list.

The Wallets list will be implemented [in this other task](https://github.com/orgs/stellar/projects/58/views/1?pane=issue&itemId=104212835&issue=stellar%7Cfreighter-mobile%7C59), please sync with the person assigned to it to avoid duplicate work.

We should probably display a "Account address copied!" [toast](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3021-7382&m=dev) to give some visual feedback to users. Please double check with Charles about this.

Image
Image

### Toast sample
Image

danilosilvackl:
This PR implements the copy action and another piece of the stellar design system:

- Implemented the `Toast` component and `ToastProvider`
  - The `Toast` component is basically the `Notification` component from the stellar design system, just adapted to the mobile usage
  - The `ToastProvider` was created to make sure that the whole application can use trigger the `Toast` whenever needed
- The `useClipboard` hook that encapsulates the `Clipboard` actions and the `Toast` usage for user feedback
- updated the direct calls for the `Clipboard.setString` to use the new hook
- updated the `HomeScreen` tests
  
## iOS 
https://github.com/user-attachments/assets/cae2a665-3d1a-471f-93e9-60870a2b776d

## Android 🤖
[Screen_recording_20250403_150952.webm](https://github.com/user-attachments/assets/c524f4ad-7203-46c6-a34f-9e0a77ef8b69)

closes #61


src/hooks/useClipboard.ts
```diff
@@ -0,0 +1,62 @@
...
+  const { showToast } = useToast();
+
+  const copyToClipboard = useCallback(
+    (text: string, options: CopyToClipboardOptions = {}) => {
+      try {
+        Clipboard.setString(text);
+
+        if (options.showNotification !== false) {
+          showToast({
+            title: options.notificationMessage || "Copied to clipboard!",
```
Brunonascdev, line 47: move texts to translation file

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,129 @@
...
+ * showToast({
+ *   variant: "success",
+ *   title: "Success!",
+ *   message: "Your action was completed successfully.",
+ * });
+ */
+export const useToast = (): ToastContextType => {
+  const context = useContext(ToastContext);
+  if (!context) {
+    throw new Error("useToast must be used within a ToastProvider");
```
Brunonascdev, line 69: move error messages to translation file

src/components/sds/Toast/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+      return PALETTE.dark.red["09"];
+    case "warning":
+      return PALETTE.dark.amber["09"];
+    case "secondary":
+      return PALETTE.dark.gray["11"];
+    case "primary":
+    default:
+      return PALETTE.dark.lilac["09"];
+  }
+};
```
Brunonascdev, line 69: we can do something like this:

```typescript
const variantColors = {
  success: {
    iconColor: PALETTE.dark.green["09"],
    backgroundColor: PALETTE.dark.green["03"]
  },
  // ...
}
```

and then use it like that:

```typescript
const getIconColor = (variant: ToastVariant): string => {
  return variantColors[variant];
};
```

this may help maintaining the colors here

Brunonascdev:
Looking good! left a few comments related to i18n and small suggestions.

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,129 @@
...
+    );
+  }, []);
+
+  const contextValue = useMemo(() => ({ showToast }), [showToast]);
+
+  return (
+    <ToastContext.Provider value={contextValue}>
+      {children}
+      <ToastContainer>
+        <SafeAreaView edges={["top"]} style={{ width: "100%" }}>
```
CassioMG, line 115: could we use `useSafeAreaInsets` to handle the safe area (like we have on `BaseLayout`) to keep consistency?

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,129 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  z-index: 1000;
+  align-items: center;
+  pointer-events: none;
+`;
+
+const ToastWrapper = styled(View)`
```
CassioMG, line 43:
```suggestion
const ToastWrapper = styled.View`
```

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,129 @@
+import { Toast, ToastProps, ToastVariant } from "components/sds/Toast";
+import { px } from "helpers/dimensions";
+import React, {
+  createContext,
+  useCallback,
+  useContext,
+  useState,
+  useMemo,
+} from "react";
+import { View } from "react-native";
```
CassioMG, line 10:
```suggestion
```

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,129 @@
...
+  right: 0;
+  z-index: 1000;
+  align-items: center;
+  pointer-events: none;
+`;
+
+const ToastWrapper = styled(View)`
+  width: 100%;
+  padding-top: ${px(16)};
+  pointer-events: auto;
```
CassioMG, line 46: does `auto` effectively work here considering we are setting `none` on the parent component?

src/providers/ToastProvider.tsx
danilosilvackl: done!

src/components/sds/Toast/index.tsx
danilosilvackl: great idea, applied this to both the background and the icon color

src/providers/ToastProvider.tsx
danilosilvackl: done, thanks for pointing it out

src/providers/ToastProvider.tsx
danilosilvackl: you were right, i was updating the styles and forgot to take this of afterwards

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,129 @@
...
+ * </ToastProvider>
+ */
+export const ToastProvider: React.FC<ToastProviderProps> = ({ children }) => {
+  const [toasts, setToasts] = useState<ToastPropsWithId[]>([]);
+
+  const showToast = useCallback((options: ToastOptions) => {
+    const newToast: ToastPropsWithId = {
+      ...options,
+      id: Date.now().toString(),
+      onDismiss: () => {
```
CassioMG, line 96: I'm confused, it looks like `onDismiss` is being set here but also on line 121. Is this a leftover?

src/components/screens/HomeScreen.tsx
```diff
@@ -58,9 +58,19 @@ export const HomeScreen = () => {
...
 
   const { formattedBalance } = useTotalBalance();
 
+  const handleCopyAddress = () => {
+    if (publicKey) {
+      copyToClipboard(publicKey, {
+        notificationMessage: t("accountAddressCopied"),
+        toastVariant: "success",
+      });
+    }
```
CassioMG, line 71:
```suggestion
    if (!publicKey) {
      return;
     }
    
     copyToClipboard(publicKey, {
        notificationMessage: t("accountAddressCopied"),
        toastVariant: "success",
      });
```

src/hooks/useClipboard.ts
```diff
@@ -0,0 +1,64 @@
+import Clipboard from "@react-native-clipboard/clipboard";
+import useAppTranslation from "hooks/useAppTranslation";
+import { useToast } from "providers/ToastProvider";
+import { useCallback } from "react";
+
+interface CopyToClipboardOptions {
+  /** Whether to show a toast notification (default: true) */
+  showNotification?: boolean;
```
CassioMG, line 8: Since the default would be showing the notification how do you feel about using a `hideNotification` instead?

I think it'd read a _little_ better on line 47 so we could do something like:
`if (options.hideNotification) return;`

I'm not strong about it though!

src/components/sds/Toast/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+  border-radius: ${px(8)};
+  padding: ${px(12)};
+  margin: ${px(8)} 0;
+  flex-direction: column;
+  gap: ${px(8)};
+  width: 100%;
+  shadow-color: #000;
+  shadow-offset: 0px 2px;
+  shadow-opacity: 0.25;
+  shadow-radius: 3.84px;
```
CassioMG, line 91: should we use `px()` here?

src/components/sds/Toast/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+          transform: [{ translateY: slideAnim }],
+        }}
+      >
+        <TitleContainer>
+          <IconContainer>
+            {icon || (
+              <Icon.InfoCircle size={16} color={getIconColor(variant)} />
+            )}
+          </IconContainer>
+          <TitleText>{title}</TitleText>
```
CassioMG, line 199:
```suggestion
          <Text sm semiBold>{title}</Text>
```

src/components/sds/Toast/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+        </TitleContainer>
+        {message && <MessageText>{message}</MessageText>}
```
CassioMG, line 201:
```suggestion
        {message && <Text sm secondary>{message}</Text>}
```

src/components/sds/Toast/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+  line-height: ${px(20)};
+  font-weight: 600;
+`;
+
+const MessageText = styled(Text)`
+  color: ${PALETTE.dark.gray["11"]};
+  font-size: ${px(14)};
+  line-height: ${px(20)};
+  font-weight: 400;
+`;
```
CassioMG, line 120:
```suggestion
```

CassioMG: We should always try using the existing `Typography/Text` component directly which should support most (or all) of Figma styles, it has some cool shorthands for that!

We'll rarely (maybe never) set custom font size, height or weight differently than what we already have on `Typography/Text`.

CassioMG: We could though wrap the existing `Text` component to tweak other styles like paddings or margins if needed

src/components/sds/Toast/index.tsx
CassioMG: @danilosilvackl @Brunonascdev hot-hint: you can usually check which Text style you need by checking this comment on Figma Typography code like on the image below
Screenshot 2025-04-03 at 12 32 00
you'll only need to check the Typography file to realize that weight `600` is the same as `semiBold`

src/components/sds/Toast/index.tsx
CassioMG: @danilosilvackl @Brunonascdev also for the text color you can check it in a few places like the below:
Screenshot 2025-04-03 at 12 36 56

src/providers/ToastProvider.tsx
danilosilvackl: yeah, i forgot to remove this one after extracting the logic, thanks!

src/components/screens/HomeScreen.tsx
danilosilvackl: done

src/components/sds/Toast/index.tsx
danilosilvackl: yes indeed, done

src/hooks/useClipboard.ts
danilosilvackl: i like the idea, the early return makes the code cleaner too

src/components/sds/Toast/index.tsx
danilosilvackl: its slightly different for me because i don't have the dev mode, but it does have all those infos there! thanks for sharing
some styling was different because i got the exact same colors and styling from the scss from the stellar design system project instead of figma, but its fixed now!

src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,125 @@
...
+  pointer-events: none;
+`;
+
+interface ToastWrapperProps {
+  $insets: EdgeInsets;
+}
+
+const ToastWrapper = styled.View<ToastWrapperProps>`
+  width: 100%;
+  padding-top: ${({ $insets }: ToastWrapperProps) => px($insets.top + 16)};
```
CassioMG, line 48:
```suggestion
  padding-top: ${({ $insets }: ToastWrapperProps) =>  `${$insets.top + pxValue(16)}px`};
```

CassioMG (approved):
LGTM! Good job on making the the Clipboard and Toast reusable 👏 
Just left a comment regarding insets vs px()

PR #91, opened 2025-04-11
Fix infinity loop on new accounts

CassioMG:
Calling `handleRefresh` from `useBalancesList` in `FriendbotButton` was triggering an infinity loop. This PR fixes that by removing `useBalancesList` from `FriendbotButton` and making a single call to `fetchAccountBalances` from `useBalancesStore` like it was originally implemented.

### Before fix
https://github.com/user-attachments/assets/37bc6806-d548-424c-b737-510b1d640520

### After fix
https://github.com/user-attachments/assets/0e8e8811-5b68-4c43-a9f2-55aabbe1eda8

### DIFF
--- a/src/hooks/useBalancesList.ts
+++ b/src/hooks/useBalancesList.ts
@@ -1,7 +1,7 @@
 import { NETWORKS } from "config/constants";
 import { PricedBalance } from "config/types";
 import { useBalancesStore } from "ducks/balances";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 
 interface UseBalancesListResult {
   balanceItems: Array<PricedBalance & { id: string }>;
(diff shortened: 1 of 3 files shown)


Brunonascdev (approved):

src/hooks/useBalancesList.ts
```diff
@@ -91,12 +91,16 @@ export const useBalancesList = ({
   }, [fetchAccountBalances, publicKey, network]);
 
   // Convert balances object to array
-  const balanceItems = Object.entries(pricedBalances).map(
-    ([id, balance]) =>
-      ({
-        id,
-        ...balance,
-      }) as PricedBalance & { id: string },
+  const balanceItems = useMemo(
```
CassioMG, line 94: This change is not related to the bug but it improves app performance by preventing unnecessary re-renders due to changes in pricedBalances object refs

PR #95, opened 2025-04-14
Settings - Version label

### LINKED ISSUE #63: Settings - Version label
Add the "Version" label component to the [Settings screen](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3062-194354&t=2t5EQGlViuteDaoe-1).
The "Update" label/feature is not part of this task.

Image

danilosilvackl:
Added the version label to the settings screen
_disclaimer: we are currently getting the version from the package.json_

## iOS 
Simulator Screenshot - iPhone 16 Pro - 2025-04-14 at 17 42 46

## Android 🤖
Screenshot_1744664038

closes #63

### DIFF
--- a/src/components/screens/SettingsScreen.tsx
+++ b/src/components/screens/SettingsScreen.tsx
@@ -6,9 +6,10 @@ import Icon from "components/sds/Icon";
 import { SETTINGS_ROUTES, SettingsStackParamList } from "config/routes";
 import { THEME } from "config/theme";
 import { useAuthenticationStore } from "ducks/auth";
+import { getAppVersion } from "helpers/version";
 import useAppTranslation from "hooks/useAppTranslation";
 import React, { useEffect } from "react";
-import { TouchableOpacity } from "react-native";
+import { TouchableOpacity, View } from "react-native";
 
 type SettingsScreenProps = NativeStackScreenProps<
   SettingsStackParamList,
@@ -20,6 +21,7 @@ export const SettingsScreen: React.FC<SettingsScreenProps> = ({
 }) => {
   const { logout } = useAuthenticationStore();
   const { t } = useAppTranslation();
+  const appVersion = getAppVersion();
 
   useEffect(() => {
     navigation.setOptions({
@@ -45,9 +47,20 @@ export const SettingsScreen: React.FC<SettingsScreenProps> = ({
     },
   ];
 
+  const updateListItems = [
+    {
+      icon: <Icon.GitCommit size={24} color={THEME.colors.list.disabled} />,
+      title: t("settings.version", { version: appVersion }),
+      testID: "update-button",
+    },
+  ];
+
   return (
     <BaseLayout insets={{ top: false }}>
-      <List items={listItems} />
+      <View className="flex flex-col gap-6">
+        <List items={listItems} />
+        <List items={updateListItems} />
+      </View>
     </BaseLayout>
   );
 };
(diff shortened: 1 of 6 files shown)


src/components/screens/SettingsScreen.tsx
```diff
@@ -45,9 +47,20 @@ export const SettingsScreen: React.FC<SettingsScreenProps> = ({
     },
   ];
 
+  const updateListItems = [
```
CassioMG, line 50:
```suggestion
  const bottomListItems = [
```

src/components/screens/SettingsScreen.tsx
```diff
@@ -45,9 +47,20 @@ export const SettingsScreen: React.FC<SettingsScreenProps> = ({
...
+      testID: "update-button",
+    },
+  ];
+
   return (
     <BaseLayout insets={{ top: false }}>
-      <List items={listItems} />
+      <View className="flex flex-col gap-6">
+        <List items={listItems} />
+        <List items={updateListItems} />
```
CassioMG, line 62:
```suggestion
        <List items={bottomListItems} />
```

src/components/screens/SettingsScreen.tsx
```diff
@@ -45,9 +47,20 @@ export const SettingsScreen: React.FC<SettingsScreenProps> = ({
...
+      title: t("settings.version", { version: appVersion }),
+      testID: "update-button",
+    },
+  ];
+
   return (
     <BaseLayout insets={{ top: false }}>
-      <List items={listItems} />
+      <View className="flex flex-col gap-6">
+        <List items={listItems} />
```
CassioMG, line 61:
```suggestion
        <List items={topListItems} />
```

danilosilvackl: done!

src/helpers/version.ts
```diff
@@ -0,0 +1,3 @@
+import packageJson from "package.json";
+
+export const getAppVersion = (): string => packageJson.version;
```
CassioMG, line 3: @danilosilvackl @Brunonascdev how do we feel about using [getVersion()](https://github.com/react-native-device-info/react-native-device-info?tab=readme-ov-file#getversion) from `react-native-device-info` to fetch the version? I think it'd be more reliable than the package.json version because it would show the actual app version which is displayed in the app store.

Besides that, I think we should display the build number ([getBuildNumber()](https://github.com/react-native-device-info/react-native-device-info?tab=readme-ov-file#getbuildnumber)) in addition to the app version which is useful for debugging something like: `v{appVersion} ({buildVersion})`

danilosilvackl: i did experiment with it for a bit, its good and works well, i didnt follow up with it because it raised some other questions as well.

first, today we don't have a reliable way to manipulate the versions right, so we have to do it manually on the android files and on ios files then maybe on the package.json

then i tried making a bash script that did that, it worked fine but then the react-native-device-info was having some trouble to get the updated version after a build, so i gave up for now hehe

but anyways, i think we can use that package to get the versions for sure!

CassioMG: > it worked fine but then the react-native-device-info was having some trouble to get the updated version after a build

this is unexpected, does this issue also happens after you manually update an app/build version?

danilosilvackl: @CassioMG @Brunonascdev updated to use the `react-native-device-info` now

CassioMG (approved):

Issue #73, opened 2025-03-31
Data flow - Add and manage assets/trustlines


CassioMG:
This task is about implementing the data flow (business logic) of the [Manage Assets flow](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3050-82429&m=dev) where users should be able to add Classic Assets, SACs and Custom Soroban Tokens (Does the extension support adding Custom Tokens? If not, we could skip it for now).

The UI should be implemented on this [other task](https://github.com/orgs/stellar/projects/58/views/1?pane=issue&itemId=104212886&issue=stellar%7Cfreighter-mobile%7C60).

Please note that on this phase we should ignore the `hide asset`, `assets list check` and `Blockaid asset verification` components as they should be added in future phases.

<img width="2504" alt="Image" src="https://github.com/user-attachments/assets/1d038732-3b00-46d3-b24c-04412c038c23" />

CassioMG:
@Brunonascdev working!

PR #96, opened 2025-04-15
Manage Assets Data Flow

Brunonascdev:
This PR implements the data flow for the Manage Assets feature. It allows the user to add different assets categories (classic and SAC) using the Stellar Expert and Freighter Backend APIs.

closes issue #73 

## Android

[manage-assets-flow-android.webm](https://github.com/user-attachments/assets/d15007fb-cdf5-415f-a93e-3fd0f6551126)

## iOS

https://github.com/user-attachments/assets/f4d867cd-153a-495d-a5b0-e8b58a1776ab

## Adding Soroban Custom Asset

[adding-soroban-asset.webm](https://github.com/user-attachments/assets/e6825bb1-8ef5-43ff-b9b7-0409b3f9f0d2)


src/components/Spinner.tsx
```diff
@@ -0,0 +1,15 @@
+import { THEME } from "config/theme";
+import React from "react";
+import { ActivityIndicator } from "react-native";
+
+interface SpinnerProps {
+  size?: "small" | "large";
+  color?: string;
+}
+
+const Spinner: React.FC<SpinnerProps> = ({
```
danilosilvackl, line 10: i like this wrapper on the `ActivityIndicator` can you take a look at the code and update the other places that uses it to the Spinner please?

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+      {asset.domain && (
+        <Text sm secondary>
+          {asset.domain}
+        </Text>
+      )}
+      <View className="mt-2" />
+      <Badge
+        variant="secondary"
+        size="md"
+        icon={<Icon.ShieldPlus size={pxValue(16)} />}
```
danilosilvackl, line 66: you can remove the `pxValue` here

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <View className="flex-row items-center p-6 bg-background-tertiary rounded-xl justify-center">
+        <Text md secondary style={{ textAlign: "center" }}>
+          {t("addAssetScreen.disclaimer")}
+        </Text>
+      </View>
+      <View className="mt-6" />
+      <View className="w-full flex-row items-center px-6 py-4 bg-background-primary border border-border-primary rounded-xl justify-between">
+        <View className="flex-row items-center">
+          <Icon.UserCircle
+            size={pxValue(16)}
```
danilosilvackl, line 81: remove pxValue

src/helpers/soroban.ts
danilosilvackl, line 1: i don't think it makes sense to make this change now, but I think those values could be moved to a .env file in the future

src/hooks/useColors.ts
danilosilvackl, line 1: this hook here is beautiful

src/services/stellar.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (ledgerCapacityUsageNum > 0.5 && ledgerCapacityUsageNum <= 0.75) {
+      networkCongestion = NetworkCongestion.MEDIUM;
+    } else if (ledgerCapacityUsageNum > 0.75) {
+      networkCongestion = NetworkCongestion.HIGH;
+    } else {
+      networkCongestion = NetworkCongestion.LOW;
+    }
+  } catch (e) {
+    // use default values
+    recommendedFee = "100";
```
danilosilvackl, line 108: extract this to a const to avoid using magic string

src/services/stellarExpert.ts
```diff
@@ -0,0 +1,35 @@
...
+      logger.error(
+        "stellarExpertApi.searchAsset",
+        "Invalid response from stellarExpert",
+        response.data,
+      );
+
+      throw new Error("Invalid response from stellarExpert");
+    }
+
+    return response.data;
```
danilosilvackl, line 30: i think we should explore the usage of Models in the future, just to make it easier to use on the UI, instead of having to navigate through the entire response we can make Models more straightforward. 
but again, its just a thing to think about for the future, since this is not the only place we work like this, its a huge refactor point

danilosilvackl (approved):
looking very neat! left some suggestions

src/services/indexer.ts
```diff
@@ -0,0 +1,132 @@
+/* eslint-disable no-underscore-dangle */
+import { AxiosError } from "axios";
+import { INDEXER_URL, NETWORKS } from "config/constants";
+import { logger } from "config/logger";
+import { GetTokenDetailsParams, TokenDetailsResponse } from "config/types";
+import { getNativeContractDetails } from "helpers/soroban";
+import { createApiService } from "services/apiFactory";
+
+const indexerApi = createApiService({
+  baseURL: INDEXER_URL,
```
CassioMG, line 10: @Brunonascdev we already have the `backend.ts` service which points to this indexer url, I think we should probably merge them. What do you think?

I thought about naming it as `backend` as the `INDEXER_URL` looks like this `https://freighter-backend-prd.stellar.org/api/v1` - maybe we should rename `INDEXER_URL` to `FREIGHTER_BACKEND_URL` and the service could be named `freighterBackend` for consistency and clarity. What do you think?

src/services/stellarExpert.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-underscore-dangle */
+import { NETWORKS } from "config/constants";
+import { logger } from "config/logger";
+import { SearchAssetResponse } from "config/types";
+import { getApiStellarExpertUrl } from "helpers/stellarExpert";
+import { createApiService } from "services/apiFactory";
+
+export const searchAsset = async (asset: string, network: NETWORKS) => {
+  const stellarExpertApi = createApiService({
```
CassioMG, line 9: will this create a new service instance every time `searchAsset` is called? Should we maybe create a map of static Stellar Expert services per network so we prevent creating multiple service instances? Wdyt?

CassioMG: since Stellar Expert has support for only Pubnet and Testnet maybe we don't even need a map, we could probably just create `stellarExpertPubnetApi` and `stellarExpertTestnetApi` on this same file and alternate between them depending if `NETWORKS` is Testnet or not

src/services/stellar.ts
```diff
@@ -0,0 +1,150 @@
...
+  });
+
+export const submitTx = async (
+  input: SubmitTxParams,
+): Promise<Horizon.HorizonApi.SubmitTransactionResponse> => {
+  const { network, tx } = input;
+  const { networkUrl, networkPassphrase } = mapNetworkToNetworkDetails(network);
+  const server = stellarSdkServer(networkUrl);
+
+  const txXdr =
```
CassioMG, line 67:
```suggestion
  const transaction =
```

src/services/stellar.ts
```diff
@@ -0,0 +1,150 @@
...
+export const buildChangeTrustTx = async (input: BuildChangeTrustTxParams) => {
+  const { network, publicKey, assetIdentifier, isRemove = false } = input;
+  const { assetCode, issuer } = formatAssetIdentifier(assetIdentifier);
+  const { networkUrl, networkPassphrase } = mapNetworkToNetworkDetails(network);
+
+  const server = stellarSdkServer(networkUrl);
+  const account = await server.loadAccount(publicKey);
+  const { recommendedFee } = await getNetworkFees(server);
+
+  const tx = new TransactionBuilder(account, {
```
CassioMG, line 124:
```suggestion
  const txBuilder = new TransactionBuilder(account, {
```

src/services/stellar.ts
```diff
@@ -0,0 +1,150 @@
...
+  ).setTimeout(DEFAULT_TRANSACTION_TIMEOUT);
+
+  return tx.build().toXDR();
+};
+
+export const signTransaction = (input: SignTxParams): string => {
+  const { tx, secretKey, network } = input;
+  const { networkPassphrase } = mapNetworkToNetworkDetails(network);
+  const txXdr = typeof tx === "string" ? tx : tx.toXDR();
+  const txEnvelope = TransactionBuilder.fromXDR(txXdr, networkPassphrase);
```
CassioMG, line 144:
```suggestion
  const transaction = TransactionBuilder.fromXDR(txXdr, networkPassphrase);
```

src/config/types.ts
```diff
@@ -197,6 +203,13 @@ export type PricedBalance = Balance &
     fiatCode?: string;
     fiatTotal?: BigNumber | null;
     displayName?: string;
+    token: {
```
CassioMG, line 206: Considering the `Balance` type currently is:
```
export type Balance =
  | NativeBalance
  | ClassicBalance
  | SorobanBalance
  | LiquidityPoolBalance;
```
Overwriting the `token` property here would break for `LiquidityPoolBalance` since they don't have a token property associated to it.

@Brunonascdev what do you think? Would it be possible to not overwrite `token` here and handle it differently?

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+            color={themeColors.foreground.primary}
+          />
+          <Text
+            md
+            secondary
+            style={{ textAlign: "center", marginLeft: pxValue(8) }}
```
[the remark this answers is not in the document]
Brunonascdev: it's kinda tricky because the `Text` component is currently not prepared for handling nativewind styling 😬

src/services/stellarExpert.ts
Brunonascdev: I was thinking the same thing. I'd love to use some `zod` validation here to create a couple of schemas and validate the responses we receive from these external services.

Maybe we can create a task for future development @CassioMG?

src/services/indexer.ts
Brunonascdev: makes sense! I'll implement that

src/components/screens/ManageAssetsScreen.tsx
```diff
@@ -12,105 +12,100 @@ import {
...
         </TouchableOpacity>
       ),
     });
-  }, [navigation, t]);
+  }, [navigation, t, themeColors]);
 
-  const handleCopyTokenAddress = (balance: PricedBalance) => {
-    copyAssetAddress(balance, "manageAssetsScreen.tokenAddressCopied");
-  };
+  const removeAsset = async (assetId: string) => {
```
CassioMG, line 68: should we extract this `removeAsset`  logic to a store or a hook?

src/components/screens/AddAssetScreen/AddAssetRightContent.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <Button
+      secondary
+      squared
+      lg
+      testID="add-asset-button"
+      icon={
+        <Icon.PlusCircle
+          size={pxValue(16)}
```
CassioMG, line 26:
```suggestion
          size={16}
```

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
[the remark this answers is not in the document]
danilosilvackl: aw too bad D:

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
```diff
@@ -0,0 +1,137 @@
...
+          ...(asset.isNative
+            ? { type: "native" }
+            : {
+                type: "credit_alphanum4",
+              }),
+          code: asset.assetCode,
+          issuer: {
+            key: asset.issuer,
+          },
+        }}
```
CassioMG, line 50: looks like this code is duplicated on `AddAssetBottomSheetContent` and `AssetItem` components. Could we create a helper for it?

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
CassioMG, line 66:
```suggestion
        icon={<Icon.ShieldPlus size={16} />}
```

src/components/screens/AddAssetScreen/AddAssetBottomSheetContent.tsx
CassioMG, line 81:
```suggestion
            size={16}
```

src/components/screens/AddAssetScreen/AddAssetScreen.tsx
```diff
@@ -0,0 +1,411 @@
...
+    })
+    .sort((a) => {
+      if (a.hasTrustline) {
+        return -1;
+      }
+
+      return 1;
+    });
+
+const AddAssetScreen: React.FC<AddAssetScreenProps> = ({ navigation }) => {
```
CassioMG, line 121: @Brunonascdev do you think it would be worth to extract some of the logic from this screen to a hook or store?

src/services/stellarExpert.ts
CassioMG: Sounds good to me, let's create a task to explore this improvement. @danilosilvackl would you mind creating a task for it? Maybe adding some simple code or schema sample in the task description for visualization - thanks!

src/components/screens/AddAssetScreen/AssetItem.tsx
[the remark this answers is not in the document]
Brunonascdev: it seems kinda different from the purpose of this assetItem 🤔

src/helpers/soroban.ts
Brunonascdev: this likely won't change since it's the native XLM contract

danilosilvackl: but is it ok to be public like this? its genuine question because i'm not a 100% sure on this

src/services/stellarExpert.ts
danilosilvackl: here it is: https://github.com/stellar/freighter-mobile/issues/98

CassioMG: Looks great, thanks Danilo! 👏 
I've added it to Backlog so we can go through it during Backlog grooming meetings

src/components/AssetIcon.tsx
```diff
@@ -100,6 +111,20 @@ export const AssetIcon: React.FC<AssetIconProps> = ({
     );
   }
 
+  if (token.type === "custom_token") {
```
CassioMG, line 114: Comment suggestion:
// For Soroban custom tokens, use the Soroban logo

src/services/backend.ts
```diff
@@ -288,14 +289,17 @@ export const handleContractLookup = async (
...
   return {
     assetCode: tokenDetails.symbol,
     domain: "Stellar Network",
     hasTrustline: false,
-    issuer: isSacContract
-      ? (tokenDetails.name.split(":")[1] ?? "")
-      : contractId,
+    issuer,
     isNative: false,
+    assetType: isSacContract ? getAssetType(contractId) : "custom_token",
```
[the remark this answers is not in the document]
CassioMG: Or maybe create an enum with all the asset types we use in `getAssetType` in case we can't import them from `@stellar/stellar-sdk`:

```
 if (assetIdentifier === "XLM") {
    return "native";
  }

  if (assetIdentifier.includes(":")) {
    const { assetCode, issuer } = formatAssetIdentifier(assetIdentifier);

    if (issuer.startsWith("C")) {
      return "custom_token";
    }

    if (assetCode.length <= 4) {
      return "credit_alphanum4";
    }

    if (assetCode.length >= 5 && assetCode.length <= 12) {
      return "credit_alphanum12";
    }
  }

  return "liquidity_pool_shares";
```

Wdyt?

src/config/types.ts
```diff
@@ -272,3 +267,30 @@ export interface GetTokenDetailsParams {
...
+  isNative: boolean;
+  assetType?: AssetTypeWithCustomToken;
+};
+
+/**
+ * Custom token metadata for a single token
+ */
+export type CustomToken = {
+  contractId: string;
+  symbol: string;
```
CassioMG, line 285: should we also have `name` and `decimals` here?

src/config/constants.ts
```diff
@@ -110,10 +111,13 @@ export const mapNetworkToNetworkDetails = (network: NETWORKS) => {
  *
  * ACTIVE_ACCOUNT The active account is the account that is currently being used.
  * ACCOUNT_LIST The account list is used to keep track of all the accounts stored in the key manager.
+ * CUSTOM_TOKEN_LIST The custom token list is used to keep track of all the custom soroban tokens stored in the key manager.
+ * Formatted as: { [publicKey: string]: { network: NETWORKS, contractId: string, symbol: string }[]} @see CustomTokenStorage
  * */
 export enum STORAGE_KEYS {
   ACTIVE_ACCOUNT_ID = "activeAccountId",
   ACCOUNT_LIST = "accountList",
+  CUSTOM_TOKEN_LIST = "customTokenList",
```
CassioMG, line 120: are we removing this on logout?

Brunonascdev: no, this is an object containing a list of custom assets for all the accounts that used freighter on that device. since the API today doesn't store user's custom tokens, we're following this approach (it's handled the same way as the extension)

src/config/constants.ts
```diff
@@ -114,7 +114,8 @@ export const mapNetworkToNetworkDetails = (network: NETWORKS) => {
  * ACTIVE_ACCOUNT The active account is the account that is currently being used.
  * ACCOUNT_LIST The account list is used to keep track of all the accounts stored in the key manager.
  * CUSTOM_TOKEN_LIST The custom token list is used to keep track of all the custom soroban tokens stored in the key manager.
- * Formatted as: { [publicKey: string]: { network: NETWORKS, contractId: string, symbol: string }[]} @see CustomTokenStorage
+ * Formatted as: { [publicKey: string]: { [network: string]: CustomToken[] } } @see CustomTokenStorage
+ * The CUSTOM_TOKEN_LIST is not removed during the logout process. It is used to keep the custom tokens even after the user logs out, since the API does not store custom tokens.
```
CassioMG, line 118: Would it result in all users that use a same device to see the same custom tokens across all accounts even though they haven't added them?

Brunonascdev: no because it's stored by publicKey *and* network. 

when we're loading the balances I get this entry from the storage, filter by the active user public key and active network to get the custom tokens.

if another user logs in (not even a different user, could be the same with a different public key) and they don't have any custom token list associated with their public key *and* current network there's no custom tokens to show.

CassioMG:
@Brunonascdev for the Custom Token case I think we should replace "Trustline" with something else as it only makes sense for classic assets. I think we also don't currently support trading Custom Tokens, do we?

How is the copy looking on the extension while adding a custom token?

Screenshot 2025-04-23 at 12 01 43

CassioMG:
> @Brunonascdev for the Custom Token case I think we should replace "Trustline" with something else as it only makes sense for classic assets. I think we also don't currently support trading Custom Tokens, do we?
> 
> How is the copy looking on the extension while adding a custom token?
> 

@Brunonascdev I think we'll also need to rename some functions like `handleAddAssetTrustline` to `handleAddAsset` instead and maybe other functions with same naming issue to avoid confusion. Wdyt?

src/helpers/__mocks__/dimensions.ts
```diff
@@ -0,0 +1,46 @@
+/**
```
CassioMG, line 1: there is an existing __mocks__ folder in the root directory, would you mind trying moving this file to there so we centralize our mocks there? Thanks!

src/hooks/useAssetLookup.ts
```diff
@@ -0,0 +1,175 @@
...
+}: UseAssetLookupProps) => {
+  const [searchTerm, setSearchTerm] = useState("");
+  const [searchResults, setSearchResults] = useState<
+    FormattedSearchAssetRecord[]
+  >([]);
+  const [status, setStatus] = useState<AssetLookupStatus>(
+    AssetLookupStatus.IDLE,
+  );
+
+  const checkIfUserHasTrustline = (
```
CassioMG, line 43: I'd suggest we rename it to something like `checkHasTrustline` or `checkTrustline` since an `User` could have multiple accounts/balances and this function only checks for a single account/balances. Does that make sense?

src/services/stellar.ts
```diff
@@ -0,0 +1,153 @@
...
+    .setTimeout(DEFAULT_TRANSACTION_TIMEOUT);
+
+  return txBuilder.build().toXDR();
+};
+
+export const signTransaction = (input: SignTxParams): string => {
+  const { tx, secretKey, network } = input;
+  const { networkPassphrase } = mapNetworkToNetworkDetails(network);
+  const transaction = typeof tx === "string" ? tx : tx.toXDR();
+  const txEnvelope = TransactionBuilder.fromXDR(transaction, networkPassphrase);
```
CassioMG, line 147:
```suggestion
  const transaction = TransactionBuilder.fromXDR(transactionXDR, networkPassphrase);
```

Brunonascdev:
> @Brunonascdev for the Custom Token case I think we should replace "Trustline" with something else as it only makes sense for classic assets. I think we also don't currently support trading Custom Tokens, do we?
> 
> How is the copy looking on the extension while adding a custom token?
> 
> Screenshot 2025-04-23 at 12 01 43

it's very different from the mobile version:

adding USDC:
Screenshot 2025-04-23 at 15 10 01

adding a custom soroban token:
Screenshot 2025-04-23 at 15 05 01

Brunonascdev:
For future reference:

We’ve aligned on updating the copy in the Add Asset BottomSheet to remove any mention of “trustlines,” since no ChangeTrust operation is triggered when adding a custom Soroban token.

Here’s the revised copy currently in place:

Screenshot_1745436621

CassioMG (approved):
LGTM - great job on this @Brunonascdev! 🚀 🚀 🚀

PR #103, opened 2025-04-23
Android release configs

CassioMG:
This PR adds the Android configs needed for signing and building an app bundle in `release` mode.

I'll still create a document with the instructions mentioned on the commits below.


android/app/build.gradle
```diff
@@ -92,6 +92,14 @@ android {
             keyAlias 'androiddebugkey'
             keyPassword 'android'
         }
+        release {
+            storeFile file('release.keystore')
+            // Replace with keystore password
+            storePassword '<release-keystore-password>'
```
CassioMG, line 98: we should later move this to env variables so we don't need to manually replace them every time

package.json
```diff
@@ -5,6 +5,8 @@
   "license": "UNLICENSED",
   "scripts": {
     "android": "react-native run-android",
+    "android-release": "yarn android --mode release",
```
CassioMG, line 8: script to run the app in release mode in your device or emulator

package.json
```diff
@@ -5,6 +5,8 @@
...
+    "android:bundle:release": "npx react-native build-android --mode=release",
```
CassioMG, line 9: script to generate the app bundle file which we need to upload to Google Play Console to distribute to testers

the bundle file is named `app-release.aab` and will be placed on `/freighter-mobile/android/app/build/outputs/bundle/release/app-release.aab`

Brunonascdev (approved):

Issue #136, opened 2025-05-19
Insufficient Top Padding on View Components - ManageAssetsScreen and HistoryScreen


Brunonascdev:
Reported by @JakeUrban 

> The components at the top of the view seem unusually high, very close to the literal top of the screen. Should we add some more padding?

![Image](https://github.com/user-attachments/assets/3d23d6e1-aed9-4586-8b3f-c4fd35e11e81)

![Image](https://github.com/user-attachments/assets/e6feb912-88ff-42fe-a5da-40f2cd68986c)

PR #146, opened 2025-06-02
FIX: add extra top padding to ManageAssets and History screens

luiz-jung:
add extra top padding to ManageAssets and History screens

# What's new:
- Added extra top padding to both ManageAssets and History screen

Closes: #136


danilosilvackl:
Can you make a test just to make sure? if possible remove the padding you added and remove the prop `insets={{ top: false }}` from those screens to check if the spacing matches the figma design.

its just out of curiosity, if it matches good we can follow with that, if not i would prefer to stick with your fix

Issue #135, opened 2025-05-19
Keyboard Not Appearing When Text Field Is Focused - LockScreen


Brunonascdev:
Reported by @JakeUrban:

> I ran into this interesting bug where it didn't bring up my keyboard when I selected the text field

![Image](https://github.com/user-attachments/assets/3d18dac8-9b06-4996-8527-3cf7febb7bcd)

CassioMG:
@JakeUrban I'd like to get a bit more context on this one. Did it work on the second tap? Are you able to repro this issue on the MVP build?

cc @luiz-jung

CassioMG:
Yeah, so @luiz-jung is suggesting to add the `autoFocus` property for the lock screen on [this PR](https://github.com/stellar/freighter-mobile/pull/150), which means whenever the user is navigated to this screen the keyboard will automatically show up with the input field selected. I think it could be a good remedy for this mysterious bug, plus we would be removing 1 tap from the UX so users could immediately type their password. How does that sound?

PR #150, opened 2025-06-06
fix: keyboard not appearing when text field is focused on lockscreen

luiz-jung:
Ensure keyboard appears wen focusing input at lockscreen

# What's new:
- Adding methods to force focus on input and keyboard to make sure it opens when touching the input

## iOS 

https://github.com/user-attachments/assets/a4ab4010-265f-477c-b54a-a838db7eb571

## Android 🤖

https://github.com/user-attachments/assets/07713578-2aa6-4138-87ee-b801307602b9

Closes: #135

### DIFF
--- a/src/components/sds/Input/index.tsx
+++ b/src/components/sds/Input/index.tsx
@@ -2,8 +2,8 @@ import Clipboard from "@react-native-clipboard/clipboard";
 import { Text } from "components/sds/Typography";
 import { THEME } from "config/theme";
 import { fs, px } from "helpers/dimensions";
-import React from "react";
-import { Platform, TouchableOpacity } from "react-native";
+import React, { useState } from "react";
+import { Platform, TouchableOpacity, TextInput } from "react-native";
 import styled from "styled-components/native";
 
 // =============================================================================
@@ -255,6 +255,7 @@ interface InputProps {
   editable?: boolean;
   autoCapitalize?: "none" | "sentences" | "words" | "characters";
   autoCorrect?: boolean;
+  autoFocus?: boolean;
   keyboardType?:
     | "default"
     | "number-pad"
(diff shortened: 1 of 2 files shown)


src/components/sds/Input/index.tsx
```diff
@@ -264,142 +265,154 @@ interface InputProps {
...
+                {labelSuffix}
+              </Text>
+            )}
+          </Text>
         )}
 
-        <StyledTextInput
-          testID={testID}
+        <InputContainer
+          testID={testID ? `${testID}-container` : undefined}
```
danilosilvackl, line 333:
```suggestion
          testID={testID ? `${testID}-input-container` : undefined}
```

just to make it more unique

src/components/sds/Input/index.tsx
```diff
@@ -264,142 +265,154 @@ interface InputProps {
...
           $fieldSize={fieldSize}
-          $hasLeftElement={!!leftElement || copyButton?.position === "left"}
-          $hasRightElement={
-            !!rightElement || copyButton?.position === "right" || !!endButton
-          }
+          $isError={Boolean(isError || error)}
```
danilosilvackl, line 335: is this casting really necessary?

src/components/sds/Input/index.tsx
```diff
@@ -264,142 +265,154 @@ interface InputProps {
...
+            placeholder={placeholder}
+            placeholderTextColor={THEME.colors.text.secondary}
+            value={value}
+            onChangeText={onChangeText}
+            editable={editable}
+            secureTextEntry={isPassword && !showPassword}
+            autoCorrect={autoCorrect}
+            autoFocus={props.autoFocus}
+            $fieldSize={fieldSize}
+            $hasLeftElement={Boolean(
```
danilosilvackl, line 355: is the casting to `Boolean` really necessary for those conditions?

src/components/sds/Input/index.tsx
danilosilvackl, line 1: i like your work here, but i'm also thinking about how this was basically a refactor maybe we could use a bit more of time and convert this to native-wind already
what do you think?
what your input here too @CassioMG ?

src/components/templates/InputPasswordTemplate.tsx
```diff
@@ -37,6 +43,19 @@ const InputPasswordTemplate: React.FC<InputPasswordTemplateProps> = ({
...
+    const focusTimeout = setTimeout(() => {
+      inputRef.current?.focus();
+      Keyboard.dismiss();
+      setTimeout(() => {
+        inputRef.current?.focus();
+      }, 50);
+    }, 100);
+
+    return () => clearTimeout(focusTimeout);
+  }, []);
```
danilosilvackl, line 58: nit:
```suggestion
  useEffect(() => {
    const focusTimeout = setTimeout(() => {
      inputRef.current?.focus();
      
      Keyboard.dismiss();
      
      setTimeout(() => {
        inputRef.current?.focus();
      }, 50);
    }, 100);

    return () => clearTimeout(focusTimeout);
  }, []);
```

src/components/sds/Input/index.tsx
CassioMG: Hmm I'm note sure this looks like a big refactor. If you [remove the whitespaces on this PR](https://github.com/stellar/freighter-mobile/pull/150/files?w=1) you'll see that it is not applying many changes on the Input component. Wdyt?

I'd personally prefer migrating this component to native-wind on a separate task

src/components/sds/Input/index.tsx
luiz-jung: made the id more unique ${testID}-input-container

src/components/sds/Input/index.tsx
danilosilvackl: makes total sense, also I learned how to use the filter for better PR reviews, thanks!

src/components/templates/InputPasswordTemplate.tsx
luiz-jung: I ended up completely removing the timeout as it made the template complex with no real visible result on using it

Issue #124, opened 2025-05-09
Account details


CassioMG:
This task is about adding the last missing account detail action compared to the extension

We should add a new option to this dropdown menu that _will_ open the account view on the stellar.expert 
`"View on stellar.expert"`

<img width="322" alt="Image" src="https://github.com/user-attachments/assets/7dae3e01-b38e-4b2e-a7df-5c1464f458ad" />

and update the copy of the rename account modal `"Skip"` button to `"Cancel"` following the [design specs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3062-197380&m=dev)
<img width="322" alt="Image" src="https://github.com/user-attachments/assets/5ea45f3b-3274-48f5-a627-2763ff73fa3d" />

PR #151, opened 2025-06-06
feat: add view on stellar.expert button and change skip to cancel in rename modal

luiz-jung:
Making some tweaks at account details

# What's new:
- Added a new button to wallet options that takes the user to stellar.expert website
- Changed the button from skip to cancel at rename wallet

## iOS 

https://github.com/user-attachments/assets/908429e1-be1f-40dd-adf6-e9465b7be696

## Android 🤖

https://github.com/user-attachments/assets/33e14425-f877-469b-9154-d050ce7eeea1

Closes: #124

### DIFF
--- a/src/components/screens/HomeScreen/AccountItemRow.tsx
+++ b/src/components/screens/HomeScreen/AccountItemRow.tsx
@@ -3,8 +3,10 @@ import ContextMenuButton, { MenuItem } from "components/ContextMenuButton";
 import Avatar from "components/sds/Avatar";
 import Icon from "components/sds/Icon";
 import { Text } from "components/sds/Typography";
+import { NETWORKS } from "config/constants";
 import { Account } from "config/types";
 import { truncateAddress } from "helpers/stellar";
+import { getStellarExpertUrl } from "helpers/stellarExpert";
 import useAppTranslation from "hooks/useAppTranslation";
 import useColors from "hooks/useColors";
 import React from "react";
@@ -44,7 +46,7 @@ const AccountItemRow: React.FC<AccountItemRowProps> = ({
   });
 
   const handleViewOnExplorer = async () => {
-    const url = `https://stellar.expert/explorer/public/account/${account.publicKey}`;
+    const url = `${getStellarExpertUrl(NETWORKS.PUBLIC)}/account/${account.publicKey}`;
     await Linking.openURL(url);
   };
 
--- a/src/components/screens/HomeScreen/RenameAccountModal.tsx
+++ b/src/components/screens/HomeScreen/RenameAccountModal.tsx
@@ -85,7 +85,7 @@ const RenameAccountModal: React.FC<RenameAccountModalProps> = ({
             onPress={() => setModalVisible(false)}
             disabled={isRenamingAccount}
           >
-            {t("renameAccountModal.cancel")}
+            {t("common.cancel")}
           </Button>
         </View>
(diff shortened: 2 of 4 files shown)


src/components/screens/HomeScreen/AccountItemRow.tsx
```diff
@@ -34,13 +34,20 @@ const AccountItemRow: React.FC<AccountItemRowProps> = ({
...
     },
     android: {
       renameWallet: "baseline_edit",
       copyAddress: "copy",
+      viewOnExplorer: "public",
     },
   });
 
+  const handleViewOnExplorer = async () => {
+    const url = `https://stellar.expert/explorer/public/account/${account.publicKey}`;
```
danilosilvackl, line 47: build the url using one of the consts from constants.ts 
eg: `STELLAR_EXPERT_URL` or `STELLAR_EXPERT_API_URL`

src/components/screens/HomeScreen/RenameAccountModal.tsx
```diff
@@ -85,7 +85,7 @@ const RenameAccountModal: React.FC<RenameAccountModalProps> = ({
             onPress={() => setModalVisible(false)}
             disabled={isRenamingAccount}
           >
-            {t("renameAccountModal.skip")}
+            {t("renameAccountModal.cancel")}
```
danilosilvackl, line 88: nit: you can use the common keys so you don't have to add another key for this
```suggestion
            {t("common.cancel")}
```

danilosilvackl (approved):
left just one comment, but it lgtm!

src/components/screens/HomeScreen/AccountItemRow.tsx
CassioMG: +1

CassioMG: You'll actually need to use `getStellarExpertUrl()` here to handle the public and test networks accordingly. We have a sample of this on the `TransactionDetailsBottomSheet` component.

src/components/screens/HomeScreen/RenameAccountModal.tsx
luiz-jung: removed extra xancel and using common.cancel

src/components/screens/HomeScreen/AccountItemRow.tsx
luiz-jung: changed the url to use getStellarExpertUrl()

src/components/screens/HomeScreen/AccountItemRow.tsx
```diff
@@ -44,7 +46,7 @@ const AccountItemRow: React.FC<AccountItemRowProps> = ({
   });
 
   const handleViewOnExplorer = async () => {
-    const url = `https://stellar.expert/explorer/public/account/${account.publicKey}`;
+    const url = `${getStellarExpertUrl(NETWORKS.PUBLIC)}/account/${account.publicKey}`;
```
CassioMG, line 49: @luiz-jung thanks for making the change. I believe the idea here would be passing the `network` variable we have in store so that the url would automatically point to either `public` or `testnet` depending on which network the user is using. Does that make sense? Do you think we could do that here? Similar to what we have on `TransactionDetailsBottomSheet`

cc @danilosilvackl

PR #165, opened 2025-06-09
WC - Move project id to env var

CassioMG:
Closes https://github.com/orgs/stellar/projects/58/views/1?pane=issue&itemId=114310062

This PR moves the Wallet Connect project id to an env var and update README with instructions.


luiz-jung (approved):
LGTM

Issue #117, opened 2025-05-09
Import Secret key


CassioMG:
Implement the `data flow` for "Import a wallet using a `Stellar secret key + password`."

Please note that we should have a `Imported` label on the imported account on the `Wallets` list modal so users can easily distinguish between the accounts which were derived from the mnemonic phrase from the imported accounts.

[Current Figma designs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=4098-39831&t=EkzvmtTsBz1720oF-1):
<img width="851" alt="Image" src="https://github.com/user-attachments/assets/25549e99-2fb9-43cf-8031-4a39b8a392a2" />
<img width="851" alt="Image" src="https://github.com/user-attachments/assets/31217647-4988-4f97-a291-141e0853976b" />

PR #168, opened 2025-06-10
FEAT: Import Secret key

luiz-jung:
Importing new wallets using their secret keys.

# What's new:
- implementing the logic to import a wallet by it's secret key
- adding a Imported label to imported wallets
- adding logic to automatically move to imported wallet screen

## iOS 

https://github.com/user-attachments/assets/c84d9d1a-2c7b-43f0-8bd5-af7f6d92da3d

## Android 🤖

https://github.com/user-attachments/assets/5d97581f-e625-47eb-bbf4-9f635af0f8bb

Closes: #117

### DIFF
--- a/src/components/screens/HomeScreen/AccountItemRow.tsx
+++ b/src/components/screens/HomeScreen/AccountItemRow.tsx
@@ -70,9 +70,21 @@ const AccountItemRow: React.FC<AccountItemRowProps> = ({
             <Text md primary medium numberOfLines={1}>
               {account.name}
             </Text>
-            <Text sm secondary medium numberOfLines={1}>
-              {truncatedPublicKey}
-            </Text>
+            <View className="flex-row items-center">
+              <Text sm secondary medium numberOfLines={1}>
+                {truncatedPublicKey}
+              </Text>
+              {account.imported && (
+                <>
+                  <Text sm secondary medium>
+                    {" • "}
+                  </Text>
+                  <Text sm secondary medium>
+                    Imported
+                  </Text>
+                </>
+              )}
+            </View>
           </View>
         </View>
       </TouchableOpacity>
--- a/src/components/screens/ImportSecretKeyScreen.tsx
+++ b/src/components/screens/ImportSecretKeyScreen.tsx
@@ -1,3 +1,4 @@
+import { CommonActions } from "@react-navigation/native";
 import { NativeStackScreenProps } from "@react-navigation/native-stack";
 import { StrKey } from "@stellar/stellar-sdk";
 import { BaseLayout } from "components/layout/BaseLayout";
@@ -8,6 +9,8 @@ import { Text, Display } from "components/sds/Typography";
 import {
   MANAGE_WALLETS_ROUTES,
   ManageWalletsStackParamList,
+  MAIN_TAB_ROUTES,
+  ROOT_NAVIGATOR_ROUTES,
 } from "config/routes";
 import { useAuthenticationStore } from "ducks/auth";
 import useAppTranslation from "hooks/useAppTranslation";
@@ -54,10 +57,28 @@ const ImportSecretKeyScreen: React.FC<ImportSecretKeyScreenProps> = ({
     }
 
     try {
-      await importSecretKey(secretKey);
-      navigation.goBack();
+      await importSecretKey(secretKey, password);
+
+      navigation.dispatch(
+        CommonActions.reset({
+          index: 0,
+          routes: [
+            {
+              name: ROOT_NAVIGATOR_ROUTES.MAIN_TAB_STACK,
+              state: {
+                index: 1,
+                routes: [
+                  { name: MAIN_TAB_ROUTES.TAB_HISTORY },
+                  { name: MAIN_TAB_ROUTES.TAB_HOME },
+                  { name: MAIN_TAB_ROUTES.TAB_DISCOVERY },
+                ],
+              },
+            },
+          ],
+        }),
+      );
     } catch (err) {
-      // Error from the store will be handled separately
+      // Error handling is managed by the auth store
     }
   };
 
--- a/src/ducks/auth.ts
+++ b/src/ducks/auth.ts
@@ -187,6 +187,7 @@ interface DeriveKeypairParams {
  * @property {Function} signUp - Signs up a new user with the provided credentials
  * @property {Function} signIn - Signs in a user with the provided password
  * @property {Function} importWallet - Imports a wallet with the provided seed phrase and password
+ * @property {Function} importSecretKey - Imports a secret key
  * @property {Function} getAuthStatus - Gets the current authentication status
  * @property {Function} fetchActiveAccount - Fetches the currently active account
  * @property {Function} refreshActiveAccount - Refreshes the active account data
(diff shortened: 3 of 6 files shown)


src/components/screens/HomeScreen/AccountItemRow.tsx
```diff
@@ -70,9 +70,21 @@ const AccountItemRow: React.FC<AccountItemRowProps> = ({
...
+              <Text sm secondary medium numberOfLines={1}>
+                {truncatedPublicKey}
+              </Text>
+              {account.imported && (
+                <>
+                  <Text sm secondary medium>
+                    {" • "}
+                  </Text>
+                  <Text sm secondary medium>
+                    Imported
```
CassioMG, line 83: reminder to add translation for this text

src/ducks/auth.ts
```diff
@@ -1701,5 +1727,58 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => {
...
+        };
+
+        await storeAccount({
+          mnemonicPhrase: temporaryStore.mnemonicPhrase,
+          password,
+          keyPair,
+          accountNumber: existingAccounts.length + 1,
+          shouldRefreshHashKey: false,
+          shouldSetActiveAccount: true,
+          imported: true,
```
CassioMG, line 1765:
```suggestion
          importedFromSecretKey: true,
```

I'm suggesting to make it more descriptive so we don't get confused with the other existing `importWallet` function. Let's make it explicit that this one is being imported from an external secret key so we know it's not being derived from the seed phrase. Does that make sense?

src/ducks/auth.ts
```diff
@@ -824,6 +849,7 @@ const verifyAndCreateExistingAccountsOnNetwork = async (
         id: keyStore.id,
         name: accountName,
         publicKey: keyPair.publicKey,
+        imported: false,
```
CassioMG, line 852:
```suggestion
        importedFromSecretKey: false,
```

src/ducks/auth.ts
```diff
@@ -659,6 +683,7 @@ const storeAccount = async ({
       id: keyStore.id,
       name: accountName,
       publicKey,
+      imported,
```
CassioMG, line 686:
```suggestion
      importedFromSecretKey,
```

src/ducks/auth.ts
```diff
@@ -624,6 +647,7 @@ const storeAccount = async ({
   accountNumber,
   shouldRefreshHashKey = true,
   shouldSetActiveAccount = true,
+  imported = false,
```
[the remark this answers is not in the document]
CassioMG: let's also remove `isAppendingAccount` from `StoreAccountParams` interface as it's not being used anywhere.
Friendly reminder to update the JSDoc comments of the `StoreAccountParams` interface to reflect all the existing params.

src/ducks/auth.ts
```diff
@@ -291,7 +293,28 @@ const getAllAccounts = async (): Promise<Account[]> => {
...
 
-  return JSON.parse(accountListRaw) as Account[];
+  const accounts = JSON.parse(accountListRaw) as Account[];
+
+  // Migration: ensure all accounts have the imported property
+  let needsMigration = false;
+  const migratedAccounts = accounts.map((account) => {
+    if (account.imported === undefined) {
+      needsMigration = true;
+      return { ...account, imported: false }; // Default to false for existing accounts
```
CassioMG, line 303:
```suggestion
    if (account.importedFromSecretKey === undefined) {
      needsMigration = true;
      return { ...account, importedFromSecretKey: false }; // Default to false for existing accounts
```

src/ducks/auth.ts
```diff
@@ -291,7 +293,28 @@ const getAllAccounts = async (): Promise<Account[]> => {
     return [];
   }
 
-  return JSON.parse(accountListRaw) as Account[];
+  const accounts = JSON.parse(accountListRaw) as Account[];
+
+  // Migration: ensure all accounts have the imported property
```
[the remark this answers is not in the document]
CassioMG: it will also affect the performance _a little bit_ since every time we call `getAllAccounts` it would be looping/mapping over all accounts and creating a new array

src/ducks/auth.ts
```diff
@@ -1701,5 +1727,58 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => {
...
+        const temporaryStore = await getTemporaryStore();
+        if (!temporaryStore) {
+          throw new Error(t("authStore.error.temporaryStoreNotFound"));
+        }
+
+        const keyPair = {
+          publicKey,
+          privateKey,
+        };
+
```
CassioMG, line 1757:
```suggestion
```

CassioMG: I'm suggesting removing it from here and making it inline since we already have a `keypair` const on line 1735 so it looks a bit less confusing

src/ducks/auth.ts
```diff
@@ -1701,5 +1727,58 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => {
...
+      try {
+        const keypair = Keypair.fromSecret(secretKey);
+        const publicKey = keypair.publicKey();
+        const privateKey = keypair.secret();
+
+        const existingAccounts = await getAllAccounts();
+        const accountExists = existingAccounts.some(
+          (account) => account.publicKey === publicKey,
+        );
```
CassioMG, line 1742: nit: you could use `hasAccountInAccountList()` here

src/ducks/auth.ts
```diff
@@ -200,6 +201,7 @@ interface AuthActions {
   signUp: (params: SignUpParams) => void;
   signIn: (params: SignInParams) => Promise<void>;
   importWallet: (params: ImportWalletParams) => void;
+  importSecretKey: (secretKey: string, password: string) => Promise<void>;
```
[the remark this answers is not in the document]
CassioMG: I know this code convention is not being applied everywhere in our codebase but would be awesome if we could stick to it as much as possible

src/ducks/auth.ts
```diff
@@ -1701,5 +1727,58 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => {
...
+
+        await storeAccount({
+          mnemonicPhrase: temporaryStore.mnemonicPhrase,
+          password,
+          keyPair,
+          accountNumber: existingAccounts.length + 1,
+          shouldRefreshHashKey: false,
+          shouldSetActiveAccount: true,
+          imported: true,
+        });
```
CassioMG, line 1766: Could we extract this chunk of code to a local `importSecretKey` function so we keep this auth action clean?
The same we are doing with `createAccount`, something like:

```
    importSecretKey: async (...) => {
      set({ ... });

      try {
        await importSecretKey(...);

        await Promise.all([get().getAllAccounts(), get().fetchActiveAccount()]);

        set({ ... });
      } catch (error) {
        set({ ... });
        throw error;
      }
    },
```

src/components/screens/ImportSecretKeyScreen.tsx
```diff
@@ -54,10 +57,28 @@ const ImportSecretKeyScreen: React.FC<ImportSecretKeyScreenProps> = ({
...
+                routes: [
+                  { name: MAIN_TAB_ROUTES.TAB_HISTORY },
+                  { name: MAIN_TAB_ROUTES.TAB_HOME },
+                  { name: MAIN_TAB_ROUTES.TAB_DISCOVERY },
+                ],
+              },
+            },
+          ],
+        }),
+      );
```
CassioMG, line 79: On VerifyPasswordScreen we have something similar but cleaner, have we tried using the same here?:
```
        navigation.reset({
          index: 0,
          // @ts-expect-error: This is a valid route.
          routes: [{ name: ROOT_NAVIGATOR_ROUTES.MAIN_TAB_STACK }],
        });
```

CassioMG (approved):
Lgtm

danilosilvackl (approved):

PR #181, opened 2025-06-16
Several UI tweaks

CassioMG:
This PR implements several UI tweaks and bug fixes as listed below.

- Add top spacing on "You can not send to yourself" error message.
Screenshot 2025-06-16 at 15 36 24

- Allow `swipe up to dismiss toasts`. Users can now also hold a toast on screen for as long as they want.

https://github.com/user-attachments/assets/74cebb52-3ec3-4127-9a39-96d6fb8f61e0

- `Avoid hiding input field with keyboard` on "Show recovery phrase" screen.
- Go all the way back to initial `Settings` screen when `Done` is tapped on "Show recovery phrase" screen.
*Note: first video shows "before fix", second video shows "after fix" for the above 2 changes.

https://github.com/user-attachments/assets/56adf63a-2e05-4f62-8f6d-4cf6ffdc6d72

https://github.com/user-attachments/assets/b7d7d9ee-d3c1-47cb-8058-0c799e3c37e4

- Always use `input` component of size `lg` since the `md` size gets cut in some Android devices.
    - The app is already displaying `lg` input fields in 90% of cases so sticking to `lg` also increases consistency. We can revisit this in case we have a screen that strictly requires an input of `md` size.
*Note: first video shows "before fix", second video shows "after fix" for the above change.

https://github.com/user-attachments/assets/c2b783e5-824a-4b66-a088-4ccbd3655139

https://github.com/user-attachments/assets/e029d1e0-8486-4bf5-8a6e-ec0b96223810

- Only `capitalize keyboard` where it makes sense.
    - E.g. we shouldn't capitalize on password inputs, but should capitalize on rename account input.
- Lock app in `portrait orientation` mode

### DIFF
--- a/src/components/screens/SettingsScreen/SecurityScreen/ShowRecoveryPhraseScreen/ShowRecoveryPhraseScreen.tsx
+++ b/src/components/screens/SettingsScreen/SecurityScreen/ShowRecoveryPhraseScreen/ShowRecoveryPhraseScreen.tsx
@@ -95,6 +95,7 @@ const ShowRecoveryPhraseScreen: React.FC<ShowRecoveryPhraseScreenProps> = ({
 
         <View className="mb-6">
           <Input
+            fieldSize="lg"
             label={t("showRecoveryPhraseScreen.password")}
             placeholder={t("showRecoveryPhraseScreen.passwordInputPlaceholder")}
             value={password}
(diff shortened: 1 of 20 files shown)


src/components/screens/SettingsScreen/SecurityScreen/ShowRecoveryPhraseScreen/ShowRecoveryPhraseScreen.tsx
```diff
@@ -49,76 +49,90 @@ const ShowRecoveryPhraseScreen: React.FC<ShowRecoveryPhraseScreenProps> = ({
...
-            <View className="mb-6">
-              <Text secondary>
-                {t("showRecoveryPhraseScreen.accessWarning")}
-              </Text>
-            </View>
+        <ScrollView
+          className="flex-1"
+          showsVerticalScrollIndicator={false}
+          keyboardShouldPersistTaps="handled"
+        >
```
CassioMG, line 62: @luiz-jung @danilosilvackl FYI, when we have too much content on screen, you'll probably need to do something like this in conjunction with `useKeyboardAvoidingView` so the whole view is scrolled up to display the input field.

luiz-jung (approved):
lgtm

Issue #118, opened 2025-05-09
UI - Swap


CassioMG:
This task is about implementing `only the UI` for the `Swap` flow. We should be able to `reuse existing components` like the balances list, numeric keyboard screen, set fee, set memo, processing and successful screens `previously built for the Home and Send flow` since they look very similar to the ones we need here.

[Figma Designs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3918-56845&m=dev):

<img width="2059" alt="Image" src="https://github.com/user-attachments/assets/3c1966ca-3bc7-4d8b-ac1b-e953fec8394f" />

Issue #119, opened 2025-05-09
Data Flow - Swap


CassioMG:
This task is about implementing `the Data Flow` for the `Swap` flow. We should use the business logic of the extension as inspiration. For now we can add support for `classic assets` swapping only, which means we can leave Soroswap out of it (as on the extension we also [only support classic Swap on Mainnet](https://stellarfoundation.slack.com/archives/C03347FNAHK/p1746812532029589)).

[Task for the UI implementation](https://github.com/stellar/freighter-mobile/issues/118).

[Figma Designs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3918-56845&m=dev):

<img width="2059" alt="Image" src="https://github.com/user-attachments/assets/3c1966ca-3bc7-4d8b-ac1b-e953fec8394f" />

PR #182, opened 2025-06-18
Feat: Swap screen

danilosilvackl:
This PR covers the implementation of the swap screen and data flow as well as many other improvements.

## Whats implemented here 🤔
- Complete swap payment, UI and data flow
- Added segmented control component
- Added some tests
- Refactored some old files to be usable on the swap flow

### Swapping XLM to USDC

https://github.com/user-attachments/assets/17aa3809-c4d4-4e5b-b37b-e736d36c3b6a

### Swapping USDC to BTC

https://github.com/user-attachments/assets/8e0d7d94-60cd-43e0-994c-164b09b39e0a

### Swapping XLM to USDC
[android1.webm](https://github.com/user-attachments/assets/ebdea2de-6061-43c1-995e-3c27f3fd344e)

### Swapping USDC to BTC
[android 2.webm](https://github.com/user-attachments/assets/3a773078-c896-49c4-879c-2ed7ace91212)

## Settings Screens

closes: #118 #119


JakeUrban:
Looks great Danilo, a couple questions / suggestions. @sdfcharles I'd like your input on this too.

- Where are we sourcing the "popular tokens" list?
- Should we be providing the user's tokens first, above the "popular" list? It feels more likely that users will want to swap between tokens they already have in their wallet.
- What happens when I search for a token in the search bar? Can I add a new asset to my wallet from there?
- Lets change the "no path found between assets". Users may not know what a path is.
- We may need to round amounts when the number of decimals is large. Attaching a screenshot where this may be needed.
- Can we vertically center the swap results text, by putting "was swapped for" in the middle of the first and second amounts, rather than it being wrapped?

Screenshot 2025-06-18 at 11 08 32 AM
Screenshot 2025-06-18 at 11 38 31 AM

src/helpers/balances.ts
```diff
@@ -298,3 +298,86 @@ export const getAssetType = (
...
+ * @returns {BigNumber} The spendable amount after considering all constraints
+ *
+ * @example
+ * // Calculate spendable XLM amount
+ * const spendable = calculateSpendableAmount(xlmBalance, 5, "0.00001");
+ *
+ * // Calculate spendable amount for other assets
+ * const spendable = calculateSpendableAmount(usdcBalance, 0, "0.00001");
+ */
+export const calculateSpendableAmount = (
```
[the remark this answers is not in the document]
danilosilvackl: yes, makes a lot of sense! I updated it to return the total balance for liquidity pools and the available balance when it exists for all non native tokens

src/ducks/auth.ts
```diff
@@ -1184,11 +1186,31 @@ const getActiveAccount = async (): Promise<ActiveAccount | null> => {
         throw new Error(t("authStore.error.privateKeyNotFound"));
       }
 
+      // Fetch subentry count from the network
```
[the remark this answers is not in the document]
CassioMG: I see we are already fetching this value which is returned on `FetchBalancesResponse`. Would it be too troublesome to reuse that prop so we avoid affecting the performance of the `getActiveAccount` function which now have 2 extra await calls?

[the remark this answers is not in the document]
danilosilvackl: thats not troublesome at all, I mapped it on the balances store and now access it directly, removing the awaits, thanks for the tip!

sdfcharles:
CleanShot 2025-06-18 at 15 57 57
can we decrease the space between the notification and the asset selection?

CleanShot 2025-06-18 at 15 57 38
CleanShot 2025-06-18 at 15 59 18
can we increase the size of the token name on the input? when the user adds a lot of numbers, the whole text "XX XLM" should shrink to fit

CleanShot 2025-06-18 at 15 57 33
can we increase the space between the swap success message and the box

CleanShot 2025-06-18 at 16 01 12
can we add the percentage buttons on the swap screen as well

JakeUrban:
> for sure, I can will update the screen to have the texts that way, do you think we should update the Send flow with this layout as well?

How is this relevant to the send flow? We don't use path payments in the send flow right?

CassioMG:
> > for now it's just looking filtering the list of tokens that the user already have, but it is possible to implement the add flow like the manage assets screen if we refactor and thinker with it a bit
> 
> Ok, yes we need to allow users to swap to a token they don't currently have.

@JakeUrban @danilosilvackl I think we should treat this as a separate task and probably tackle it after V1 is launched. The extension does not currently support it and it wouldn't be quick to implement.

src/services/transactionService.ts
```diff
@@ -105,6 +118,57 @@ export const validateTransactionParams = (
   return null;
 };
 
+/**
+ * Validates swap transaction parameters
+ * Returns an error message if any validation fails
+ */
+export const validateSwapTransactionParams = (params: {
+  senderAddress: string;
```
CassioMG, line 126: looks like senderAddress is not being used here?

src/services/transactionService.ts
```diff
@@ -40,6 +40,19 @@ export interface BuildPaymentTransactionParams {
   senderAddress?: string;
 }
 
+export interface BuildSwapTransactionParams {
+  tokenAmount: string;
+  fromBalance: PricedBalance;
+  toBalance: PricedBalance;
```
CassioMG, line 46:
```suggestion
  sourceAmount: string;
  sourceBalance: PricedBalance;
  destinationBalance: PricedBalance;
```

src/services/transactionService.ts
```diff
@@ -105,6 +118,57 @@ export const validateTransactionParams = (
...
 
+/**
+ * Validates swap transaction parameters
+ * Returns an error message if any validation fails
+ */
+export const validateSwapTransactionParams = (params: {
+  senderAddress: string;
+  fromBalance: PricedBalance;
+  toBalance: PricedBalance;
+  amount: string;
```
CassioMG, line 129:
```suggestion
  sourceBalance: PricedBalance;
  destinationBalance: PricedBalance;
  sourceAmount: string;
```

danilosilvackl:
> > for sure, I can will update the screen to have the texts that way, do you think we should update the Send flow with this layout as well?
> 
> How is this relevant to the send flow? We don't use path payments in the send flow right?

you're right there, I mean this layout from the send screen here:

src/services/transactionService.ts
```diff
@@ -105,6 +118,57 @@ export const validateTransactionParams = (
   return null;
 };
 
+/**
+ * Validates swap transaction parameters
+ * Returns an error message if any validation fails
```
CassioMG, line 123: Will those error messages be displayed to the user? If so, I think we should translate them

src/services/transactionService.ts
```diff
@@ -471,3 +535,115 @@ export const buildPaymentTransaction = async (
...
+  if (!fee) {
+    throw new Error("Transaction fee is required");
+  }
+
+  if (!timeout) {
+    throw new Error("Transaction timeout is required");
+  }
+
+  if (!currentNetwork) {
+    throw new Error("Network is required");
```
CassioMG, line 580: will those appear to users? should we translate them as well?

src/helpers/formatAmount.ts
```diff
@@ -29,34 +29,37 @@ const convertToBigNumber = (
...
+  const val =
+    typeof amount === "string" ? amount : convertToBigNumber(amount).toString();
+
+  const decimal = new Intl.NumberFormat("en-US", { style: "decimal" });
+  const [wholeVal, remainderVal] = val.split(".");
+  const formattedWholeVal = decimal.format(Number(wholeVal)).toString();
+
+  let formattedAmount: string;
+  if (remainderVal) {
+    formattedAmount = `${formattedWholeVal}.${remainderVal}`;
```
CassioMG, line 58: @danilosilvackl is this commit fixing something? Otherwise I'm not sure how big of an improvement it would be. It's breaking internationalization since it's assuming we should always use a dot "." as decimal separator which is not true for many countries.

src/ducks/swap.ts
```diff
@@ -0,0 +1,202 @@
...
+  asset_code?: string;
+  asset_issuer?: string;
+}
+
+interface SwapState {
+  fromTokenId: string;
+  toTokenId: string;
+  fromTokenSymbol: string;
+  toTokenSymbol: string;
+  swapAmount: string;
```
CassioMG, line 29:
```suggestion
  sourceAmount: string;
```

src/ducks/swap.ts
```diff
@@ -0,0 +1,202 @@
...
+  isBuilding: boolean;
+  buildError: string | null;
+
+  setFromToken: (tokenId: string, tokenSymbol: string) => void;
+  setToToken: (tokenId: string, tokenSymbol: string) => void;
+  setSwapAmount: (amount: string) => void;
+  findSwapPath: (params: {
+    fromBalance: PricedBalance;
+    toBalance: PricedBalance;
+    amount: string;
```
CassioMG, line 43:
```suggestion
    sourceAmount: string;
```

src/ducks/swap.ts
```diff
@@ -0,0 +1,202 @@
...
+  destinationAmount: "0",
+  pathResult: null,
+  isLoadingPath: false,
+  pathError: null,
+  isBuilding: false,
+  buildError: null,
+};
+
+const computeDestMinWithSlippage = (
+  amount: string,
```
CassioMG, line 67:
```suggestion
  sourceAmount: string,
```

CassioMG, line 67:
```suggestion
  destinationAmount: string,
```

src/ducks/swap.ts
```diff
@@ -0,0 +1,202 @@
...
+};
+
+/**
+ * Finds the best swap path using Horizon's strict send paths endpoint
+ * This is for classic asset swaps using Stellar's built-in DEX
+ */
+const findClassicSwapPath = async (params: {
+  fromBalance: PricedBalance;
+  toBalance: PricedBalance;
+  amount: string;
```
CassioMG, line 81:
```suggestion
  sourceBalance: PricedBalance;
  destinationBalance: PricedBalance;
  sourceAmount: string;
```

src/ducks/swap.ts
```diff
@@ -0,0 +1,202 @@
...
+    });
+
+    const sourceAmountBN = new BigNumber(amount);
+    const destAmountBN = new BigNumber(bestPath.destination_amount);
+    const conversionRate = destAmountBN.dividedBy(sourceAmountBN).toFixed(7);
+
+    return {
+      sourceAmount: amount,
+      destinationAmount: bestPath.destination_amount,
+      destinationAmountMin: bestPath.destination_amount,
```
CassioMG, line 118: did we repeat `bestPath.destination_amount` here on purpose?

src/ducks/swap.ts
```diff
@@ -0,0 +1,202 @@
...
+        fromBalance,
+        toBalance,
+        amount,
+        network,
+      });
+
+      if (!pathResult) {
+        set({
+          isLoadingPath: false,
+          pathError: "No path found between these assets",
```
CassioMG, line 160: translate?

src/components/BalancesList.tsx
```diff
@@ -69,11 +74,24 @@ interface BalancesListProps {
...
+  customTitle,
   showTitleIcon = false,
   onTokenPress,
+  disableNavigation = false,
+  renderRightContent,
 }) => {
   const { t } = useAppTranslation();
-  const navigation = useNavigation<NavigationProp<RootStackParamList>>();
+
+  // Always call the hook, but handle navigation context errors gracefully
```
CassioMG, line 86: When/where could this navigation error happen?

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -287,20 +269,11 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
-              routes: [{ name: MAIN_TAB_ROUTES.TAB_HISTORY }],
-            },
-          },
-        ],
-      }),
-    );
+    navigation.reset({
+      index: 0,
+      // @ts-expect-error: This is a valid route.
+      routes: [{ name: ROOT_NAVIGATOR_ROUTES.MAIN_TAB_STACK }],
```
CassioMG, line 275: Will this take users back to the balances tab? I believe the intent on the Send flow is taking users back to the history tab

src/components/screens/SwapScreen/helpers/swapCalculations.ts
```diff
@@ -0,0 +1,108 @@
...
+    BigNumber(allowedSlippage).dividedBy(100),
+  );
+
+  return toAmountBN.multipliedBy(slippageMultiplier).toFixed(7);
+};
+
+/**
+ * Formats transaction date for display
+ */
+export const formatTransactionDate = (createdAt?: string): string => {
```
CassioMG, line 65: I think we are also formatting transaction date in other places like the send flow. If so, we should probably export this to our main helper folder so we could use it anywhere.

src/components/screens/SwapScreen/helpers/swapCalculations.ts
```diff
@@ -0,0 +1,108 @@
...
+  if ("contractId" in balance && balance.contractId) {
+    return balance.contractId;
+  }
+
+  if ("token" in balance && balance.token && "issuer" in balance.token) {
+    return balance.token.issuer.key;
+  }
+
+  if (balance.id === "native") {
+    return null;
```
CassioMG, line 104: why are we returning "null" for native?

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -0,0 +1,146 @@
...
+    SwapStackParamList,
+    typeof SWAP_ROUTES.SWAP_AMOUNT_SCREEN
+  >;
+  resetSwap: () => void;
+}
+
+interface UseSwapTransactionResult {
+  isProcessing: boolean;
+  executeSwap: () => Promise<void>;
+  prepareSwapTransaction: () => Promise<void>;
```
CassioMG, line 36: should we rename it to something else? "prepare" makes me think it's handling a soroban transaction which is not the case here

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    try {
+      signTransaction({
+        secretKey: account.privateKey,
+        network,
+      });
+
+      await submitTransaction({ network });
+    } catch (error) {
+      logger.error("SwapTransaction", "Swap failed", error);
```
CassioMG, line 111: shouldn't we throw here?

src/components/screens/SwapScreen/components/TokenSelectionContent.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const { network } = useAuthenticationStore();
+  const { getClipboardText } = useClipboard();
+  const { themeColors } = useColors();
+  const { t } = useAppTranslation();
+  const [searchTerm, setSearchTerm] = useState("");
+  const [filteringTerm, setFilteringTerm] = useState("");
+
+  const debouncedUpdateFilteringTerm = useDebounce(() => {
+    setFilteringTerm(searchTerm);
+  }, 300);
```
CassioMG, line 40: should we move this 300 to a const?

src/components/screens/SwapScreen/hooks/useSwapPathFinding.ts
```diff
@@ -0,0 +1,67 @@
...
+        toBalance: swapToTokenBalance,
+        amount: swapAmount,
+        slippage: swapSlippage,
+        network,
+        publicKey,
+      });
+    } else {
+      clearPath();
+    }
+  }, 500);
```
CassioMG, line 53: should we move this 500 to a const?

src/components/screens/SwapScreen/screens/SwapAmountScreen.tsx
```diff
@@ -0,0 +1,441 @@
...
+    setFromToken,
+    setToToken,
+    setSwapAmount,
+    resetSwap,
+  } = useSwapStore();
+
+  const swapFromTokenBalance = balanceItems.find(
+    (item) => item.id === fromTokenId,
+  );
+  const swapToTokenBalance = balanceItems.find((item) => item.id === toTokenId);
```
CassioMG, line 86: could we wrap those 2 in `useMemo` to prevent excessive re-rendering?

src/components/screens/SwapScreen/screens/SwapAmountScreen.tsx
```diff
@@ -0,0 +1,441 @@
...
+        </ContextMenuButton>
+      ),
+    });
+  }, [navigation, menuActions, themeColors]);
+
+  const handleSelectSwapToToken = () => {
+    selectTokenBottomSheetModalRef.current?.present();
+  };
+
+  const handleTokenSelect = (tokenId: string, tokenSymbol: string) => {
```
CassioMG, line 232:
```suggestion
  const handleToTokenSelect = (tokenId: string, tokenSymbol: string) => {
```

src/components/screens/SwapScreen/helpers/swapTokenHelpers.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Calculates fiat amount for a token using multiple price sources
+ * This provides robust price calculation with fallbacks
+ */
+export const calculateTokenFiatAmount = (
+  token: AssetToken | NativeToken,
+  amount: string | BigNumber,
+  balanceItems: PricedBalance[],
+  prices?: TokenPricesMap,
```
CassioMG, line 89: could we please transform this into a single object?

src/components/screens/SwapScreen/helpers/swapTokenHelpers.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+};
+
+/**
+ * Finds a balance item that matches the given token using multiple strategies
+ * This is more robust than simple ID matching as it tries multiple approaches
+ */
+export const findBalanceForToken = (
+  token: AssetToken | NativeToken,
+  balanceItems: PricedBalance[],
```
CassioMG, line 31: single object

src/components/screens/SwapScreen/components/SwapReviewBottomSheet.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  const { fromTokenId, toTokenId } = useSwapStore();
+  const { balanceItems } = useBalancesList({
+    publicKey: account?.publicKey ?? "",
+    network,
+    shouldPoll: false,
+  });
+
+  const fromTokenBalance = balanceItems.find((item) => item.id === fromTokenId);
+  const toTokenBalance = balanceItems.find((item) => item.id === toTokenId);
```
CassioMG, line 88: wrap with `useMemo` to prevent excessive re-rendering

src/helpers/formatAmount.ts
danilosilvackl: I was trying to fix an issue with very small values, but after your comment i was incrementing the tests to make sure it was working, but it did not make much difference, so i rolled back the change

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -305,17 +308,20 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+                  adjustsFontSizeToFit
+                  numberOfLines={1}
+                  minimumFontScale={0.6}
                   {...(Number(tokenAmount) > 0
                     ? { primary: true }
                     : { secondary: true })}
                 >
-                  {tokenAmount}
+                  {tokenAmount}{" "}
+                  <RNText style={{ color: themeColors.text.secondary }}>
```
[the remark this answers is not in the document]
danilosilvackl: i needed to change just the color of the text, and our custom Text component enforces the Text size by default, so i used the RNText since it inherits the text size from the parent

src/components/BalancesList.tsx
danilosilvackl: This happens on bottom sheets, because they don't have a navigation object, this fix was specially made because we are using this BalancesList on the bottom sheet to select the token to swap to

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
danilosilvackl: yes, I had this fixed on the swap flow but forgot to update it here, fixed now!

src/components/screens/SwapScreen/helpers/swapCalculations.ts
danilosilvackl: I forgot to update it, I was supposed to get it from the `getNativeContractDetails`

src/components/BalancesList.tsx
```diff
@@ -153,7 +171,8 @@ export const BalancesList: React.FC<BalancesListProps> = ({
           />
         </NotificationWrapper>
 
-        {!isTestNetwork && (
+        {/* Only show fund account button if navigation is available and not test network */}
+        {!disableNavigation && !isTestNetwork && navigation && (
```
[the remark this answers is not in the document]
danilosilvackl: No, that will never happen, its just a fail safe for the navigation, since now it is nullable we must make sure its not null when its being used

removed it and switched for an optional chaining on line: 181

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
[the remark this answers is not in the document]
CassioMG: aaah got it, nice catch on using it for the dynamic sizing!

danilosilvackl:
@JakeUrban @CassioMG @sdfcharles 
latest version test on mainnet:
## XLM to USDC

https://github.com/user-attachments/assets/5e023ef5-faa0-4f49-9268-05742debf856

## XRP to SSLX

https://github.com/user-attachments/assets/43c0843a-16fc-4b11-ab01-79f000d0f807

CassioMG:
> @JakeUrban @CassioMG @sdfcharles latest version test on mainnet:
> 
> ## XLM to USDC
>  Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.14.36.18.mp4 
> ## XRP to SSLX
>  Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.16.01.10.mp4

@danilosilvackl nice improvements! 🚀 
Some comments:
- the numeric keyboard seems too far away from the review button at the bottom. Could we evenly space the number keys there so that it fulfills the whole area?
- it looks like the decimal places length is suddenly changing on some screens (e.g. jump from 4 decimal places to 7 decimal places)
- unrelated issue: on History, I see that the swap prices are not being shown. I believe Bruno left it like that because we don't have historical prices data yet. I think it'd look better if we just remove the prices label from there instead of always displaying the placeholder "-" label. Would you mind cleaning that up? (on the codebase we have 4 places marked with `TODO: priced amount`, I'd suggest removing all of them and double checking how the UI would look like without the "-")

danilosilvackl:
> > @JakeUrban @CassioMG
> > > I noticed the receive amount and the rate changes once the transaction is finalized. @sdfcharles what do you think about this? Should we remove the estimated receive amount from the pending screen and transaction details and only add the finalized amounts once we have them?
> > 
> > 
> > > > @JakeUrban @CassioMG @sdfcharles latest version test on mainnet:
> > > > ## XLM to USDC
> > > > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.14.36.18.mp4
> > > > ## XRP to SSLX
> > > > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.16.01.10.mp4
> > > 
> > > 
> > > @danilosilvackl nice improvements! 🚀 Some comments:
> > > 
> > > * the numeric keyboard seems too far away from the review button at the bottom. Could we evenly space the number keys there so that it fulfills the whole area?
> > > * it looks like the decimal places length is suddenly changing on some screens (e.g. jumping from 4 decimal places to 7 decimal places)
> > > * unrelated issue: on History, I see that the swap prices are not being shown. I believe Bruno left it like that because we don't have historical prices data yet. I think it'd look better if we just remove the prices label from there instead of always displaying the placeholder "-" label. Would you mind cleaning that up? (on the codebase we have 4 places marked with `TODO: priced amount`, I'd suggest removing all of them and double checking how the UI would look like without the "-")
> > 
> > 
> > i was working on some way to fix this and this was my train of thought so far: on the review bottom sheet:
> > 
> > * use only the data that came from the response of the horizon api when searching for the best path
> > 
> > on the processing screen:
> > 
> > * while its "processing" use the data that came from the api like on the review bottom sheet
> > * when it changes to "success" (transaction is done), get the final data from the transaction and show the updated values that will be consider the slippage and updated ratings
> > 
> > for the transaction details bottom sheet:
> > 
> > * use the same data from the transaction as the success/done state
> 
> @JakeUrban @danilosilvackl agreed its confusing that the amount on pending and success is different. ideally they are the same, is that what you are describing above danilo?
> 
> otherwise, an alternative simple solution would be to hide the amount on the pending screen, but prefer the other solution more. CleanShot 2025-06-20 at 13 59 41

@sdfcharles my idea would be during processing show the "predicted" amount and after its done we show the final amount, but it will still change and look weird, i think hiding would be cleaner on this case

danilosilvackl:
> > @JakeUrban @CassioMG @sdfcharles latest version test on mainnet:
> > ## XLM to USDC
> > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.14.36.18.mp4
> > ## XRP to SSLX
> > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.16.01.10.mp4
> 
> @danilosilvackl nice improvements! 🚀 Some comments:
> 
> * the numeric keyboard seems too far away from the review button at the bottom. Could we evenly space the number keys there so that it fulfills the whole area?
> * it looks like the decimal places length is suddenly changing on some screens (e.g. jumping from 4 decimal places to 7 decimal places)
> * unrelated issue: on History, I see that the swap prices are not being shown. I believe Bruno left it like that because we don't have historical prices data yet. I think it'd look better if we just remove the prices label from there instead of always displaying the placeholder "-" label. Would you mind cleaning that up? (on the codebase we have 4 places marked with `TODO: priced amount`, I'd suggest removing all of them and double checking how the UI would look like without the "-")

@CassioMG removed the pricing TODOs and the "-"
its looking fine:

CassioMG:
> > > @JakeUrban @CassioMG @sdfcharles latest version test on mainnet:
> > > ## XLM to USDC
> > > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.14.36.18.mp4
> > > ## XRP to SSLX
> > > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-06-19.at.16.01.10.mp4
> > 
> > 
> > @danilosilvackl nice improvements! 🚀 Some comments:
> > 
> > * the numeric keyboard seems too far away from the review button at the bottom. Could we evenly space the number keys there so that it fulfills the whole area?
> > * it looks like the decimal places length is suddenly changing on some screens (e.g. jumping from 4 decimal places to 7 decimal places)
> > * unrelated issue: on History, I see that the swap prices are not being shown. I believe Bruno left it like that because we don't have historical prices data yet. I think it'd look better if we just remove the prices label from there instead of always displaying the placeholder "-" label. Would you mind cleaning that up? (on the codebase we have 4 places marked with `TODO: priced amount`, I'd suggest removing all of them and double checking how the UI would look like without the "-")
> 
> @CassioMG removed the pricing TODOs and the "-" its looking fine: 

@danilosilvackl thanks for cleaning that up :)
I'd suggest vertically centering the amount labels with the icon on the right. Unless @sdfcharles prefers them top aligned as they are now:
457508639-8ceabc97-3929-40b0-b593-820e621cdd29

danilosilvackl:
> @danilosilvackl thanks for cleaning that up :)
> I'd suggest vertically centering the amount labels with the icon on the right. Unless @sdfcharles prefers them top aligned as they are now:

@sdfcharles @CassioMG for reference here it is with the price and asset icon centered, just let me know which one we should follow up with!

CassioMG:
> > @danilosilvackl thanks for cleaning that up :)
> > I'd suggest vertically centering the amount labels with the icon on the right. Unless @sdfcharles prefers them top aligned as they are now:
> 
> @sdfcharles @CassioMG for reference here it is with the price and asset icon centered, just let me know which one we should follow up with! 

I prefer the centered one, looks great :)

CassioMG (approved):

Issue #180, opened 2025-06-16
Connected dapps should be scoped to a specific account


JakeUrban:
### What version are you using?

MVP v2

### What did you do?

1. Connected my primary account to Blend via wallet connect
2. I switched to my secondary account
3. I initiated a transaction on the Blend dapp

### What did you expect to see?

No dapps connected in my dapps tab. I also expected to not get a request to sign with my secondary account.

### What did you see instead?

Blend was listed as a connected dapp. When I initiated the transaction on the blend dapp, I got a request to sign with my secondary account, even though I connected my primary account.

PR #194, opened 2025-06-25
Wallet Connect - Account Session Management

CassioMG:
Closes: #180 

This PR implements a proper `account session management` for the Wallet Connect integration so that it will:
- Only display the `available connections` for a given `account` and `network`
    - Which means you'll see a different list of "Connected Apps" for each account and network
- Only display `transaction sign requests` related to the currently `selected account`
    - Which means you won't ever be asked to sign a transaction from a "wrong" account
- Allow `disconnecting dApps` only from currently `selected account`
- `Clear` all existing `connections` from ALL addresses when users `sign-out`
- Display `informative error toasts` if:
  - You receive a Main Net request while the app is on Test Net network
  - You receive a request while the wallet is locked (not authenticated)
  - You receive a request from a dApp that does not support the Stellar chain

### Connecting multiple dApps to multiple accounts

[connecting-multiple-accounts.webm](https://github.com/user-attachments/assets/207d3ff9-e448-404f-b226-3154ac6b7b63)

### Switching between networks

https://github.com/user-attachments/assets/0a53efb7-e0b6-4f93-917a-787e4c5806cd

### Receiving a Main Net connection request while the app is on Test Net network

https://github.com/user-attachments/assets/9de6a4a1-8ce0-4faf-8aac-9a88cb3b240a

### Receiving a transaction request from account A while the app is on account B

https://github.com/user-attachments/assets/33037c2c-6d32-4565-b171-8c1362bddfe9

### Disconnecting dApps from current account only

https://github.com/user-attachments/assets/ae452a99-9928-443a-830c-c9bc50e09989

### Clear all connections from ALL addresses when signing-out

[clear-all-on-sign-out.webm](https://github.com/user-attachments/assets/abb30bbf-3e74-49c2-86cb-f74ac1d91f40)

### Receiving requests while app is locked (not authenticated)

https://github.com/user-attachments/assets/caeec939-e4fa-4d8b-8a63-2b6ac88a029d

### Receiving request from a dApp that does not support the Stellar chain

https://github.com/user-attachments/assets/cbc205fe-220c-40b0-a836-94933c8ec088


CassioMG:
> is it possible to clear connections for _ALL_ addresses, instead of just the current wallet? this is the UX I've seen on other apps and I think security wise would be preferable as well. wdyt?

@sdfcharles this is what the function is already doing on sign-out :) I've updated the PR description to make it more clear

@danilosilvackl @sdfcharles thanks for the feedback regarding the lock screen UX, I think that makes sense, I'll take a look here!

CassioMG:
@sdfcharles @danilosilvackl I've updated the "Wallet Locked" UX to the below. It'll now save the event and automatically display the bottom sheet after the wallet is unlocked as opposed to rejecting the event so that users can continue from there. It'll just discard the event in case the user is completely logged out on the welcome screen in which case the app will display a message asking the user to log in and try again as shown on the video.

### Receiving a connection request while wallet is locked

https://github.com/user-attachments/assets/6c42d1f6-b2f5-43f5-a520-0511e6a60cae

### Receiving a tx signature request while wallet is locked

https://github.com/user-attachments/assets/32aff6df-6d9c-4a70-a7af-de3bc2c8f31b

### Receiving a connection request while the user is completely logged out on welcome screen
*In this case it'll just reject the request and inform user to try again after logging in for simplicity

https://github.com/user-attachments/assets/360dc50a-14ed-461f-b361-7307fa30bdc2

PR #198, opened 2025-06-27
Fix responsiveness of top bar buttons

### LINKED ISSUE #192: Increase responsiveness of top bar buttons
It appears the touching action of the top bar buttons (e.g. "x", "back" and "info" buttons) are not working well.

Since going back using Android back button or iOS swipe gesture works well, it seems like it's probably a UI-only issue related to the top bar buttons.

It also appears this is worst on iOS, so maybe the issue is related to how close the buttons are from the status bar as on my Android device they are located lower when compared to my iPhone (it is just a wild guess).

Aside from fixing the responsiveness we should also ensure we have at least 40px of touching area for those buttons.

CassioMG:
Closes #192 

Included in this PR:

- `Fixes the responsiveness` of the top bar buttons by refactoring the existing `CustomNavigationHeader` component to ensure it'll add `enough padding below the status bar + notch area` of mobile devices. It also re-writes the `CustomNavigationHeader` using `nativewind` instead of styled components.

- Creates a reusable `CustomHeaderButton` component which should be used to render either the `left` or `right` buttons of the top bar ensuring consistency and lowering maintenance efforts.
   - The `left` button should usually be set in the navigator level as it'll usually be used to take navigation actions such as "go back" or "close". The `CustomNavigationHeader` component will render a `CustomHeaderButton` as the "back" arrow by default.
   - The `right` button should usually be set in the component level as it'll usually be used to present something over the current component. The `CustomHeaderButton` component will render a `HelpCircle` icon on the `right` by default since we use it in many screens to render the bottom sheet with more info.
  - We also have the flexibility of setting either button on navigator or component level if needed.

- Moves `CustomNavigationHeader`, `CustomHeaderButton` and `ScrollableKeyboardView` components to the layout folder.

- Re-writes the `HomeScreen` component using `nativewind`.

- Re-writes the `OfflineMessage` component using `nativewind` + fix the `font family`.

### Before fix, tap only works at the bottom of the button

https://github.com/user-attachments/assets/39d9d27c-d35e-494f-9ae3-46cf1bb60e5e

### After fix, tap works in a comfortable area around the button

https://github.com/user-attachments/assets/c9a9d356-176c-4efc-b41e-c735abe02ea3

### Offline banner with correct font family

Screenshot 2025-06-27 at 17 49 52


src/components/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -338,10 +288,16 @@ export const HomeScreen: React.FC<HomeScreenProps> = ({ navigation }) => {
...
+        </View>
+      </View>
+
+      <View
+        className="border-b mb-6 -ml-6"
+        style={{
+          width,
+          borderBottomColor: themeColors.border.primary,
+        }}
+      />
```
danilosilvackl, line 300:
```suggestion
      <View
        className="border-b mb-6 -ml-6 border-border-primary"
        style={{ width }}
      />
```

danilosilvackl (approved):
left a few suggestions, lgtm!

__tests__/layout/CustomHeaderButton.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+  describe("Custom onPress handlers", () => {
+    it("calls custom onPress when provided for left position", () => {
+      const mockOnPress = jest.fn();
+      const { getByTestId } = renderWithProviders(
+        <CustomHeaderButton position="left" onPress={mockOnPress} />,
+      );
+
+      const button = getByTestId("SvgMock").parent;
+      if (button) {
+        fireEvent.press(button);
```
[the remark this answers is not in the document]
CassioMG: https://github.com/stellar/freighter-mobile/pull/198/commits/49447135aa75eafe983235dcf730e34baab3352e

src/components/screens/HomeScreen/HomeScreen.tsx
CassioMG: nice catch! https://github.com/stellar/freighter-mobile/pull/198/commits/189b0998ec45edbe1efe1521fe0afe9ecd2eb759

Issue #157, opened 2025-06-08
Implementation - Frontend metrics


CassioMG:
Dependency: 

This task is about installing and configuring the analytics tool and also implementing some initial main metric calls according to the SPIKE results.

Issue #158, opened 2025-06-08
Settings - Metrics toggle


CassioMG:
We should have a toggle on Settings to let users enable/disable analytics tracking and ensure the analytics tool is respecting this toggle.

<img width="341" alt="Image" src="https://github.com/user-attachments/assets/abb1c004-00f6-42dc-ae20-72401e1bda62" />

PR #213, opened 2025-07-14
Freighter mobile Analytics

danilosilvackl:
This PR implements the analytics flow and all the related necessary code

## Analytics debug drawer
- this view is for development only, the aim is to help us when working with analytics
- it will show the current analytics state
- show the events log

### iOS 
|  |  |
| :--- | :--- |
| Home screen | Debug drawer |
### Android 🤖
|  |  |
| :--- | :--- |
| Home screen | Debug drawer |

### iOS 
|  |  |  |
| :--- | :--- |  :--- |
| Analytics disabled by the user | Analytics enabled but no api key set | Analytics disabled by the user and no api key set |
### Android 🤖
| Screenshot_1752504201 | Screenshot_1752504496 | Screenshot_1752504479 |
| :--- | :--- |  :--- |
| Analytics disabled by the user | Analytics enabled but no api key set | Analytics disabled by the user and no api key set |

## Metrics toggle
The user can opt out from the analytics using this screen
additionally we added the toggle component following the [stellar design system](https://design-system.stellar.org/toggle) existing component with some slight adjustments for mobile 
## Events on amplitude
Here we have a sample video of how we are tracking the events on Amplitude and some properties that we exclusively added for the more mobile devices information
<strong>For the full event list that are currently implemented, [click here](https://docs.google.com/document/d/19g4DGEsiSWR17utRfAetZzZt7O7z6AzeKACZMyf6imk/edit?usp=sharing)</strong>

https://github.com/user-attachments/assets/8ebd30bf-66a4-4789-a82c-d425eb574f6b

closes: #157 #158


src/services/analytics/constants.ts
```diff
@@ -0,0 +1,44 @@
...
+} as const;
+
+// -----------------------------------------------------------------------------
+// TIMING CONSTANTS
+// -----------------------------------------------------------------------------
+
+export const TIMING = {
+  THROTTLE_DELAY_MS: 500,
+  DEBUG_REFRESH_INTERVAL_MS: 2000,
+  QR_SCAN_TIMEOUT_MS: 30000,
```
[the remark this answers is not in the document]
danilosilvackl: actually not, i used them while making the debug drawer, but it was removed during iterations, thanks for noticing!

src/config/analyticsEvents.ts
```diff
@@ -0,0 +1,125 @@
...
+  VIEW_VALIDATE_RECOVERY_PHRASE = "loaded screen: confirm mnemonic phrase",
+  VIEW_IMPORT_WALLET = "loaded screen: recover account",
+  VIEW_LOCK_SCREEN = "loaded screen: unlock account",
+  VIEW_HOME = "loaded screen: account",
+  VIEW_HISTORY = "loaded screen: account history",
+  VIEW_DISCOVERY = "loaded screen: discover",
+  VIEW_TOKEN_DETAILS = "loaded screen: asset detail",
+  VIEW_ACCOUNT_QR_CODE = "loaded screen: view public key generator",
+  VIEW_GRANT_ACCESS = "loaded screen: grant access",
+  VIEW_SIGN_TRANSACTION = "loaded screen: sign transaction",
```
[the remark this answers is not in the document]
danilosilvackl: This falls in the same pitfall of the other one that I answered [here](https://github.com/stellar/freighter-mobile/pull/213#discussion_r2205491941)

package.json
```diff
@@ -41,9 +41,10 @@
     ]
   },
   "dependencies": {
+    "@amplitude/analytics-react-native": "^1.4.13",
     "@babel/runtime": "^7.26.10",
     "@gorhom/bottom-sheet": "^5.1.2",
-    "@react-native-async-storage/async-storage": "^2.1.2",
+    "@react-native-async-storage/async-storage": "^2.2.0",
```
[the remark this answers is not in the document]
CassioMG: nevermind I misread it 😅

src/config/analyticsEvents.ts
```diff
@@ -0,0 +1,125 @@
...
+  ACCOUNT_SCREEN_ADD_ACCOUNT = "account screen: created new account",
+  ACCOUNT_SCREEN_COPY_PUBLIC_KEY = "account screen: copied public key",
+  ACCOUNT_SCREEN_IMPORT_ACCOUNT = "account screen: imported new account",
+  ACCOUNT_SCREEN_IMPORT_ACCOUNT_FAIL = "account screen: imported new account: error",
+  VIEW_PUBLIC_KEY_ACCOUNT_RENAMED = "viewPublicKey: renamed account",
+  VIEW_PUBLIC_KEY_CLICKED_STELLAR_EXPERT = "viewPublicKey: clicked StellarExpert",
+
+  // WalletConnect/dApp Events
+  GRANT_ACCESS_SUCCESS = "grant access: granted",
+  GRANT_ACCESS_FAIL = "grant access: rejected",
```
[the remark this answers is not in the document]
danilosilvackl: great! I did update the constants though, at least during development it will be clear whats the event is related to

JakeUrban:
Can we update the copy on the metrics toggle? We will collect public keys, transaction amounts, and balances. We will never collect PII like name/address.

src/components/App.tsx
```diff
@@ -13,16 +14,38 @@ import { ToastProvider } from "providers/ToastProvider";
...
 export const navigationRef = createNavigationContainerRef<RootStackParamList>();
 
 export const App = (): React.JSX.Element => {
+  const { onStateChange } = useNavigationAnalytics();
+
   useEffect(() => {
     Appearance.setColorScheme("dark");
+    initAnalytics();
+
+    analytics.trackAppOpened();
```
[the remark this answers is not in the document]
CassioMG: or even use different event names in case that's not disruptive with extension's events

src/components/screens/AddAssetScreen/AddAssetScreen.tsx
```diff
@@ -86,6 +88,8 @@ const AddAssetScreen: React.FC<AddAssetScreenProps> = () => {
       return;
     }
 
+    analytics.trackAddTokenConfirmed(selectedAsset.assetCode);
```
CassioMG, line 91: this is unrelated to this PR but I just noticed the function here is called handleAddAssetTrustline although I believe it's being used to add both Classic Trustlines and Custom Tokens. Would you mind doing a quick check and rename the function name so it's not misleading? thanks!

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -113,9 +115,22 @@ export const useSwapTransaction = ({
       if (!transactionHash) {
         throw new Error("Failed to submit transaction");
       }
+
+      analytics.trackSwapSuccess({
+        sourceAsset: sourceBalance?.tokenCode || NATIVE_TOKEN_CODE,
+        destAsset: destinationBalance?.tokenCode || NATIVE_TOKEN_CODE,
```
CassioMG, line 121: Should we fallback to something more explicit here instead of NATIVE_TOKEN_CODE? E.g. "Unknown Asset"

I'm afraid it could potentially pollute our metrics with misleading data.

Or maybe we could throw an error in case it doesn't make sense to execute a swap without the source or destination assets. I see that the buildSwapTransaction requires them.

src/config/analyticsRoutes.ts
```diff
@@ -0,0 +1,271 @@
...
+  [SEND_PAYMENT_ROUTES.TRANSACTION_TIMEOUT_SCREEN]:
+    "loaded screen: send payment timeout" as AnalyticsEvent,
+  [SETTINGS_ROUTES.CHANGE_NETWORK_SCREEN]:
+    "loaded screen: manage network" as AnalyticsEvent,
+  [SETTINGS_ROUTES.NETWORK_SETTINGS_SCREEN]:
+    "loaded screen: network settings" as AnalyticsEvent,
+  [SETTINGS_ROUTES.SHARE_FEEDBACK_SCREEN]:
+    "loaded screen: leave feedback" as AnalyticsEvent,
+  [SETTINGS_ROUTES.SHOW_RECOVERY_PHRASE_SCREEN]:
+    "loaded screen: show recovery phrase" as AnalyticsEvent,
```
CassioMG, line 83: automatic?

src/config/analyticsRoutes.ts
```diff
@@ -0,0 +1,271 @@
...
+    ROOT_NAVIGATOR_ROUTES,
+    AUTH_STACK_ROUTES,
+    MAIN_TAB_ROUTES,
+    MANAGE_ASSETS_ROUTES,
+    SETTINGS_ROUTES,
+    MANAGE_WALLETS_ROUTES,
+    BUY_XLM_ROUTES,
+    SEND_PAYMENT_ROUTES,
+    SWAP_ROUTES,
+  ] as const;
```
CassioMG, line 128: would be nice if we could find a way of raising a lint error here or something that would let devs know in case some route object is missing here. I'm thinking about the future when we start creating new routes, we'd need to remember to add them here as well. Wdyt?

src/config/analyticsRoutes.ts
```diff
@@ -0,0 +1,271 @@
...
+ * ## How it works:
+ * 1. **Automatic**: "WelcomeScreen" → "loaded screen: welcome"
+ * 2. **Automatic**: "SettingsScreen" → "loaded screen: settings"
+ * 3. **Override only**: Special cases that don't follow the pattern
+ *
+ * ## Adding new screens:
+ * 1. Add route to routes.ts (normal process)
+ * 2. Analytics tracking works automatically ✨
+ * 3. Only add override if the auto-generated name is wrong
+ */
```
CassioMG, line 29: I like that we are automating the route events based on the route name, but this also means we should be more mindful from now on when editing route names. Wdyt?
Would you mind adding comments on the `routes.ts` file to make it very explicit that the route name is also being used for analytics? So we prevent analytics disruption

src/config/analyticsRoutes.ts
```diff
@@ -0,0 +1,271 @@
...
+const ROUTES_WITHOUT_ANALYTICS = new Set<string>([
+  // Navigator-level routes (not actual screens)
+  "MainTabStack",
+  "AuthStack",
+  "SettingsStack",
+  "SendPaymentStack",
+  "SwapStack",
+  "ManageAssetsStack",
+  "ManageWalletsStack",
+  "BuyXLMStack",
```
CassioMG, line 41: should we use a filter by "Stack" string on `ROOT_NAVIGATOR_ROUTES` values here so we automate this `ROUTES_WITHOUT_ANALYTICS` Set?

src/config/analyticsEvents.ts
```diff
@@ -0,0 +1,124 @@
...
+  VIEW_BUY_XLM = "loaded screen: add fund",
+  VIEW_SEARCH_ASSET = "loaded screen: search asset",
+  VIEW_ADD_ASSET_MANUALLY = "loaded screen: add asset manually",
+
+  // User Action Events
+  CREATE_PASSWORD_SUCCESS = "account creator: create password: success",
+  CREATE_PASSWORD_FAIL = "account creator: create password: error",
+  VIEWED_RECOVERY_PHRASE = "account creator: viewed phrase",
+  CONFIRM_RECOVERY_PHRASE_SUCCESS = "account creator: confirm phrase: confirmed phrase",
+  CONFIRM_RECOVERY_PHRASE_FAIL = "account creator: confirm phrase: error confirming",
```
CassioMG, line 58: it seems we are missing the below here, is it because we don't have the screens or the action on mobile yet?
```
  accountCreatorConfirmMnemonicBack:
    "account creator: confirm phrase: back to phrase",
  accountCreatorFinished:
    "account creator finished: closed account creator flow",
```

src/ducks/networkInfo.ts
```diff
@@ -1,25 +1,36 @@
...
+    set(() => ({
+      isConnected: payload.isConnected,
+      isInternetReachable: payload.isInternetReachable,
+      isOffline: !payload.isConnected || !payload.isInternetReachable,
+      connectionType: payload.type,
+      effectiveType:
+        payload.type === NetInfoStateType.cellular &&
+        payload.details?.cellularGeneration
+          ? payload.details.cellularGeneration
+          : null,
```
CassioMG, line 34: just out of curiosity, is this fixing some network detection behavior?

src/helpers/date.ts
```diff
@@ -67,3 +67,21 @@ export const formatTransactionDate = (
 
   return `${formattedDate} · ${formattedTime}`;
 };
+
+/**
+ * Formats a timestamp as relative time (e.g., "just now", "5m ago", "2h ago")
```
CassioMG, line 72:
```suggestion
 * Formats a timestamp as relative time (e.g., "just now", "5m ago", "2h ago")
 * It doesn't need to be translated as it's only being used for DEBUGGING info
```

src/config/analyticsEvents.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  // Transaction & Simulation Events
+  SIMULATE_TOKEN_PAYMENT_ERROR = "failed to simulate token payment",
+  SIGN_TRANSACTION_SUCCESS = "sign transaction: confirmed",
+  SIGN_TRANSACTION_FAIL = "sign transaction: rejected",
+  SIGN_TRANSACTION_MEMO_REQUIRED_FAIL = "sign transaction: memo required error",
+
+  // Asset Management Events
+  ADD_ASSET_SUCCESS = "manage asset: add asset",
+  ADD_TOKEN_SUCCESS = "manage asset: add token",
```
CassioMG, line 90: nite: looks like this `manage asset: add toke` event is not being used on both mobile and extension - cc @danilosilvackl @aristidesstaffieri

src/providers/AuthCheckProvider.tsx
```diff
@@ -14,6 +16,14 @@ export const AuthCheckProvider: React.FC<AuthCheckProviderProps> = ({
   children,
 }) => {
   const { panHandlers } = useAuthCheck();
+  const { account } = useAuthenticationStore();
+
+  useEffect(() => {
+    // Checking for public key to make sure the account is loaded
+    if (account?.publicKey) {
+      analytics.identifyUser();
```
CassioMG, line 24: is it important to wait for the publicKey here since the analytics userId is not tied to it? is the extension also waiting for the publicKey to identify users?

src/providers/NetworkProvider.tsx
```diff
@@ -19,7 +19,7 @@ export const NetworkProvider = ({ children }: Props) => {
 
     // Initial network check to consume initial "null" values
     // and avoid UI glitches due to initial network state
-    NetInfo.fetch();
+    NetInfo.fetch().then(setNetworkInfo);
```
CassioMG, line 22: given the above comment on line 20 it seems this call was meant to be here just to consume the initial "null" values, are we sure we should be calling `setNetworkInfo` here?

src/providers/WalletKitProvider.tsx
```diff
@@ -162,6 +164,10 @@ export const WalletKitProvider: React.FC<WalletKitProviderProps> = ({
 
     setIsConnecting(true);
 
+    analytics.trackGrantAccessSuccess(
+      proposalEvent.params.proposer.metadata.url,
+    );
```
CassioMG, line 169: looks like we are not calling `trackGrantAccessFail` anywhere

src/services/analytics/constants.ts
```diff
@@ -0,0 +1,42 @@
...
+  LOG_PREFIX: "Analytics",
+} as const;
+
+// -----------------------------------------------------------------------------
+// ANALYTICS CONFIGURATION
+// -----------------------------------------------------------------------------
+
+export const ANALYTICS_CONFIG = {
+  DEFAULT_ENABLED: false,
+  INCLUDE_PLATFORM_CONTEXT: true,
```
CassioMG, line 40:
```suggestion
  INCLUDE_COMMON_CONTEXT: true,
```

I'd suggest naming it like this as it includes more than just platform info

src/services/analytics/core.ts
```diff
@@ -0,0 +1,238 @@
...
+    amplitude.init(AMPLITUDE_API_KEY);
+
+    // Get initial state
+    const { isEnabled } = useAnalyticsStore.getState();
+    amplitude.setOptOut(!isEnabled);
+
+    logger.debug(
+      DEBUG_CONFIG.LOG_PREFIX,
+      `Initialized with key: ${AMPLITUDE_API_KEY.slice(0, 8)}...`,
+    );
```
CassioMG, line 53: I think we should remove this log to prevent exposing the API key in any way, even if it's truncated

src/services/analytics/debug.ts
```diff
@@ -0,0 +1,73 @@
...
+  // 3. Analytics has been initialized
+  const isSendingToAmplitude = hasApiKey && isEnabled && hasInit;
+
+  return {
+    isEnabled,
+    userId,
+    hasInitialized: hasInit,
+    environment: __DEV__ ? "development" : "production",
+    amplitudeKey: AMPLITUDE_API_KEY
+      ? `${AMPLITUDE_API_KEY.slice(0, 6)}...`
```
CassioMG, line 61: same here, I think we should avoid displaying part of the API key

src/services/analytics/core.ts
```diff
@@ -0,0 +1,238 @@
...
+export const trackAppOpened = (): void => {
+  track(AnalyticsEvent.APP_OPENED);
+};
+
+// -----------------------------------------------------------------------------
+// STORE SUBSCRIPTION (Must be after track function is defined)
+// -----------------------------------------------------------------------------
+
+// Set up analytics store subscription to handle user toggling analytics
+useAnalyticsStore.subscribe((state) => {
```
CassioMG, line 223: this is great, I didn't know we could easily subscribe to store changes like that

src/services/analytics/user.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Handles graceful fallbacks if analytics is disabled or not initialized.
+ */
+export const identifyUser = async (): Promise<void> => {
+  const {
+    isEnabled,
+    setUserId,
+    userId: currentUserId,
+  } = useAnalyticsStore.getState();
+
+  if (!isEnabled) {
```
CassioMG, line 95: should we also check for `isInitialized()` here?

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+  TransactionSuccessEvent,
+  SwapSuccessEvent,
+  TransactionErrorEvent,
+} from "services/analytics/types";
+
+// -----------------------------------------------------------------------------
+// TRANSACTION ANALYTICS
+// -----------------------------------------------------------------------------
+
+export const trackSignedTransaction = (data: SignedTransactionEvent): void => {
```
CassioMG, line 14: should this one be used when signing dApp transactions through WC?

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+export const trackSendPaymentSuccess = (
+  data: TransactionSuccessEvent,
+): void => {
+  track(AnalyticsEvent.SEND_PAYMENT_SUCCESS, {
+    sourceAsset: data.sourceAsset,
+    transactionType: data.transactionType,
+  });
+};
+
+export const trackSendPaymentPathPaymentSuccess = (
```
CassioMG, line 41: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+    transactionType: data.transactionType,
+    isSwap: data.isSwap,
+  });
+};
+
+export const trackSendPaymentSetMax = (): void => {
+  track(AnalyticsEvent.SEND_PAYMENT_SET_MAX);
+};
+
+export const trackSendPaymentTypeSelected = (
```
CassioMG, line 78: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+): void => {
+  const event =
+    paymentType === "payment"
+      ? AnalyticsEvent.SEND_PAYMENT_TYPE_PAYMENT
+      : AnalyticsEvent.SEND_PAYMENT_TYPE_PATH_PAYMENT;
+
+  track(event, { paymentType });
+};
+
+export const trackSendPaymentRecentAddress = (): void => {
```
CassioMG, line 89: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+
+// -----------------------------------------------------------------------------
+// ASSET MANAGEMENT ANALYTICS
+// -----------------------------------------------------------------------------
+
+export const trackAddTokenConfirmed = (asset?: string): void => {
+  track(AnalyticsEvent.ADD_TOKEN_CONFIRMED, { asset });
+};
+
+export const trackAddTokenRejected = (asset?: string): void => {
```
CassioMG, line 101: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+
+// -----------------------------------------------------------------------------
+// ACCOUNT MANAGEMENT ANALYTICS
+// -----------------------------------------------------------------------------
+
+export const trackAccountScreenAddAccount = (): void => {
+  track(AnalyticsEvent.ACCOUNT_SCREEN_ADD_ACCOUNT);
+};
+
+export const trackAccountScreenCopyPublicKey = (): void => {
```
CassioMG, line 117: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+  track(AnalyticsEvent.ACCOUNT_SCREEN_COPY_PUBLIC_KEY);
+};
+
+export const trackAccountScreenImportAccount = (): void => {
+  track(AnalyticsEvent.ACCOUNT_SCREEN_IMPORT_ACCOUNT);
+};
+
+export const trackAccountScreenImportAccountFail = (error: string): void => {
```
CassioMG, line 125: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+  oldName: string,
+  newName: string,
+): void => {
+  track(AnalyticsEvent.VIEW_PUBLIC_KEY_ACCOUNT_RENAMED, {
+    oldName,
+    newName,
+  });
+};
+
+export const trackViewPublicKeyClickedStellarExpert = (): void => {
```
CassioMG, line 139: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+  reason?: string,
+): void => {
+  track(AnalyticsEvent.GRANT_DAPP_ACCESS_FAIL, { domain, reason });
+};
+
+// -----------------------------------------------------------------------------
+// HISTORY ANALYTICS
+// -----------------------------------------------------------------------------
+
+export const trackHistoryOpenFullHistory = (): void => {
```
CassioMG, line 162: not being used

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+  action: string,
+): void => {
+  track(event, { context, action });
+};
+
+export const trackCopyPublicKey = (): void => {
+  trackUserAction(AnalyticsEvent.COPY_PUBLIC_KEY, "home_screen", "copy");
+};
+
+export const trackCopyBackupPhrase = (): void => {
```
CassioMG, line 215: not being used

CassioMG:
@danilosilvackl the overall implementation looks good to me! Good job with the structuring and route logging automatization. Left some improvement suggestions and I've also marked all events that I saw were not being called.

src/ducks/networkInfo.ts
danilosilvackl: It's adding new props, we weren't getting the `connectionType` and `effectiveType` before
we use those two to send to analytics the data like: 
([connectionType: wifi, effectiveType: null] or [connectionType: cellular, effectiveType: 5g])

src/services/analytics/user.ts
danilosilvackl: good point, added it

src/i18n/locales/en/translations.json
```diff
@@ -52,12 +52,19 @@
     "title": "Settings",
     "logout": "Log Out",
     "network": "Network",
+    "preferences": "Preferences",
```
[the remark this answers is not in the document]
danilosilvackl: added the missing ones

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
danilosilvackl: you're right, we are now throwing the error before this so we won't have this misleading data

src/config/analyticsRoutes.ts
```diff
@@ -0,0 +1,271 @@
...
+  [SEND_PAYMENT_ROUTES.TRANSACTION_MEMO_SCREEN]:
+    "loaded screen: send payment settings" as AnalyticsEvent,
+  [SEND_PAYMENT_ROUTES.TRANSACTION_FEE_SCREEN]:
+    "loaded screen: send payment fee" as AnalyticsEvent,
+  [SEND_PAYMENT_ROUTES.TRANSACTION_TIMEOUT_SCREEN]:
+    "loaded screen: send payment timeout" as AnalyticsEvent,
+  [SETTINGS_ROUTES.CHANGE_NETWORK_SCREEN]:
+    "loaded screen: manage network" as AnalyticsEvent,
+  [SETTINGS_ROUTES.NETWORK_SETTINGS_SCREEN]:
+    "loaded screen: network settings" as AnalyticsEvent,
```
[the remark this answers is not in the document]
danilosilvackl: yes! I also cleaned the other that should work out of the box

src/config/analyticsRoutes.ts
danilosilvackl: good idea, this made things much easier

src/config/analyticsEvents.ts
danilosilvackl: We did have them, but I missed their implementation before, its fixed now

src/services/analytics/core.ts
danilosilvackl: good point, removed any UI mention of it

src/services/analytics/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+
+export const trackSignedTransaction = (data: SignedTransactionEvent): void => {
+  track(AnalyticsEvent.SIGN_TRANSACTION_SUCCESS, {
+    transactionHash: data.transactionHash,
+    transactionType: data.transactionType,
+    network: data.network,
+  });
+};
+
+export const trackSimulationError = (
```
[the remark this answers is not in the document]
danilosilvackl: for this and the other events, I checked all of them, some were removed and others simplified, but now it should be all correct triggering and otherwise, removed

src/services/analytics/user.ts
```diff
@@ -0,0 +1,137 @@
...
+// -----------------------------------------------------------------------------
+
+// Track last identified user to prevent duplicate identification
+let lastIdentifiedUserId: string | null = null;
+
+/**
+ * Identifies the user with analytics providers.
+ * Handles graceful fallbacks if analytics is disabled or not initialized.
+ */
+export const identifyUser = async (): Promise<void> => {
```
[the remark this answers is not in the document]
CassioMG: let's keep per device for now

src/providers/AuthCheckProvider.tsx
CassioMG: I see we are still using `account` as dep here. Is that really necessary?
I'm thinking about a few downsides:
- I'm assuming it'll only start sending events after user has logged in, which means we'd miss events like the recovery phrase creation and verification. Is that the case?
- It'll call `identifyUser()` again every time the user switches between its wallet accounts since it'll modify the `account` dependency on this `useEffect`
- Is `AuthCheckProvider` the best place to call `identifyUser()` considering it shouldn't rely on auth statuses? Would `RootNavigator` be a good place?

src/helpers/walletKitUtil.ts
```diff
@@ -273,18 +313,31 @@ export const approveSessionRequest = async ({
...
     );
     signedTransaction = signTransaction(transaction);
+
+    // Look up the dApp domain from the active session to send on analytics
+    const activeSessions = getActiveSessions(
+      publicKey,
+      activeChain as NETWORKS,
+    );
+    const session = activeSessions[topic];
+    const dappDomain = session?.peer?.metadata?.url;
```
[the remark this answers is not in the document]
danilosilvackl: thanks for the insight, i just updated it, made some modifications to the hook so it can be used both on react style components and on helpers such as the walletKitUtil, now we get this: 
Screenshot 2025-07-21 at 18 16 07
Screenshot 2025-07-21 at 18 16 26

src/helpers/walletKitUtil.ts
```diff
@@ -273,18 +313,31 @@ export const approveSessionRequest = async ({
...
+
+    analytics.trackSignedTransaction({
+      transactionHash: transaction.hash().toString("hex"),
+      transactionType: TransactionType.Classic,
+      network: activeChain,
```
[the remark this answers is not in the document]
danilosilvackl: perfect, just removed it!

src/hooks/useDappMetadata.ts
```diff
@@ -1,22 +1,88 @@
...
   description: "",
   url: "",
   icons: [],
 };
 
+/**
+ * Pure utility to extract dApp metadata from a WalletKit event or session request.
+ */
+export function extractDappMetadata(
+  eventOrRequest: WalletKitEvent | WalletKitSessionRequest | null,
```
CassioMG, line 25:
```export type WalletKitEvent =
  | WalletKitSessionProposal
  | WalletKitSessionRequest
  | typeof noneEvent;
```

considering this, I believe this function declaration could be something like this:

`export function extractDappMetadata(event: WalletKitEvent  | null)`

does that make sense?

src/hooks/useDappMetadata.ts
```diff
@@ -40,51 +106,5 @@ const emptyMetadata: CoreTypes.Metadata = {
...
-    }
-
-    return emptyMetadata;
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [event, activeSessionsKey]);
-
-  return dappMetadata;
-};
+): CoreTypes.Metadata | null =>
+  useMemo(() => extractDappMetadata(event), [event]);
```
CassioMG, line 110: I think it's missing the `activeSessionsKey` as dep here, it's important things get updated whenever there is a change on the active sessions

src/hooks/useDappMetadata.ts
```diff
@@ -1,22 +1,88 @@
...
   icons: [],
 };
 
+/**
+ * Pure utility to extract dApp metadata from a WalletKit event or session request.
+ */
+export function extractDappMetadata(
+  eventOrRequest: WalletKitEvent | WalletKitSessionRequest | null,
+): CoreTypes.Metadata {
+  if (!eventOrRequest) return emptyMetadata;
```
CassioMG: I think it's important to return `null` here instead of `emptyMetadata` due to some checks we are doing on the Dapp bottom sheets which only works when this returns `null` (or `undefined`) instead of an object

src/hooks/useDappMetadata.ts
```diff
@@ -1,22 +1,78 @@
...
+    return null;
+  }
+
+  if (event.type === WalletKitEventTypes.NONE) {
+    return emptyMetadata;
+  }
+
+  if (event.type === WalletKitEventTypes.SESSION_PROPOSAL) {
+    return "params" in event
+      ? (event.params?.proposer?.metadata ?? emptyMetadata)
```
CassioMG, line 40:
```suggestion
      ? event.params?.proposer?.metadata
```

src/hooks/useDappMetadata.ts
```diff
@@ -1,22 +1,78 @@
...
+      if (
+        matchedSessionByTopic &&
+        matchedSessionByTopic.peer &&
+        matchedSessionByTopic.peer.metadata
+      ) {
+        return matchedSessionByTopic.peer.metadata;
+      }
+    }
+
+    return emptyMetadata;
```
CassioMG, line 70:
```suggestion
      // It looks like the event.topic value for session request events is related
      // to the session key on activeSessions so let's try to use that first
      const matchedSessionByKey =
        "topic" in event ? activeSessions[event.topic] : null;
      if (matchedSessionByKey) {
        return matchedSessionByKey.peer?.metadata || emptyMetadata;
      }

      // If we don't have a match by session key, let's try to find a match by session topic
      const matchedSessionByTopic = Object.values(activeSessions).find(
        (session) => "topic" in event && event.topic === session.topic,
      );

      return matchedSessionByTopic?.peer?.metadata || emptyMetadata;
```

src/hooks/useDappMetadata.ts
```diff
@@ -1,22 +1,78 @@
...
 
+/**
+ * Pure utility to extract dApp metadata from a WalletKit event and activeSessions map.
+ */
+export function extractDappMetadata(
+  event: WalletKitEvent | null,
+  activeSessions: Record<
+    string,
+    { peer?: { metadata?: CoreTypes.Metadata }; topic?: string }
+  > = {},
```
CassioMG, line 28:
```suggestion
  activeSessions: ActiveSessions,
```

src/helpers/walletKitUtil.ts
```diff
@@ -273,18 +275,29 @@ export const approveSessionRequest = async ({
...
+    const { activeSessions } = useWalletKitStore.getState();
+    const dappMetadata = getDappMetadataFromEvent(
+      sessionRequest,
+      activeSessions,
+    );
+    const dappDomain = dappMetadata?.url;
+
+    analytics.trackSignedTransaction({
+      transactionHash: transaction.hash().toString("hex"),
+      transactionType: TransactionType.Classic,
```
CassioMG, line 288: will it always be a Classic transaction here?

CassioMG: let's not send the `transactionType` for now, we can infer they are from dApps since this is the only place logging this event

danilosilvackl: removed it for now

CassioMG (approved):
LGTM

PR #224, opened 2025-07-21
Connect In-App Browser to Discover endpoint

CassioMG:


JakeUrban:
Also @sdfcharles can we consider using "Popular" instead of "Trending"? Trending implies those dapps will change frequently, whereas popular is more steady

PR #230, opened 2025-07-24
Added permission request for iOS

danilosilvackl:
Work done here: 

- Implemented new App Tracking Transparency permission request for iOS
- fixed no network connection visual bug
- added `InfoPlist.strings` file to localize system messages
- Added new hook `useAnalyticsAndPermissions` to handle permission request and analytics initialization on the correct lifecycle of the app
- Added privacy policy link on the welcome screen

> [!IMPORTANT]  
> For iOS the analytics will always start as turned off
> the user should accept this permission request first to start tracking
> or they can use our settings page afterwards to opt in

### Opt out

https://github.com/user-attachments/assets/9c5016f4-1337-446c-8765-e56d1c6cab25

### Opt in

https://github.com/user-attachments/assets/6dba6c3d-ae72-45f3-a8bc-a3b9363b6043

### Enable Analytics

https://github.com/user-attachments/assets/0195d73b-ea01-4a19-bfde-1ad8c4d80f97

### Disable Analytics

https://github.com/user-attachments/assets/feed8e44-eb5e-447e-8fe4-6a2464fdb22e


ios/freighter-mobile.xcodeproj/project.pbxproj
```diff
@@ -174,6 +178,7 @@
 			knownRegions = (
 				en,
 				Base,
+				"pt-BR",
```
CassioMG, line 181: Let's use the more generic `pt` language to match what we have on the `i18n` translations files for consistency

ios/Podfile
```diff
@@ -1,13 +1,43 @@
...
+
+node_require('react-native/scripts/react_native_pods.rb')
+node_require('react-native-permissions/scripts/setup.rb')
 
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
+# listed all available permissions
+# uncomment as needed
+setup_permissions([
```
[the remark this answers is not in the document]
CassioMG: @danilosilvackl just tested on my iPhone, the camera permission looks fine.

I noticed one thing though related to the Analytics permission. Seems like Freighter's switch is not interacting with the actual iOS device settings.

iOS has this "Allow Tracking" switch on the device settings for each app, I think for this to work fine both should always be in sync. I am not sure if RN API will let you change the device settings, I believe the API is probably just for requesting or reading the authorization. If that's the case, then you'll need to call the `openSettings` function to navigate users to the device settings so they can change it there. Then after it's changed there the switch on Freighter settings should react to it and also change its state.

I'd suggest checking how the UX looks like in other wallets, I suspect they may let users only change this tracking authorization through device settings instead of in-app.

I'm not sure if Android also has a "Allow Tracking" device settings, but in case they do then it should work the same, always in sync.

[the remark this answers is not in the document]
danilosilvackl: Added the new ways to handle it now

src/components/screens/WelcomeScreen.tsx
```diff
@@ -75,6 +75,10 @@ export const WelcomeScreen: React.FC<WelcomeScreenProps> = ({ navigation }) => {
             <Text md medium url="https://stellar.org/terms-of-service">
               {t("welcomeScreen.terms.termsOfService")}
             </Text>
+            {t("common.and")}
+            <Text md medium url="https://stellar.org/privacy-policy">
```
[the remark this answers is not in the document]
danilosilvackl: done!

ios/freighter-mobile.xcodeproj/project.pbxproj
CassioMG: it's a bit hidden, XCode shows the most common ones, but you can find the generic language when going to "More Languages":
Screenshot 2025-07-25 at 17 31 45

ios/Podfile.lock
```diff
@@ -2502,4 +2502,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: be4b4636fdb0c60685819016cdb32a2691f91d13
 
-COCOAPODS: 1.16.2
+COCOAPODS: 1.15.2
```
CassioMG, line 2505: do we need to downgrade cocoapods?

src/components/screens/WelcomeScreen.tsx
```diff
@@ -72,9 +73,21 @@ export const WelcomeScreen: React.FC<WelcomeScreenProps> = ({ navigation }) => {
         <View>
           <StyledTermsText md secondary medium>
             {t("welcomeScreen.terms.byProceeding")}
-            <Text md medium url="https://stellar.org/terms-of-service">
+            <Text
+              md
+              medium
+              url={`${STELLAR_FOUNDATION_BASE_URL}/terms-of-service`}
```
CassioMG, line 79: I think it'd be better to export the whole url as it'll be used in more than one place: WelcomeScreen and About screen on settings

src/hooks/useAnalyticsAndPermissions.ts
```diff
@@ -0,0 +1,224 @@
...
+      return;
+    }
+
+    if (result === RESULTS.DENIED && !attRequested) {
+      await requestTrackingPermission();
+
+      setIsPermissionLoading(false);
+    } else {
+      setIsPermissionLoading(false);
+    }
```
CassioMG, line 158:
```suggestion
    if (result === RESULTS.DENIED && !attRequested) {
      await requestTrackingPermission();
    }
    
    setIsPermissionLoading(false);
```

ios/freighter-mobile.xcodeproj/project.pbxproj
danilosilvackl: Oh, awesome didn't knew about it, added now!

src/hooks/useAnalyticsAndPermissions.ts
```diff
@@ -15,39 +16,169 @@ const shouldSetAttRequested = (result: string) =>
...
+   * Handles the analytics toggle click.
+   * - On iOS 14+, if disabling analytics, also opens device settings for ATT.
+   * - On Android or iOS < 14, only disables analytics in-app.
+   * - Handles all permission request/enable/disable flows.
+   */
+  const handleAnalyticsToggleClick = useCallback(async () => {
+    setIsPermissionLoading(true);
+
+    if (isAttSupported && isEnabled) {
+      await Linking.openSettings();
```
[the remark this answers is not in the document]
danilosilvackl: good point, the `openSettings` does the same thing as the `Linking.openSettings`, but it would be really helpful in other cases that are explicitly said on the documentation.

also about the suggested flow for authentication, I made some modifications to follow the graph, but I think we will need @sdfcharles ands Phil help to improve the texts on the screen.

here is a small video of how it works with the suggested flow:
### turning analytics off

https://github.com/user-attachments/assets/14e4b2d7-2cd1-4147-9ac2-fac654fd0596

### turning analytics on

https://github.com/user-attachments/assets/e8dbaffc-ebf8-4a97-972d-10343c100eaf

danilosilvackl:
PR is going to be merged right now, but we will have some updates to the copy and interface in the near future

PR #235, opened 2025-07-30
Swap amount screen improvements

danilosilvackl:
- Improved responsiveness of the layout
- Added new method to validate the XLM balance when swapping **_TO_** XLM
- Added user friendly error message

Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 18 52 42


src/components/screens/SwapScreen/screens/SwapAmountScreen.tsx
```diff
@@ -100,12 +102,49 @@ const SwapAmountScreen: React.FC<SwapAmountScreenProps> = ({
...
+      // to avoid using outdated memoized amount
+      const xlmBalance = balanceItems.find(
+        (item) =>
+          "token" in item &&
+          item.token.type === AssetTypeWithCustomToken.NATIVE,
+      );
+
+      if (xlmBalance) {
+        return xlmBalance.total.isGreaterThanOrEqualTo(new BigNumber(swapFee));
+      }
```
JakeUrban, line 124: I'm not sure I understand, I think we need to check that we have enough XLM for fees in all cases, not just when we're swapping _to_ XLM. And importantly, we need to check that we're leaving enough for fees when we're swapping _from_ XLM and we set the "max".

CassioMG: I think it's looking good now, as long as it works correctly for the set "max" case, which from reading the code it _looks like_ it's already doing that

PR #236, opened 2025-07-31
Prevent immediate row highlight when scrolling lists

CassioMG:
This PR tweaks the scrolling experience by removing the immediate row highlight when scrolling through lists, it does that by adding a minimum `delayPressIn` of `100 ms`. This delay helps to differentiate between a tap and a scroll gesture.

### Before

https://github.com/user-attachments/assets/f13a1b0d-b83e-45cc-9cef-b23988137845

### After

https://github.com/user-attachments/assets/64fd15fc-253a-456f-9a53-4d91ed825128

### DIFF
--- a/src/services/analytics/core.ts
+++ b/src/services/analytics/core.ts
@@ -127,7 +127,6 @@ export const setAnalyticsUserId = (userId: string | null): void => {
  * - effectiveType: Cellular quality (slow-2g, 2g, 3g, 4g) when on cellular
  */
 const buildCommonContext = (): Record<string, unknown> => {
-  const { network } = useAuthenticationStore.getState();
   const { connectionType, effectiveType } = useNetworkStore.getState();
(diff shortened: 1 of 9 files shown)


src/services/analytics/core.ts
```diff
@@ -127,7 +127,6 @@ export const setAnalyticsUserId = (userId: string | null): void => {
  * - effectiveType: Cellular quality (slow-2g, 2g, 3g, 4g) when on cellular
  */
 const buildCommonContext = (): Record<string, unknown> => {
-  const { network } = useAuthenticationStore.getState();
```
danilosilvackl, line 130: thanks for this, I was about to ask if you could remove it <3

danilosilvackl (approved):
lgtm! Just for clarification, we should start adding this delay anywhere we use touchable opacity inside lists now right?

Issue #206, opened 2025-07-13
[Blockaid] Grant Access Flow


CassioMG:
We use Blockaid to scan dApp websites and check if they are malicious.

This task is about integrating with Blockaid and display the scanning warnings when trying to `connect to a dApp` through `WalletConnect`.

Please see the [milestone description](https://github.com/stellar/freighter-mobile/milestone/5) to get more context and clarity regarding Blockaid architecture.

Below is our current [Figma](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2547-35062&m=dev) designs but they are not on the `Phases` page in Figma so would be worth reviewing it and asking Charles to move it to the `Phases` page like we have for the [Add Asset](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3062-197893&m=dev) flow.
<img width="873" height="449" alt="Image" src="https://github.com/user-attachments/assets/2a21f59e-9df7-4959-9e90-4a2114e5db10" />

PR #252, opened 2025-08-05
Blockaid grant dApps access

danilosilvackl:
dApps grant access flow with blockaid scan: 

### Benign asset
site_benign_ios

### Suspicious Asset
site_suspicious_ios
site_suspicious_detail_ios

### Malicious Asset
site_malicious_detail_ios
site_malicious_ios

### Benign asset
site_benign_android

### Suspicious Asset
site_suspicious_android
site_suspicious_detail_android

### Malicious Asset
site_malicious_android
site_malicious_detail_android

_ps: there is some TODO that we have to work on next tasks_
_ps: I wasn't able to find truly malicious or not infected sites/dApps for the testing but we are using the correct response to validate from the blockaid so it should cover most cases_

close: #206

### DIFF
--- a/src/components/screens/WalletKit/DappConnectionBottomSheetContent.tsx
+++ b/src/components/screens/WalletKit/DappConnectionBottomSheetContent.tsx
@@ -1,17 +1,19 @@
+import { List } from "components/List";
+import Spinner from "components/Spinner";
 import { App } from "components/sds/App";
 import Avatar from "components/sds/Avatar";
 import { Badge } from "components/sds/Badge";
 import { Button, IconPosition } from "components/sds/Button";
 import Icon from "components/sds/Icon";
 import { Text } from "components/sds/Typography";
-import { ActiveAccount } from "ducks/auth";
+import { NETWORKS } from "config/constants";
+import { ActiveAccount, useAuthenticationStore } from "ducks/auth";
 import { WalletKitSessionProposal } from "ducks/walletKit";
-import { pxValue } from "helpers/dimensions";
 import useAppTranslation from "hooks/useAppTranslation";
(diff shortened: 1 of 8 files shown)


src/components/screens/WalletKit/DappConnectionBottomSheetContent.tsx
```diff
@@ -60,89 +76,186 @@ const DappConnectionBottomSheetContent: React.FC<
...
         </View>
+      ),
+      titleColor: themeColors.text.secondary,
+    },
+    {
+      icon: <Icon.Globe01 size={16} color={themeColors.foreground.primary} />,
+      title: t("network"),
+      trailingContent: (
+        <Text md primary>
+          {network === NETWORKS.PUBLIC ? t("mainnet") : t("testnet")}
```
CassioMG, line 107: I think the networks name here should be consistent with what we are displaying on Settings/Network regardless of how it's being displayed on Figma right now. Which would mean "Main Net" and "Test Net".

src/components/screens/WalletKit/DappConnectionBottomSheetContent.tsx
```diff
@@ -60,89 +76,186 @@ const DappConnectionBottomSheetContent: React.FC<
...
+                  : t("dappConnectionBottomSheetContent.suspiciousFlag")}
+              </Text>
+            </View>
+
+            <Icon.ChevronRight
+              size={16}
+              themeColor={isMalicious ? "red" : "amber"}
+            />
+          </View>
+        </TouchableOpacity>
```
CassioMG, line 236: use reusable component if created in the other PR

Issue #247, opened 2025-08-04
Settings - Hide small payments


CassioMG:
[Figma designs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3062-217653&t=pxLFu4iAEVZO9ISZ-1):
<img width="415" height="886" alt="Image" src="https://github.com/user-attachments/assets/da9d6af7-1c93-4a87-ad9b-9375fa6b1c21" />

PR #261, opened 2025-08-08
Add 'hide dust payments' preference toggle 

leofelix077:
Closes https://github.com/stellar/freighter-mobile/issues/247 

Tests done with asset_type 'credit_alphanum4' instead of 'native'

https://github.com/user-attachments/assets/ee498fa2-8962-4996-8a20-5aa65f117beb

### DIFF
--- a/src/hooks/useGetHistoryData.ts
+++ b/src/hooks/useGetHistoryData.ts
@@ -2,7 +2,6 @@ import { Horizon } from "@stellar/stellar-sdk";
 import { NetworkDetails } from "config/constants";
 import { BalanceMap, HookStatus } from "config/types";
 import { useBalancesStore } from "ducks/balances";
-import { usePreferencesStore } from "ducks/preferences";
 import {
   getIsDustPayment,
   getIsPayment,
@@ -10,7 +9,7 @@ import {
   filterOperationsByAsset,
   getIsCreateClaimableBalanceSpam,
 } from "helpers/history";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import { getAccountHistory } from "services/backend";
 
 export type HistoryItemOperation = Horizon.ServerApi.OperationRecord & {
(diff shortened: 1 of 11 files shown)


src/hooks/useGetHistoryData.ts
```diff
@@ -113,9 +133,10 @@ function useGetHistoryData(
 
       const balances = getBalances();
 
+      setRawHistory(history);
```
CassioMG, line 136: I see we are calling `setRawHistory` here and then `setHistoryData` on line 142 and then `setHistoryData` will also be called inside useEffect on line 106 with basically the same inputs. Why is that?

src/hooks/useGetHistoryData.ts
```diff
@@ -89,10 +92,27 @@ function useGetHistoryData(
...
+      setHistoryData({
+        balances,
+        history: createHistorySections(
+          publicKey,
+          rawHistory,
+          isHideDustEnabled,
+        ),
+      });
+    }
+  }, [isHideDustEnabled, publicKey, rawHistory, getBalances]);
```
danilosilvackl, line 115: if i understood this correctly, this is responsible to update the history in case the the settings change right?
What if we leave modify the `fetchData` to receive the isHideDustEnabled as a parameter and leave the HistoryScreen responsible to send this information? Im considering this to avoid running code in the background without a need, so the fetch function would only change when the user loads the page again, does that makes sense?

CassioMG:
> Closes #247
> 
> Tests done with asset_type 'credit_alphanum4' instead of 'native'
> 
>  Screen.Recording.2025-08-08.at.11.56.56.mov

it took a while for me to realize it, but I think on your video the payments are being hidden because you are swapping a small value of XLM for SSLX. Is that right? And not because we have small values of SSLX which is actually the desired behavior since we are only interested in XLM here so no changes needed here. I'm highlighting it here just in case someone else has the same question because the History rows do not display the XLM amounts, only the SSLX amounts.

leofelix077:
@CassioMG  yes. I mentioned on the description about the flip. At the moment i don't have residual payments in my account, so I switched the asset type in code for testing and reverted before committing

CassioMG:
> @CassioMG yes. I mentioned on the description about the flip. At the moment i don't have residual payments in my account, so I switched the asset type in code for testing and reverted before committing

Ah ok, now I understand what you mentioned by the description

Issue #262, opened 2025-08-08
Send/Swap amount very small text size in some devices


leofelix077:
Emulator:  iPhone 16 Plus - iOS 26.0 

main branch - 0.8.23

<img width="322" height="652" alt="Image" src="https://github.com/user-attachments/assets/f1dac22c-612b-4598-9641-da7ef809edfd" />

PR #263, opened 2025-08-08
fix pixel value text size calculation

leofelix077:
Closes https://github.com/stellar/freighter-mobile/issues/262

It was breaking in some devices due to the final pixel value ending up being a periodic  e.g. 68.333333333333, depending on the device scale, which breaks the layouting

could either: 
1. round it to nearest full pixel

2. (current implementation) -> limit the decimals, to keep better precision

Screenshot 2025-08-08 at 14 41 11

Screenshot 2025-08-08 at 14 46 11


src/helpers/dimensions.ts
```diff
@@ -52,7 +53,7 @@ const deviceScale = Math.min(deviceAspectRatio / DESIGN_ASPECT_RATIO, 1.2);
 export function pxValue(designValue: number): number {
   const designHeightPercentage =
     deviceScale * (designValue / DESIGN_HEIGHT) * 100;
-  return heightPercentageToDP(designHeightPercentage);
+  return round(heightPercentageToDP(designHeightPercentage), 2);
```
danilosilvackl, line 56: awesome fix, just a small nit - can we use `.toFixed(2)` instead of import `lodash` just for rounding this once or do you think that would break it?

danilosilvackl (approved):
LGTM! awesome fix and save here, left a small suggestion

CassioMG:
@leofelix077 great finding on this PR!

CassioMG (approved):

src/helpers/dimensions.ts
leofelix077: ++. yea makes more sense. adjusted now

Issue #248, opened 2025-08-04
Settings - Validate addresses that require a memo


CassioMG:
### **Feature Purpose**

This feature helps users avoid failed transactions by validating whether destination addresses require a memo before sending payments. Some Stellar accounts (particularly exchanges and custodial services) require a memo to properly credit funds to the correct user account. Without a memo, these transactions will fail or funds may be lost.

We currently do this in 2 different flows:
- `Send payment`
- `Sign dApp transaction`

You should be able to get a sample of an address that requires a memo from the [Stellar Expert API endpoint ](https://api.stellar.expert/explorer/directory?limit=20000000&tag%5B%5D=memo-required).
E.g.: `GAESQGK5TTKPT2JY4STRN6MJU56LNHQVBFROGX5GFIWUPK3JHZ5F5FCI`

To `test` this on the extension for the `Send payment` flow you could just try sending a regular payment to a memo-required address.

To `test` this on the extension for the `Sign dApp transaction` flow you could just try sending a payment to a memo-required address using [Stellar Term's Send feature](https://stellarterm.com/account/send?asset=XLM-native).

Below is how the `extension` implements it as `reference`.

Please note that it checks using both the `Stellar Expert API` _endpoint_ and  the `stellar-sdk` _checkMemoRequired_ function to verify if the destination address requires a memo.
It also caches the list of addresses that require a memo from the `Stellar Expert API` _endpoint_ , caching doesn't seem like a requirement but rather a nice-to-have to improve performance.

### **Extension's Validation Flow**

1. **User initiates transaction** (`Send payment` or `Sign dApp transaction`)
2. **Check if memo validation is enabled** and on mainnet
3. **Fetch memo required accounts** from Stellar Expert API (cached)
4. **Validate destination address** against memo required list
5. **Check transaction memo** - if missing and required, flag the transaction
6. **Display warning** and disable submit button if memo is required
7. **Allow submission** only when memo is provided

### **Extension's Configuration Options**
- **Network Scope**: Only applies to mainnet (not testnet)
- **Default Behavior**: Enabled by default
- **Caching**: Memo required accounts are cached for performance
- **Validation**: Both external API and Stellar SDK validation

#### **1. Settings Management**
- **Storage Key**: `IS_VALIDATING_MEMO_ID` in localStorage
- **Default Value**: `true` (memo validation is enabled by default)
- **Location**: `extension/src/constants/localStorageTypes.ts`

#### **2. Memo Required Accounts Data Source**
[...]

PR #268, opened 2025-08-12
Validate memo required addresses

leofelix077:
Closes https://github.com/stellar/freighter-mobile/issues/248

Added example here of the back and forth navigation, with the new settings screen.

The old screens haven't been deleted yet and still being used on the top settings screen, but could adjust it to also remove from there and link the bottom sheet. 

Or alternatively -> Move back to the fullscreen settings, but already mentioning, it's gets a bit messy mixing up the full screen with the bottom sheets in relation to the navigation, specially due to the walletKitProvider not being exactly a navigatable object and is listening for the deep links, so jumping from full-screen back to a modal, becomes in my opinion somewhat confusing. I had added it initially with the proper navigation, but decided to change the code and to the bottom sheet already to streamline it a bit

https://github.com/user-attachments/assets/ad70c414-d8f4-4988-8f4e-7a71630e6b63


src/hooks/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,136 @@
...
+    [network],
+  );
+
+  const shouldValidateMemo = useMemo(
+    () => isMemoValidationEnabled && isMainnet(network),
+    [isMemoValidationEnabled, network],
+  );
+  const [isMemoMissing, setIsMemoMissing] = useState(shouldValidateMemo);
+
+  const checkMemoRequiredFromCache = useCallback(
```
leofelix077, line 35: @CassioMG please validate the logic here -> not that familiar with the business logic. I tried to take it from the extension, but if there are edge-cases, idk, something like multi transactions (don't know if exists, just mentioning), that are not covered, then I can adjust

src/providers/WalletKitProvider.tsx
```diff
@@ -188,7 +195,7 @@ export const WalletKitProvider: React.FC<WalletKitProviderProps> = ({
       setIsConnecting(false);
       setProposalEvent(null);
       setSiteScanResult(undefined);
-
+      saveMemo("");
```
leofelix077, line 198: restting the memo after submitting the transactions on both send and confirm transaction -> let me know if it's a desired behavior

src/components/TransactionSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,183 @@
...
+  return (
+    <View className="flex-1">
+      <View className="flex-1 justify-between">
+        <View className="flex flex-col gap-2">
+          <View className="flex flex-row items-center gap-2">
+            <Text sm secondary>
+              {t("transactionSettingsBottomSheet.feeTitle")}
+            </Text>
+            <TouchableOpacity onPress={() => {}}>
+              {/* disabled until info is added */}
```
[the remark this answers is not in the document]
CassioMG: I've also dropped a comment on Figma here regarding this missing component: https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu?node-id=6953-333987#1380081536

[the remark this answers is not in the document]
CassioMG: but usually I'd prefer we avoid merging commented/empty code so we keep the codebase clean

src/providers/WalletKitProvider.tsx
```diff
@@ -435,22 +444,46 @@ export const WalletKitProvider: React.FC<WalletKitProviderProps> = ({
...
   return (
     <View className="flex-1">
       {/* Bottom sheet for dApp connection requests */}
       <BottomSheet
         modalRef={dappConnectionBottomSheetModalRef}
         handleCloseModal={handleClearDappConnection}
-        bottomSheetModalProps={{
-          onDismiss: handleClearDappConnection,
-        }}
+        enablePanDownToClose={false} // disabled due to missing custom onClose event on Sheet Handle -> when dragging down to close, event cannot be cleared consistently
```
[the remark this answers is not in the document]
CassioMG: Do we need to apply this on both `dappConnectionBottomSheetModalRef` and `dappRequestBottomSheetModalRef` considering we'd only need to handle memo on `dappRequestBottomSheetModalRef` ?

src/providers/WalletKitProvider.tsx
CassioMG: I think I'd only reset the memo on `handleClearDappRequest` and not on `handleClearDappConnection` since we shouldn't be handling any memos on the `dApp Connection` step but rather only on the `dApp Request` step. Does that make sense?

src/hooks/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,136 @@
...
+    async (txXDR: string, networkType: string): Promise<boolean> => {
```
CassioMG, line 36:
```suggestion
    async (transactionXDR: string, networkType: string): Promise<boolean> => {
```

let's give preference to using `transaction` instead of `tx` whenever possible so we keep the codebase consistency

src/hooks/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      return matchingBlockedTags.some(
+        (tag) => tag === (TRANSACTION_WARNING.memoRequired as string),
+      );
+    },
+    [],
+  );
+
+  const checkMemoRequiredFromStellarSDK = useCallback(
+    async (txXDR: string, networkUrl: string): Promise<boolean> => {
```
CassioMG, line 62:
```suggestion
    async (transactionXDR: string, networkUrl: string): Promise<boolean> => {
```

src/hooks/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,136 @@
...
+      const server = stellarSdkServer(networkUrl);
+      const transaction = TransactionBuilder.fromXDR(txXDR, network);
+
+      try {
+        await server.checkMemoRequired(transaction);
+        return false;
+      } catch (error) {
```
CassioMG, line 69: should we check for a specific kind of error here? I see the extension is checking `if ("accountId" in error)` and it looks like we have this _AccountRequiresMemoError_ from the stellar-sdk

then, if it's not the kind of error we are expecting I think we should probably return `false` and print the error with `logger.error()`

what do you think?

src/hooks/useValidateTransactionMemo.ts
CassioMG: hummm ok I just saw we already have a `transactionXDR` from `useTransactionBuilderStore` so we'd need to use a different name here

src/hooks/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,136 @@
...
+import { useTransactionSettingsStore } from "ducks/transactionSettings";
+import { cachedFetch } from "helpers/cachedFetch";
+import { isMainnet } from "helpers/networks";
+import { getApiStellarExpertIsMemoRequiredListUrl } from "helpers/stellarExpert";
+import { useCallback, useEffect, useMemo, useState } from "react";
+import { stellarSdkServer } from "services/stellar";
+
+export const useValidateTransactionMemo = () => {
+  const { network } = useAuthenticationStore();
+  const { transactionXDR } = useTransactionBuilderStore();
```
CassioMG, line 20: I wonder if we should pass `transactionXDR` as an input for `useValidateTransactionMemo` like `useValidateTransactionMemo(transactionXDR)` instead of retrieving it from `useTransactionBuilderStore` since for the dApp transaction case (WalletKit) it does not use the `useTransactionBuilderStore` at all. If you look at `approveSessionRequest` in `walletKitUtil.ts` you will see that the `xdr` in this case is being retrieved directly from the `sessionRequest` event.

Are you able to currently test this Sign dApp transaction flow using StellarTerm as suggested [on the issue](https://github.com/stellar/freighter-mobile/issues/248)? Let me know in case you need any help with it, you'll need to set some `.env` vars on the project in order to test the Wallet Connect (WalletKit) flow

src/i18n/locales/pt/translations.json
```diff
@@ -83,6 +83,10 @@
     "hideDust": {
       "title": "Ocultar pagamentos pequenos",
       "description": "Ocultar pagamentos menores que 0,1 XLM"
+    },
```
CassioMG, line 86: looks like we are missing some pt translations? E.g.: `addMemoExplanationBottomSheet`

CassioMG: and `transactionSettingsBottomSheet`

src/components/sds/Input/index.tsx
```diff
@@ -116,6 +117,24 @@ export const StyledTextInput = styled.TextInput<
...
+  color: ${({ $isDisabled }: Pick<StyledProps, "$isDisabled">) =>
+    $isDisabled ? THEME.colors.text.secondary : THEME.colors.text.primary};
+  font-family: ${Platform.select({
+    ios: "Inter-Variable",
+    android: "Inter-Regular",
+  })};
+  font-weight: ${Platform.select({
+    ios: "400",
+    android: "normal",
+  })};
```
CassioMG, line 135: it looks like this code is exactly the same as the one we have insite `StyledTextInput`. Could we make it reusable so it's easier to maintain?

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -220,7 +264,26 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+  ]);
+
+  useFocusEffect(
+    useCallback(() => {
+      if (hasOpenedAddMemoExplanationBottomSheet) {
+        handleOpenReview();
+        setHasOpenedAddMemoExplanationBottomSheet(false);
+      }
+    }, [hasOpenedAddMemoExplanationBottomSheet, handleOpenReview]),
+  );
```
CassioMG, line 286: is this correct/necessary? looks like we are not setting `setHasOpenedAddMemoExplanationBottomSheet(true)` anywhere in this PR

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -268,6 +331,7 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
     };
 
     processTransaction();
+    saveMemo("");
```
CassioMG, line 334: should we `await` for `processTransaction()` before reseting the memo?

src/components/RequiredMemoMissingWarning.tsx
```diff
@@ -0,0 +1,27 @@
...
+            {t("transactionAmountScreen.memoMissing")}
+          </Text>
+        </View>
+        <View>
+          <Icon.ChevronRight size={16} color={themeColors.red[11]} />
+        </View>
+      </View>
+    </View>
+  );
+};
```
CassioMG, line 27: This component looks the same we are using on the Blockaid screens, I wonder if we already have or else should be creating a reusable component for it - cc @danilosilvackl

src/components/TransactionSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,183 @@
...
+              isBottomSheetInput
+              fieldSize="lg"
+              value={localFee}
+              leftElement={
+                <Icon.Route size={16} color={themeColors.foreground.primary} />
+              }
+              onChangeText={handleSetLocalFee}
+              keyboardType="numeric"
+              placeholder={MIN_TRANSACTION_FEE}
+              error={error}
```
CassioMG, line 94: it looks like the Fee and Memo fields are sharing the same error message, is this expected?

src/providers/WalletKitProvider.tsx
leofelix077: oh yea. only after the request itself has been handled. makes sense

src/providers/WalletKitProvider.tsx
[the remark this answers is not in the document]
leofelix077: true. only on dApp requests

src/hooks/useValidateTransactionMemo.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        const isMemoRequiredFromSDK = await checkMemoRequiredFromStellarSDK(
+          transactionXDR,
+          networkDetails.networkUrl,
+        );
+
+        setIsMemoMissing(isMemoRequiredFromSDK);
+      } catch (error) {
+        setIsMemoMissing(true);
```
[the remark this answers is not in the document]
leofelix077: business-wise, what is the side-effect?e.g. does it hurt to add a memo where it's not needed? 

as in: if we remove, and then something fails, a transaction might fail without memo, and if it's added but not needed, it goes through?

src/components/TransactionSettingsBottomSheet.tsx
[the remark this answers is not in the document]
leofelix077: thank you! will check with charles next times as well 👌 . will remove them for now

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -78,10 +89,41 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+  const transactionSettingsBottomSheetModalRef = useRef<BottomSheetModal>(null);
+  const [
+    hasOpenedAddMemoExplanationBottomSheet,
+    setHasOpenedAddMemoExplanationBottomSheet,
+  ] = useState(false);
+
+  const onConfirmAddMemo = () => {
+    reviewBottomSheetModalRef.current?.dismiss();
+    addMemoExplanationBottomSheetModalRef.current?.dismiss();
+    transactionSettingsBottomSheetModalRef.current?.present();
```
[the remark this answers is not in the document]
leofelix077: adjusted now to remove from some parts. when moving the flow forward it works -> so the sheets get stacked normally. when moving back it's not able to dismiss them

src/components/RequiredMemoMissingWarning.tsx
leofelix077: taking a look I didn't find anything directly close to it (of course, if you make icon, title, and color dynamic, then can be converted into a generic. can convert it, in case it makes sense to do it now

src/components/BottomSheet.tsx
```diff
@@ -138,7 +141,7 @@ const BottomSheet: React.FC<BottomSheetProps> = ({
       onChange={handleChange}
     >
       <BottomSheetView
-        className="flex-1 bg-background-primary pl-6 pr-6 pt-6 gap-6"
+        className="bg-background-primary pl-6 pr-6 pt-6 gap-6"
```
leofelix077, line 144: https://github.com/gorhom/react-native-bottom-sheet/issues/1573 

flex-1 on bottomsheetview breaks dynamic layouting

src/components/RequiredMemoMissingWarning.tsx
danilosilvackl: Yes, I think we can use the same as the blockaid warnings on the add token and dApp connection [see here](https://github.com/stellar/freighter-mobile/blob/3eab0ab1b03bdd4b2b20bda6fe9e5d36a6b11fd6/src/components/screens/WalletKit/DappConnectionBottomSheetContent.tsx#L213-L221)

leofelix077: adjusted. thank you 💪

src/hooks/useValidateTransactionMemo.ts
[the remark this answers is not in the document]
CassioMG: great questions. I think you are right, it wouldn't hurt to add a memo if not required but it would hurt not adding a memo when it's required 👍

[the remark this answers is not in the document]
CassioMG: so let's leave it as `true`

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
CassioMG: friendly reminder to remove this code

src/hooks/useValidateTransactionMemo.ts
CassioMG: if you are handling the error inside `checkIsMemoRequired` then you should `throw` here instead of returning `true` to forward the error

src/providers/WalletKitProvider.tsx
```diff
@@ -76,6 +80,10 @@ export const WalletKitProvider: React.FC<WalletKitProviderProps> = ({
   const { network, authStatus } = useAuthenticationStore();
   const { account, signTransaction } = useGetActiveAccount();
 
+  const addMemoExplanationBottomSheetModalRef = useRef<BottomSheetModal>(null);
+  const transactionSettingsBottomSheetModalRef = useRef<BottomSheetModal>(null);
+  const { transactionMemo, saveMemo } = useTransactionSettingsStore();
```
CassioMG, line 85: one difference about the Sign dApp Transaction flow when compared to the internal Send and Swap flows is that on the Sign dApp Transaction flow we are not gonna make any changes to the transaction. We are only gonna show the "Memo is required" warning + modal and disable the confirm button to stop the user from proceeding with the transaction. Which means we should remove `useTransactionSettingsStore` from here since we are not editing the transaction in this case. I have also commented it on Figma here: https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu?node-id=6563-76732#1382138812

leofelix077: ohhh yea. that's what I had mentioned during the last sync. I couldn't explain myself properly on what I was meaning. I saw that on the extension it was not letting me proceed, but thought we wanted to have a new behavior for it 

For now/new tickets I will just follow the extension behavior for all implementations then, if it makes sense, and we leave any changes for the future

but for sure! will post here and ping you as soon as it's done

leofelix077:
@CassioMG @danilosilvackl  -> Example video after the cleanup and blocking logic on incoming transactions

⚠️ - At 1:40 in the video you can notice a padding issue on the review modal, but it's fixed on bottom-sheet v5.1.8 (it was recorded before updating it back from 5.1.2 to 5.1.8)


https://github.com/user-attachments/assets/9106e87b-09d0-4ed3-bce5-5bf056fcdc75

src/hooks/useValidateTransactionMemo.ts
```diff
@@ -71,43 +77,49 @@ export const useValidateTransactionMemo = (incomingXdr?: string | null) => {
...
-      setIsMemoMissing(shouldValidateMemo);
+    if (!xdr || !network || !shouldValidateMemo) {
+      return;
     }
-  }, [transactionMemo, shouldValidateMemo]);
+
+    const transaction = TransactionBuilder.fromXDR(xdr, network);
+    const memo =
+      "memo" in transaction && transaction.memo.value
+        ? String(transaction.memo.value).trim()
```
CassioMG, line 91: I'd suggest not trimming the memo string value just to avoid changing it without the user consent

src/providers/WalletKitProvider.tsx
CassioMG: yeah, no worries, I agree this part was a little confusing during the conversations

please feel free to suggest any improvements you see when compared to the extension flow so we can weigh in to check if it's worth adding now or in the future

CassioMG: just checked the video, it is looking great! Thanks for recording all the flows

leofelix077: oh good catch. thanks!

src/hooks/useValidateTransactionMemo.ts
leofelix077: Had added a comment before, but was looking at the wrong place.  Just disconsider, if you got an email relating to this thread

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
CassioMG: Just noting here for tracking, I see you have removed this code on the other PR: https://github.com/stellar/freighter-mobile/pull/273/files#diff-9822eb8a445dde0926940107e4a8283b76617d3fc9f1fde28336f304acb87845L282-L290

PR #272, opened 2025-08-15
[Collectibles] Home UI

### LINKED ISSUE #238: [Collectibles] Home UI
This task is primarily focused on building the Home UI for collectibles.

[Figma designs](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=5923-141570&t=pxLFu4iAEVZO9ISZ-1):

Image

CassioMG:
Closes #238 

This PR adds the Collectibles UI for the Home and Send grids.

It is displaying dummy data at the moment, we will connect it to our backend in a separate PR.

Screenshot 2025-08-14 at 18 42 31

Screenshot 2025-08-14 at 18 42 39

Screenshot 2025-08-14 at 18 45 10

Screenshot 2025-08-14 at 18 45 28

https://github.com/user-attachments/assets/1c9fdb9e-90d3-4c87-9335-8c9c6e2b9c67

https://github.com/user-attachments/assets/4d04a23b-3504-4e29-9143-19f9df3c21ca

### DIFF
--- a/src/components/CollectiblesGrid.tsx
+++ b/src/components/CollectiblesGrid.tsx
@@ -115,7 +115,7 @@ export const CollectiblesGrid: React.FC<CollectiblesGridProps> = React.memo(
           <FlatList
             data={collections}
             renderItem={renderCollection}
-            keyExtractor={(collection) => collection.collectionName}
+            keyExtractor={(collection) => collection.collectionAddress}
             showsVerticalScrollIndicator={false}
             ListFooterComponent={DefaultListFooter}
             refreshControl={
--- a/src/components/CollectiblesGrid.tsx
+++ b/src/components/CollectiblesGrid.tsx
@@ -152,8 +152,9 @@ export const CollectiblesGrid: React.FC<CollectiblesGridProps> = React.memo(
       }
 
       return (
-        <View className="flex-1 pt-4">
-          <Text md secondary>
+        <View className="flex-row items-center justify-center pt-5 gap-2">
+          <Icon.Grid01 size={20} color={themeColors.text.secondary} />
+          <Text md medium secondary>
             {t("collectiblesGrid.empty")}
           </Text>
         </View>
@@ -163,6 +164,7 @@ export const CollectiblesGrid: React.FC<CollectiblesGridProps> = React.memo(
       isLoading,
       error,
       t,
+      themeColors.text.secondary,
       themeColors.secondary,
       renderCollection,
       handleRefresh,
--- a/src/components/CollectiblesGrid.tsx
+++ b/src/components/CollectiblesGrid.tsx
@@ -64,18 +64,26 @@ export const CollectiblesGrid: React.FC<CollectiblesGridProps> = React.memo(
     const renderCollectibleItem = useCallback(
       ({ item }: { item: Collectible }) => (
         <TouchableOpacity
-          className="w-[165px] h-[165px] rounded-2xl overflow-hidden mr-6"
+          className="w-[165px] h-[165px] rounded-2xl overflow-hidden items-center justify-center mr-6 bg-background-tertiary"
           delayPressIn={DEFAULT_PRESS_DELAY}
           onPress={() => onCollectiblePress?.(item.tokenId)}
         >
-          <Image
-            source={{ uri: item.image }}
-            className="w-full h-full"
-            resizeMode="cover"
-          />
+          {/* Placeholder icon for when the image is not loaded */}
+          <View className="absolute z-1">
+            <Icon.Image01 size={45} color={themeColors.text.secondary} />
+          </View>
+
+          {/* NFT image */}
+          <View className="absolute z-10 w-full h-full">
+            <Image
+              source={{ uri: item.image }}
+              className="w-full h-full"
+              resizeMode="cover"
+            />
+          </View>
         </TouchableOpacity>
       ),
-      [onCollectiblePress],
+      [onCollectiblePress, themeColors.text.secondary],
     );
 
     const renderCollection = useCallback(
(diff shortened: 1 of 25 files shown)


src/components/CollectiblesGrid.tsx
```diff
@@ -0,0 +1,175 @@
...
+      [renderCollectibleItem, themeColors.text.secondary],
+    );
+
+    const renderContent = useMemo(() => {
+      if (collections.length > 0) {
+        return (
+          <FlatList
+            data={collections}
+            renderItem={renderCollection}
+            keyExtractor={(collection) => collection.collectionName}
```
aristidesstaffieri, line 118: Maybe this key should be `collectionAddress`?

I don't think `collectionName` is a guaranteed unique value.

JakeUrban:
Can we put the loading spinner in the center of the space where collectibles will render, rather than the top?

src/components/CollectiblesGrid.tsx
CassioMG: great catch, yeah it should be collectionAddress!

src/components/CollectiblesGrid.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    const renderCollectibleItem = useCallback(
+      ({ item }: { item: Collectible }) => (
+        <TouchableOpacity
+          className="w-[165px] h-[165px] rounded-2xl overflow-hidden mr-6"
+          delayPressIn={DEFAULT_PRESS_DELAY}
+          onPress={() => onCollectiblePress?.(item.tokenId)}
+        >
+          <Image
+            source={{ uri: item.image }}
```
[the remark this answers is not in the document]
CassioMG: Agree. @sdfcharles could we have a fallback image for collectibles?

CassioMG:
> Can we put the loading spinner in the center of the space where collectibles will render, rather than the top?

yep, I'll adjust it

CassioMG:
> To confirm, this PR does not implement the fulls send flow, right?

Correct, the Send flow is not in scope for V1.

> 
> Also, @sdfcharles I love the button in the top right corner of the collectibles pane for users to add manually.
> 
> * Can we do something similar for tokens, rather than having the "manage assets" button in the top left of the whole view

I think this is a great idea, it will also address some feedback we've been receiving from users finding it difficult to figure how to add new tokens.

> * What other actions do we anticipate there?

The other action for Collectibles would be to see "Hidden collectibles", which is slated for after V1 launch.
For tokens I think we could probably also have both: Add Token, Hidden tokens once we have hidden tokens implemented.

CassioMG:
> Correct, the Send flow is not in scope for V1.

Huh, I just realized that since Send is not in scope for V1 we shouldn't be displaying the Collectibles grid on the Send flow 🤦  I'll fix it.

src/components/CollectiblesGrid.tsx
CassioMG: https://github.com/stellar/freighter-mobile/pull/272/commits/27b8525f64833446ddced22e769f5485e7f9f6c2

src/components/CollectiblesGrid.tsx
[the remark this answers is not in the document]
sdfcharles: Home Collectibles
placeholder

@CassioMG here's the mockup for the default image, feel free to use the icon we have in SDS or use this png

sdfcharles:
> Also, @sdfcharles I love the button in the top right corner of the collectibles pane for users to add manually.
> 
> * Can we do something similar for tokens, rather than having the "manage assets" button in the top left of the whole view

a simple implementation of this would be to add in the same icon when your on Tokens and then have two links: 1) manage tokens 2) add a token -> link directly to the add tokens flow

CleanShot 2025-08-15 at 17 29 51@2x

assuming this wouldn't be a big lift but wdyt @JakeUrban @CassioMG

CassioMG:
> assuming this wouldn't be a big lift but wdyt @JakeUrban @CassioMG

@sdfcharles not a big lift, this change looks great to me

CassioMG:
> > assuming this wouldn't be a big lift but wdyt @JakeUrban @CassioMG
> 
> @sdfcharles not a big lift, this change looks great to me

@sdfcharles I'm assuming we should remove the "Manage assets" option from the original place and leave it only on the new place, let me know if we should keep in both places

Screenshot 2025-08-15 at 15 54 43

0xchsh:
> > > assuming this wouldn't be a big lift but wdyt @JakeUrban @CassioMG
> 
> > 
> 
> > @sdfcharles not a big lift, this change looks great to me
> 
> 
> 
> @sdfcharles I'm assuming we should remove the "Manage assets" option from the original place and leave it only on the new place, let me know if we should keep in both places
> 
> 
> 
> Screenshot 2025-08-15 at 15 54 43
> 
> 

yeah lets remove from the top left menu then

sdfcharles:
> > > assuming this wouldn't be a big lift but wdyt @JakeUrban @CassioMG
> 
> > 
> 
> > @sdfcharles not a big lift, this change looks great to me
> 
> 
> 
> @sdfcharles I'm assuming we should remove the "Manage assets" option from the original place and leave it only on the new place, let me know if we should keep in both places
> 
> 
> 
> Screenshot 2025-08-15 at 15 54 43
> 
> 

yeah lets remove from the top left menu, makes more sense by Tokens

CassioMG:
> > > > assuming this wouldn't be a big lift but wdyt @JakeUrban @CassioMG
> > 
> > 
> > > 
> > 
> > 
> > > @sdfcharles not a big lift, this change looks great to me
> > 
> > 
> > @sdfcharles I'm assuming we should remove the "Manage assets" option from the original place and leave it only on the new place, let me know if we should keep in both places
> > Screenshot 2025-08-15 at 15 54 43
> 
> @CassioMG yeah lets remove from the top left menu, makes more sense by Tokens

Ok great, I'll implement this menu change in a separate PR

Issue #284, opened 2025-08-20
Swap To scroll view bottom sheet does not respect insets padding


leofelix077:
As shown here, when there's a long list of tokens, the scroll view bottom sheets hides the last elements / isnt scrollable until the end

https://github.com/user-attachments/assets/341ef625-68d9-49ee-971c-6b20d5860343

PR #282, opened 2025-08-19
Sign Transaction details and blockaid

danilosilvackl:
### Added blockaid on sign dApp transaction flow

This PR also adds the transaction detail component (eg. [SignTransactionDetails](https://github.com/stellar/freighter-mobile/blob/c823696c190bf2b61e675358aa9eca77bbf182f3/src/components/screens/SignTransactionDetails/SignTransactionDetails.tsx)) which can be used by any kind of transaction with the data from the hook [useSignTransactionDetails.ts](https://github.com/stellar/freighter-mobile/blob/1f2f8ff4ffea868648dccfb938579973a14d0cef/src/components/screens/SignTransactionDetails/hooks/useSignTransactionDetails.ts)

### Add trustline
https://github.com/user-attachments/assets/9a727bb0-534d-4c90-9742-82dffd1924b6

### Swap
https://github.com/user-attachments/assets/ca20410b-1fc3-451d-8a01-d84683883fea

### Supply
https://github.com/user-attachments/assets/f63bb1ce-124b-49fa-ad92-da38af12ed8d

### Borrow
https://github.com/user-attachments/assets/8374d9d0-e435-4b81-b204-ea65513cbcb6

### Withdraw
https://github.com/user-attachments/assets/1f0c4607-17d6-4d3e-a63e-8c6c4f7aa50d

### Claim
https://github.com/user-attachments/assets/d91d9a4d-9d5e-42ac-84b7-0205c18e0f5c

### Repay
https://github.com/user-attachments/assets/8fe127b7-4d01-41cd-b43b-10e0dedf958b

### fxDao vault *mocked
https://github.com/user-attachments/assets/d3061e0a-bd95-46d0-b385-1a265e711884

### fxDao pay_debt - burn (suspicious/error on blockaid simulation) *mocked
https://github.com/user-attachments/assets/f7dca8d7-db5f-42c8-aecc-db069b3604a3

### Malicious transfer *mocked
https://github.com/user-attachments/assets/7cd3305b-d3b7-4c05-ab3d-e61ed5f43f81

in progress:
- [⏳] record videos on Android

closes: #207

PR #285, opened 2025-08-20
fix adaptive scrolling for bottom sheet views

leofelix077:
Closes https://github.com/stellar/freighter-mobile/issues/284

Using @danilosilvackl base adaptive scroll container from https://github.com/stellar/freighter-mobile/pull/282 -> Fixes the scrollability of views on Swap To, Debug and Manage accounts screens

not basing off his branch not to block the review/merge process, but the BottomSheetAdaptiveContainer and helpers are copied from his branch

Likely broke after the gorhom/bottom-sheet package update when the dynamic sizing was fixed

### DIFF
--- a/src/components/primitives/BottomSheetAdaptiveContainer.tsx
+++ b/src/components/primitives/BottomSheetAdaptiveContainer.tsx
@@ -1,0 +1,61 @@
+import {
+  BOTTOM_SHEET_CONTENT_GAP,
+  BOTTOM_SHEET_CONTENT_TOP_PADDING,
+  BOTTOM_SHEET_MAX_HEIGHT_RATIO,
+} from "config/constants";
+import { calculateScrollableMaxHeight } from "helpers/bottomSheet";
+import React, { useMemo, useState } from "react";
+import { View } from "react-native";
+
+interface BottomSheetAdaptiveContainerProps {
+  header?: React.ReactNode;
+  children: React.ReactNode;
+  sheetMaxHeightRatio?: number;
+  topPaddingPx?: number;
+  bottomPaddingPx?: number;
+  contentGapPx?: number;
+}
+
+const BottomSheetAdaptiveContainer: React.FC<
+  BottomSheetAdaptiveContainerProps
+> = ({
+  header,
+  children,
+  sheetMaxHeightRatio = BOTTOM_SHEET_MAX_HEIGHT_RATIO,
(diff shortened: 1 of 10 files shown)


src/components/primitives/BottomSheetAdaptiveContainer.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface BottomSheetAdaptiveContainerProps {
+  header?: React.ReactNode;
+  children: React.ReactNode;
+  sheetMaxHeightRatio?: number;
+  topPaddingPx?: number;
+  bottomPaddingPx?: number;
+  contentGapPx?: number;
+}
+
+const BottomSheetAdaptiveContainer: React.FC<
```
CassioMG, line 19: could we add a JSDoc for this component so devs have more clarity on its purpose?

src/components/screens/SwapScreen/components/SelectTokenBottomSheet.tsx
```diff
@@ -22,30 +36,106 @@ const SelectTokenBottomSheet: React.FC<SelectTokenBottomSheetProps> = ({
...
+      // For Soroban contracts, pass the contract ID as symbol initially
+      // The TokenDetailsScreen will handle fetching the actual symbol
+      tokenSymbol = tokenId;
+    } else {
+      // Classic asset format: CODE:ISSUER
+      [tokenSymbol] = tokenId.split(":");
+    }
+
+    onTokenSelect(tokenId, tokenSymbol);
+  };
```
CassioMG, line 78: I see we have duplicated code from existing `TokenSelectionContent` component here, but since we'll be soon removing the bottom sheet from the swap flow and use a fullscreen instead I think it makes sense duplicating the code now for simplicity

CassioMG (approved):

src/components/primitives/BottomSheetAdaptiveContainer.tsx
leofelix077: definitely. added it now with an example

src/components/screens/SwapScreen/components/SelectTokenBottomSheet.tsx
leofelix077: yes. I took it from the other code, didn't think about making it common to leave the logic of each sheet separate until the redesign at least

I will keep it like this for now then, and keep in mind to unify (if ever needed) with the redesign. probably will just get excluded, but just in case

PR #301, opened 2025-08-25
[Blockaid] Search Asset Flow

### LINKED ISSUE #210: [Blockaid] Search Asset Flow
We use Blockaid to scan assets and check if they are malicious.

This task is about integrating with Blockaid and display the `small red badges` on top of the Asset icons when trying to `search for a new Asset` and also getting in touch with Charles/Keeks to build a `mobile UI` for it.

Please see the [milestone description](https://github.com/stellar/freighter-mobile/milestone/5) to get more context and clarity regarding Blockaid architecture.

Ideally we should try displaying the "safe" assets at first and push the "malicious" assets to the bottom of the search results list. A good way of testing this would be searching for "USDC", as it would display some "malicious" USDC assets in between some "safe" USDC assets in case we don't re-order them.

For testing this you can use the address below on `Mainnet` which is an issuer of several `malicious` assets. `Please don't finalize adding a token or sending a payment to this address` to avoid putting your account at risk.
- `Mainnet` **MALICIOUS** address: `GBXQAA7X3TBSXA7BUOCAR6U2BY7VQZZJMAOFLCGT6EL3RL22X5VFNSV6`

This is what we see on the extension for this address, note that all their assets are red flagged:
Image
NOTE: we have a separate task for the screens that appear after you tap on "Add asset" ([[Blockaid] Add Asset Flow](https://github.com/stellar/freighter-mobile/issues/159)), on this task we should focus more on displaying the badge on the search results list

danilosilvackl:
Added blockaid scan token to the Add token screen

- Scan tokens when user search for a new token
- Order tokens before showing using the following order
  1. Bening
  2. Suspicious
  3. Malicious

## iOS 
https://github.com/user-attachments/assets/c3092655-8a61-44d2-8221-cce9fcb57554

## Android 🤖
[android_token_search.webm](https://github.com/user-attachments/assets/aba82656-fc59-46a8-926c-70efe8838caf)

closes #210

### DIFF
--- a/src/services/blockaid/api.ts
+++ b/src/services/blockaid/api.ts
@@ -12,6 +12,7 @@ import {
   ScanTokenParams,
   ScanSiteParams,
   ScanTransactionParams,
+  ScanBulkTokensParams,
 } from "services/blockaid/types";
 
 const formatAddress = (tokenCode: string, tokenIssuer?: string): string => {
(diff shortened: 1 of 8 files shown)


danilosilvackl:
@JakeUrban thanks for the insight on that, the first thing is that we were reordering the entire list based on: 
`has_trustline` > `security ordering` 
now I reworked the results to follow the rules you mentioned, its looking like this now: 
simulator_screenshot_DDE42EF7-C929-40ED-A6A8-AAC22C0CAAAF

src/services/blockaid/api.ts
```diff
@@ -56,6 +57,45 @@ export const scanToken = async (
...
 };
 
+export const scanBulkTokens = async (
+  params: ScanBulkTokensParams,
+): Promise<Blockaid.TokenBulkScanResponse> => {
+  const { addressList, network } = params;
+
+  try {
+    if (!isMainnet(network)) {
+      throw new Error(BLOCKAID_ERROR_MESSAGES.NETWORK_NOT_SUPPORTED);
```
CassioMG, line 67: could you confirm it won't block the flow on Testnet?

danilosilvackl: just check to make sure, it doesn't block since we have the try catch on the lookup, if the scan fails we just follow up with whatever we got from the stellar expert

PR #305, opened 2025-08-26
Biometrics integration

### LINKED ISSUE #249: Face id (biometrics)
This task will involve creating the "Enable Face Id" screen on the Onboarding flow and the respective enable/disable switch on Settings screen.

We will use Face ID in 2 different scenarios:
- Unlocking your wallet, so you don't need to type your password every time
- Signing any transaction, so we add an extra layer of security as of today we are not asking for user password when signing transactions to reduce UX friction

We already have the face id `icon` in our codebase, you should be able to use it with the `Icon.FaceId` name.
Image

Below is the "Unlock wallet" screen as reference. It should have the `FaceId` icon next to the "Unlock" button title in case face id is `enabled`. In case it's disabled we should hide the `FaceId` icon and let the user type the password as it works today.
Image

Below are all the screens in which we sign transactions. All of them should have the `FaceId` icon on the button in case it's `enabled`. In case it's `disabled` we should not display the `FaceId` icon and just let the flow works as it does today.
- [Send](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3675-90688&t=pxLFu4iAEVZO9ISZ-1)
- [Swap](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=5277-44168&t=pxLFu4iAEVZO9ISZ-1)
- [Add Asset/Token](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=3430-76566&t=pxLFu4iAEVZO9ISZ-1)
- [Connect to dApp](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2547-35062&t=pxLFu4iAEVZO9ISZ-1) (lower priority as you are not actually signing any transaction here, but a nice-to-have)
- [Sign dApp transaction](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=2547-34944&t=pxLFu4iAEVZO9ISZ-1)

Please have in mind that you should test this Face id feature on both Android and iOS devices which could potentially have different APIs and different behaviors.

leofelix077:
Closes #249 

https://github.com/user-attachments/assets/7c4aac1e-0f82-4316-911b-1fb6cef5d526


JakeUrban:
Can we get videos for using faceID on login and signing transactions?

src/components/ConfirmationModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+  return (
+    <Modal visible={visible} onClose={onClose}>
+      <Text xl regular>
+        {title}
+      </Text>
+      <View className="h-2" />
+      <Text md regular secondary>
+        {message}
+      </Text>
+      <View className="h-8" />
```
danilosilvackl, line 136: can we maybe use `margin` or `gap` here instead of the empty views with height? I think it's easier to maintain the spacing consistent that way

ios/freighter-mobile/Info.plist
```diff
@@ -44,6 +44,8 @@
 	</dict>
 	<key>NSCameraUsageDescription</key>
 	<string>Freighter needs access to your Camera to scan QR codes.</string>
+	<key>NSFaceIDUsageDescription</key>
+	<string>Freighter needs access to your Face ID to unlock the app.</string>
```
CassioMG, line 48: In addition to this, we need to add the `en.lproj` and `pt.lproj` strings on XCode for `NSFaceIDUsageDescription` so the pop-up gets translated like we have for `NSUserTrackingUsageDescription`. Not related to this PR, but would you mind also adding the translations for `NSCameraUsageDescription` which are also missing? Thanks!

ios/en.lproj/InfoPlist.strings
```diff
@@ -1 +1,3 @@
 "NSUserTrackingUsageDescription" = "Allow Freighter to collect limited information about usage. We will collect public keys, transaction amounts, and balances. This information will not be used for marketing or to identify you personally. We will not collect information such as name, address, email addresses, phone numbers, etc.";
+"NSFaceIDUsageDescription" = "Freighter needs access to your Face ID securely log in and approve transactions.";
```
CassioMG, line 2:
```suggestion
"NSFaceIDUsageDescription" = "Freighter needs access to your Face ID to securely log in and approve transactions.";
```

package.json
```diff
@@ -79,12 +81,12 @@
     "nativewind": "^4.1.23",
     "node-libs-react-native": "^1.2.1",
     "patch-package": "^8.0.0",
-    "postinstall-postinstall": "^2.1.0",
```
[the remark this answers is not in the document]
CassioMG: happened after merging with main

src/config/constants.ts
```diff
@@ -245,8 +246,10 @@ export enum STORAGE_KEYS {
   COLLECTIBLES_LIST = "collectiblesList",
   ACTIVE_NETWORK = "activeNetwork",
   RECENT_ADDRESSES = "recentAddresses",
+  FACE_ID_ENABLED = "faceIdEnabled",
   MEMO_REQUIRED_ACCOUNTS = "memoRequiredAccounts",
   WELCOME_BANNER_SHOWN_PREFIX = "welcomeBanner_shown_",
+  HAS_SEEN_FACE_ID_ONBOARDING = "hasSeenFaceIdOnboarding",
```
CassioMG, line 252: @leofelix077 is this specific to iOS FaceID or should we rename it to say Biometrics instead?

leofelix077: oh yeah. it's biometrics in general. will rename

CassioMG (approved):
Lgtm - left a comment regarding a missing package

package.json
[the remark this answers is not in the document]
leofelix077: replied on Slack -> the normal postinstall and postinstall-postinstall were running in loop, not enabling the yarn install to run properly and finish. it was removed on purpose

Issue #160, opened 2025-06-08
Sentry setup + initial calls


CassioMG:
Let's try installing Sentry before releasing it publicly so we could track and monitor exceptions to help with debugging and fixing bugs.

This task is about installing and configuring the Sentry project and also adding the error tracking calls in all (or at least some of the) desired places and test it.

By "desired places" we can probably consider all (or most) of our current `logger.error()` calls.

We can also take a look on where the extension is tracking error events and try to port that to mobile whenever it makes sense.

After adding the tracking calls we should test all features of the app and check if there is some error event being triggered too often and analyze if it makes sense to keep tracking this specific error or not to avoid polluting Sentry.

Extension Sentry project as reference: https://stellarorg.sentry.io/projects/freighter/?project=5599259

PR #318, opened 2025-08-29
Sentry setup and initial calls

danilosilvackl:
Initial setup, configuration and calls of sentry

Sentry have a very nice wizard that configures and wire the entire project with it
so just out of the box we already have any unexpected crash and all `throw errors` and promise rejections are already tracked

- Added sentry to the logger, so now all the occurrences of logger already have sentry integration as well

sentry dashboard

https://github.com/user-attachments/assets/13b641be-b8ee-442c-97c4-13827a4335f4

extra: responsiveness adjustments for send transaction screen:
| send_before | send_after | 
| :---: | :---:  | 
| Before | After |

closes: #160


src/config/logger.ts
```diff
@@ -118,19 +119,134 @@ class Logger {
...
+};
+
+/**
+ * Initialize Sentry integration for the logger
+ * This should be called after Sentry.init() in the application bootstrap
+ */
+export const initializeSentryLogger = (): void => {
+  // Use combined adapter for development (console + Sentry for Spotlight),
+  // Sentry only for production
+  logger.setAdapter(__DEV__ ? combinedAdapter : sentryAdapter);
```
CassioMG, line 251: awesome work with the adapters

we should do a test/scan of all existing logger.error() calls to check if we should remove some to avoid polluting Sentry, or if we'd like to add some that seem worth it

src/config/logger.ts
```diff
@@ -118,19 +119,134 @@ class Logger {
...
+  warn: (context: string, message: string, ...args: unknown[]) => {
+    Sentry.addBreadcrumb({
+      level: "warning",
+      message: `[${context}] ${message}`,
+      data: args.length > 0 ? { args: sanitizeLogData(args) } : undefined,
+      category: context,
+    });
+
+    // Also capture as a warning-level message for visibility
+    Sentry.captureMessage(`[${context}] ${message}`, "warning");
```
CassioMG, line 197: is this worth (or heavy) to add breadcrumbs in all cases in addition to the error case?

src/config/sentryConfig.ts
```diff
@@ -0,0 +1,174 @@
...
+  TRACK_ERRORS_WHEN_ANALYTICS_DISABLED: true,
+
+  // Reduced context when user has disabled analytics
+  MINIMAL_CONTEXT_FIELDS: [
+    "platform",
+    "platformVersion",
+    "network",
+    "appVersion",
+    "buildVersion",
+  ] as const,
```
CassioMG, line 31: should we review this with legal?

src/config/sentryConfig.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  // Base context that's always included
+  const baseContext: Record<string, unknown> = {
+    platform: Platform.OS,
+    platformVersion: Platform.Version,
+    network: network.toUpperCase(), // Stellar network (TESTNET, PUBLIC, FUTURENET)
+    appVersion: getVersion(),
+    buildVersion: getBuildNumber(),
+    bundleId: getBundleId(),
+  };
```
CassioMG, line 55: should this automatically be based of MINIMAL_CONTEXT_FIELDS?

I see bundleId is not present on MINIMAL_CONTEXT_FIELDS, should we add it there?

src/config/sentryConfig.ts
```diff
@@ -0,0 +1,174 @@
...
+    const { isEnabled: analyticsEnabled } = useAnalyticsStore.getState();
+
+    if (analyticsEnabled) {
+      // Use the same user ID as analytics for consistency
+      const userId = await getUserId();
+
+      Sentry.setUser({ id: userId });
+    } else {
+      // Anonymous user when analytics is disabled
+      Sentry.setUser({ id: "anonymous" });
```
CassioMG, line 95: will Sentry merge all anonymous events `from everyone` under this single `anonymous` user?
should we use something like `anonymous-randomRash()` to also make each anonymous user anonymous but unique at the same time?

src/config/sentryConfig.ts
CassioMG: humm I guess this is what the code on line 127 is doing? is the `appContext` below based of the `baseContext` above?

```
        SENTRY_CONFIG.MINIMAL_CONTEXT_FIELDS.forEach((field) => {
          if (event.contexts?.appContext?.[field]) {
            minimalContext[field] = event.contexts.appContext[field];
          }
        });
```

src/config/sentryConfig.ts
```diff
@@ -0,0 +1,174 @@
...
+};
+
+/**
+ * Sets up Sentry context based on current app state and user preferences
+ */
+export const updateSentryContext = async (): Promise<void> => {
+  try {
+    const context = buildSentryContext();
+
+    Sentry.setContext("appContext", context);
```
CassioMG, line 84: does it matter if we are calling `Sentry.setContext` before or after `Sentry.setUser` ?

.gitignore
```diff
@@ -77,5 +77,9 @@ yarn-error.log
 # Environment variables
 .env
 
+# Sentry configuration
+android/sentry.properties
+ios/sentry.properties
```
CassioMG, line 82: can you send those and the SENTRY_DSN on keybase please? Thanks!

CassioMG:
what do you see when you tap "All Envs" here?
Screenshot 2025-08-29 at 17 19 46

CassioMG:
@danilosilvackl great work with re-using the logger structure and redacting sensitive info 🚀 

Left some suggestions - thanks!

src/config/logger.ts
CassioMG: eg.: could you please remove the log for `WalletConnect URI param not found:` and just leave the early return there? I don't think it's worth reporting that "error" as I see some dApps are systematically issuing intermediary deep-links which we don't use which results in that "error" being logged all the time (it is not really an error but rather an unexpected extra deep-link)

src/config/logger.ts
danilosilvackl: the idea behind that is to have a more complete trace on the user session, then it will appear here
image

but we could totally tone it down and leave only the `captureException` on the error and optionally the `captureMessage` on the warn

src/config/sentryConfig.ts
danilosilvackl: only if you think thats necessary, I used the same fields from amplitude here for reference

.gitignore
danilosilvackl: yes! i'm still talking with Bruno about these, i'm using a mock account for now, will send when I have the correct data

src/config/logger.ts
danilosilvackl: thats a nice idea, I think its safe to remove most `logger.error()` occurrences when we are already throwing the error because that will be already be sent to sentry

but I will scan all the places and check where we can make those changes

src/config/sentryConfig.ts
danilosilvackl: exactly, in the base full context we add the public key and some other details, but in lines :124:136 we use the `MINIMAL_CONTEXT_FIELDS` to scrub the data again just in case the context was set before and now the user disabled the data sharing

src/config/sentryConfig.ts
danilosilvackl: yeah, after your comment I took some time to simplify this, since our userId is already a random generated one, that is not used to identify the user I'm now just using this as a default.
so now its the same behavior as the amplitude:

> initialize sentry
> generate user id
> identify user
> use the same id for all subsequent sentry calls

src/config/logger.ts
danilosilvackl: awesome, I cleaned the logger of this then, if we need breadcrumbs for some reason we add them manually

src/config/logger.ts
CassioMG: > thats a nice idea, I think its safe to remove most `logger.error()` occurrences when we are already throwing the error because that will be already be sent to sentry

this is a nice feature, but the downside of it is that we probably won't be able to add `extra` info to those errors being catch by this generic catcher, is that right? So depending on the error we may still want to use our custom catcher to add useful debug `extra` info to it

danilosilvackl: exactly, I cleaned most places for now, but we may want to add them back in some places

src/config/logger.ts
```diff
@@ -118,19 +119,134 @@ class Logger {
...
+    "ipAddress",
+    "location",
+    "coordinates",
+  ];
+
+  piiFields.forEach((field) => {
+    if (field in sanitized) {
+      sanitized[field] = "[REDACTED]";
+    }
+  });
```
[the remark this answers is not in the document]
danilosilvackl: it does! I added those fields and made it work with case insensitive

src/config/sentryConfig.ts
danilosilvackl: on amplitude we treat the flag a bit different, if it is turned off we don't send the events at all.
for sentry I built different since it tracks error we may want some of that data to be able to reproduce errors better even with the flag turned off

CassioMG:
> @CassioMG might need to adjust the CI pipeline to make sure the iOS/Android debug symbols are uploaded during the build process. Not sure if this is still the case, but if so, it will have the stacktrace as like `js.min line 1:300000 (anonymous)` instead of the corresponding source map on prod builds

@leofelix077 are you talking about the app distribution pipeline? We don't have it configured yet, we are manually releasing the builds 😅  we planned on integrating GH Actions/fastlane for app distribution after launch

danilosilvackl:
@CassioMG and @leofelix077 I generated a release build and tested on my android device and got this video here, its from a error that i manually forced it, but we can see that it shows the `env` as `production` and way less breadcrumbs now that we cleaned the other levels on the logger.

https://github.com/user-attachments/assets/3895f455-2a90-472d-a1d8-a11baca7c80a

PR #319, opened 2025-09-01
Coinbase onRamp integration

leofelix077:
@CassioMG  @danilosilvackl -> OAuth isn't supported (nor allowed as far as I have seen on the Apple and Google guidelines) inside webviews, I initially tried to make it work with sending the user to the Discovery Screen and open a new tab to handle the flow inside the app, but it's not possible

As coinbase has the authentication flows that allows for social sign in with passkey/apple/goole, if the user tries to use OAuth, it gets stuck trying to login. For this reason the flow follows more simplified, just opening the user's system browser


src/hooks/useCoinbaseOnramp.tsx
```diff
@@ -0,0 +1,90 @@
...
+    try {
+      const data = await fetchData();
+      const url = getCoinbaseUrl({ sessionToken: data.token, token });
+
+      const supported = await Linking.canOpenURL(url);
+      if (!supported) {
+        throw new Error("Cannot open Coinbase URL");
+      }
+
+      await Linking.openURL(url);
```
leofelix077, line 75: https://github.com/stellar/freighter-mobile/pull/319#issue-3373643902

src/hooks/useCoinbaseOnramp.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }, []);
+
+  const openCoinbaseUrl = useCallback(async () => {
+    if (isLoading || !publicKey) {
+      return;
+    }
+
+    // Check if Coinbase onramp is available on this platform
+    if (!isCoinbaseOnrampAvailable()) {
+      throw new Error("Coinbase onramp is not availabl");
```
CassioMG, line 72:
```suggestion
      throw new Error("Coinbase onramp is not available");
```

PR #321, opened 2025-09-02
QR Code "Send To" Flow

### LINKED ISSUE #288: Send - QR Code Scan
We should be able to scan a QR code of another Freighter wallet on the Send flow to use as destination address.

We have an existing `ScanQRCodeScreen` screen which is being used to [scan dApp QR codes](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=5496-211327&t=YxRy3oXUiTXiPleO-1), notice that the scan functionality is centralized on the `QRScanner` component which should easily be reused. We should either try reusing the existing `ScanQRCodeScreen` screen on both flows or else rename it to `ScanDappQRCodeScreen` and create a separate `ScanSendQRCodeScreen` to handle Send scanning.

[Scan on Send flow](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=6739-268719&t=YxRy3oXUiTXiPleO-1):
Image

This task also includes creating the "Receive XLM" modal on the `AccountQRCodeScreen` screen which will teach users they can scan QR codes from each other. Please notice that the navigation right header button will show either the info `(?)` button or the `scan` button depending on where users are navigating from as explained below.

This should appear when `showNavigationAsCloseButton=false` :
Image

This should appear when `showNavigationAsCloseButton=true` :
Image

[Receive XLM modal](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=6597-195688&t=YxRy3oXUiTXiPleO-1):
Image

leofelix077:
Closes #288 

2 questions here for @sdfcharles  and @CassioMG :

In the sense of validation, continue flow if the user scans an 'invalid' qr code? e.g. Only recognize walletconnect scan on its own flow and Wallet scan on its own flow, or let it recognize any QR and leave it for the responsibility of the user?

And 2. On the send flow: Just paste it as it's currently implemented, or already redirect the user back to the Amount screen as a 'confirmed' action?

(Also ignore the 'Scan QR Code' title. I just saw on Figma afterwards. it has been changed to 'Scan to send'

https://github.com/user-attachments/assets/ec8b6440-6f9e-493d-ab36-e1318112ec2e


src/helpers/qrValidation.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Validates if a string is a valid WalletConnect URI
+ * WalletConnect URIs typically start with "wc:" followed by a topic and version
+ * Format: wc:{topic}@{version}?bridge={bridge}&key={key}
+ *
+ * @param uri The URI string to validate
+ * @returns True if the URI appears to be a valid WalletConnect URI
+ */
+export const isValidWalletConnectURI = (uri: string): boolean => {
```
leofelix077, line 12: please check the logic here in regards to parameters. added with Cursor help to define the different values

CassioMG: it looks like Cursor added more complexity to it, I believe it may still work like that but I'd prefer keeping our simplified check for now so that we don't need to test the URLs from all dApps all over again

src/i18n/locales/en/translations.json
```diff
@@ -129,6 +129,9 @@
     "pairingError": "Failed to pair",
     "invalidUriError": "Please enter a valid QR Code URI"
   },
+  "qrCodeScannerScreen": {
+    "title": "Scan to send"
+  },
```
CassioMG, line 134: are we missing this for pt?

src/hooks/useSendFlowQrCodeScanner.tsx
```diff
@@ -0,0 +1,194 @@
...
+      !isSearching &&
+      searchResults.length > 0
+    ) {
+      // Save the recipient address to transaction settings store
+      // The send recipient store already has the correct address from the search
+      saveRecipientAddress(destinationAddress);
+
+      // Navigate directly to transaction amount screen with the selected token
+      // Reset the navigation stack so back action goes to main tab instead of QR scanner
+      navigation.reset({
```
CassioMG, line 142: this seems like a bit of a brute force, could we use `navigation.pop(n)`, `navigation.popTo(route)` or `navigation.replace(route)` instead to handle this ?

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -60,39 +76,76 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
+      scannedData &&
+      storedSource === QRCodeSource.ADDRESS_INPUT &&
+      !isConsumed
+    ) {
+      // Use the scanned data as the search input
+      handleSearch(scannedData);
+      // Mark the data as consumed
+      consumeQRData();
+    }
+  }, [scannedData, storedSource, isConsumed, consumeQRData, handleSearch]);
```
CassioMG, line 140: do we still need to handle this here in case we navigate users straight to the Amount screen as mentioned [here](https://github.com/stellar/freighter-mobile/pull/321#issuecomment-3246659361)?

src/hooks/useWalletConnectQrCodeScanner.tsx
```diff
@@ -0,0 +1,270 @@
...
+  scannerTitle: string;
+  /** Context for analytics */
+  context: QRCodeSource.WALLET_CONNECT;
+}
+
+interface QRCodeScreenConfig {
+  /** Whether to show header right button */
+  showHeaderRight: true;
+  /** Whether to use popToTop for closing */
+  usePopToTop: true;
```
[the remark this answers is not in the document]
leofelix077: not needed anymore

src/components/screens/AccountQRCodeScreen/AccountQRCodeScreen.tsx
```diff
@@ -49,23 +54,47 @@ const AccountQRCodeScreen: React.FC<AccountQRCodeScreenProps> = ({
...
-      } else {
-        navigation.navigate(ROOT_NAVIGATOR_ROUTES.SCAN_QR_CODE_SCREEN);
+    // If the scan route is already in the stack, go back to it
+    // Otherwise, navigate to it
+    if (scanRouteIndex !== -1) {
+      // Calculate how many screens to go back
+      const currentIndex = routes.length - 1;
+      const stepsBack = currentIndex - scanRouteIndex;
+      for (let i = 0; i < stepsBack; i++) {
+        navigation.goBack();
```
CassioMG, line 74: why not use `navigation.popTo(ROOT_NAVIGATOR_ROUTES.SCAN_QR_CODE_SCREEN);` in place of this loop?

src/helpers/qrValidation.ts
CassioMG: On Allbridge it failed because you were on the `Arbitrum` network, if you switch to the `Stellar` network it should work

PR #325, opened 2025-09-03
[Discovery] Add blur to tabs thumbnails

danilosilvackl:
Now the tab overview thumbnails show a blurred image

- added the [react-native-fast-opencv](https://github.com/lukaszkurantdev/react-native-fast-opencv) package
- now when we take screenshots it works like this: 
  - take screenshot from WebView
  - use the `OpenCV` to blur the image
  - save image on storage

### iOS 
https://github.com/user-attachments/assets/9b55d11e-7240-4afa-9e5b-415ac2c2e1c9

### Android 🤖
https://github.com/user-attachments/assets/b28214f6-0892-4816-8b4c-9043266fc926


src/helpers/screenshots.ts
```diff
@@ -209,20 +265,26 @@ export const captureTabScreenshot = async ({
...
 
         await saveScreenshot(screenshotData);
-        updateTab(tabId, { screenshot: uri });
+        updateTab(tabId, { screenshot: blurredUri });
 
         logger.debug(source, `Screenshot captured for tab ${tabId}`);
+
+        // Clear OpenCV buffers
+        // otherwise it will cause a memory leak
+        OpenCV.clearBuffers();
```
CassioMG, line 287: should we call it on `finally` so we rest assure it'll always be called?

danilosilvackl: oh, yeah, updated!

danilosilvackl:
@CassioMG here is the new blur for iOS 
Screenshot 2025-09-03 at 20 59 20

PR #333, opened 2025-09-05
[FIX] Improve balance Items percentage comparison

danilosilvackl:
Added new condition to the percentage color comparison: 

- before: greater than 0 
- now: greater than 0.01

https://github.com/user-attachments/assets/2ee1d771-748f-4184-aa59-f598dea94969


aristidesstaffieri:
Should we add a unit test to cover this?

CassioMG (approved):

PR #336, opened 2025-09-08
[FIX] Improve balance Items percentage comparison

danilosilvackl:
Added new condition to the percentage color comparison:

- before: greater than 0.01
- now: greater than 0.0099999

https://github.com/user-attachments/assets/a25c9b90-6532-4821-845e-8525c576e407

### DIFF
--- a/__tests__/components/screens/ValidateRecoveryPhraseScreen.test.tsx
+++ b/__tests__/components/screens/ValidateRecoveryPhraseScreen.test.tsx
@@ -2,6 +2,7 @@ import type { RouteProp } from "@react-navigation/native";
 import type { NativeStackNavigationProp } from "@react-navigation/native-stack";
 import { userEvent, screen, waitFor } from "@testing-library/react-native";
 import { ValidateRecoveryPhraseScreen } from "components/screens/ValidateRecoveryPhraseScreen";
+import { BiometricsSource } from "config/constants";
 import { AUTH_STACK_ROUTES } from "config/routes";
 import type { AuthStackParamList } from "config/routes";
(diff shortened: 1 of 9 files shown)


CassioMG:
> > @danilosilvackl how does it appear when it's exactly zero? Should it be "0.00%" or "+0.00%" ?
> 
> when its exactly zero it appears "0.00%"
> 
>  Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-09-05.at.16.29.40.mp4 
> the [formatting method](https://github.com/stellar/freighter-mobile/blob/0bf9110ce030a889f1f3a1b26e21ea44426e06a3/src/helpers/formatAmount.ts#L121-L123) only applies the `+` for values greater than `0`

can we change the formatter to also apply the "+" when it's exactly zero? So we keep consistency on the UI

CassioMG:
> > > @danilosilvackl how does it appear when it's exactly zero? Should it be "0.00%" or "+0.00%" ?
> > 
> > 
> > when its exactly zero it appears "0.00%"
> > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2025-09-05.at.16.29.40.mp4
> > the [formatting method](https://github.com/stellar/freighter-mobile/blob/0bf9110ce030a889f1f3a1b26e21ea44426e06a3/src/helpers/formatAmount.ts#L121-L123) only applies the `+` for values greater than `0`
> 
> can we change the formatter to also apply the "+" when it's exactly zero? So we keep consistency on the UI

or leave the formatter as "0.00%" but make sure it always appears as "0.00%" for numbers like 0.001

PR #343, opened 2025-09-09
upgrade to yarn v4.9.4 and add resolutions

### LINKED ISSUE #345: Upgrade to Yarn 4 and add resolutions for malware
* Upgrade to Yarn 4+
* Use Yarn 4 resolutions to make sure we do not install compromised npm packages

piyalbasu:
Closes #345

This upgrades to a more modern version Yarn which allows us to use `resolutions`, a mechanism by which we can replace specific versions of a package with a different version of our choosing. Please note that devs need to migrate to yarn v4.9.4 using the following guide:

https://yarnpkg.com/migration/guide#migration-steps


package.json
```diff
@@ -200,5 +200,22 @@
     "stream": "stream-browserify",
     "vm": "vm-browserify",
     "fs": "react-native-level-fs"
-  }
+  },
+  "resolutions": {
+    "chalk@5.6.1": "5.6.2",
```
piyalbasu, line 205: this is the pattern for replacing affected versions with a new version.

For each of the affected versions, I'm replacing with the latest unaffected version on npm

PR #346, opened 2025-09-09
Add necessary android permissions

danilosilvackl:
Add android permissions that were missing

This should fix the Save to Photos on android 13 or less


CassioMG (approved):
LGTM

PR #347, opened 2025-09-09
Delete account functionality

### LINKED ISSUE #331: "Delete account" functionality
https://stellarfoundation.slack.com/archives/C03347FNAHK/p1756936811879849

CassioMG:
Closes #331 

The "Delete account" button will take the same action as "Log out => Forgot password" so it wipes out the wallet data.

This PR also makes sure we are using the ENV var for the backend url instead of the hardcoded `stg` that was being used for testing.

### Android
https://github.com/user-attachments/assets/7cb28e3c-886d-48b0-a97a-7e24d1542dae

### iOS
https://github.com/user-attachments/assets/8889d964-1d99-4a00-b070-58901a349bd5


sdfcharles:
> @CassioMG can we have one more step after they click `Delete Account`? i can imagine a user would panic if they accidentally tapped that and then it immediately takes them to the login screen
> 
> designs from the figma CleanShot 2025-09-09 at 13 05 51@2x

update: need the button only as the legal team said it would be enough

danilosilvackl (approved):

PR #351, opened 2025-09-09
Allow automatic Collectibles fetching

CassioMG:
This PR makes it so the app will fetch Collectibles early on Home screen using only the wallet public key so that users don't necessarily need to manually add a Collectible beforehand to trigger the fetching.

https://github.com/user-attachments/assets/24e79c65-b8cc-4e7d-8581-8a05dc20dd74


src/navigators/TabNavigator.tsx
```diff
@@ -75,6 +76,12 @@ export const TabNavigator = () => {
     network: networkDetails.network,
   });
 
+  // Fetch collectibles when component mounts or when publicKey/network changes
+  useFetchCollectibles({
```
CassioMG, line 80: This will fetch collectibles as soon as we have a publicKey available

aristidesstaffieri (approved):

PR #369, opened 2025-09-13
WIP: Improved WebView management

danilosilvackl:
TODO:

- [ ] Record evidence videos


src/ducks/browserTabs.ts
```diff
@@ -220,6 +229,65 @@ export const useBrowserTabsStore = create<BrowserTabsState>()(
...
+      },
+
+      unregisterWebView: (tabId: string) => {
+        set((state) => ({
+          ...state,
+          activeWebViewIds: state.activeWebViewIds.filter((id) => id !== tabId),
+        }));
+      },
+
+      getWebViewDisposalCandidates: () => {
```
CassioMG, line 282: is this function not being used? should we remove it?

danilosilvackl: yeah, it was unused because I simplified the cleanup of the `webviews`
but after your comment I went back to the more complete cleanup because that way I have more control over the cache, cookies and ref cleanup

danilosilvackl:
> @danilosilvackl thanks for working on this.
> 
> So on iOS it appears to be working just fine which is awesome.
> 
> But on Android the webview scrolling (and touching in general) stops working after a tab switch. Could you take a look? Thanks

After we discussed offline I made an adjustment and now its working just fine, it looks like this: 
[pointer_events.webm](https://github.com/user-attachments/assets/42270595-bc7d-4623-826c-7323d964b90d)

PR #373, opened 2025-09-16
Fix Icon buttons not responding to touch

CassioMG:
Reverting [this change](https://github.com/stellar/freighter-mobile/pull/370#discussion_r2350129319) from other recent PR since it was added only to fix tests but resulted in some buttons not being responsive (e.g. top navigation buttons and tab bar buttons)

### DIFF
--- a/src/components/sds/Icon/index.tsx
+++ b/src/components/sds/Icon/index.tsx
@@ -1181,7 +1181,7 @@ import { THEME } from "config/theme";
 import { px, pxValue } from "helpers/dimensions";
 import useColors from "hooks/useColors";
 import React from "react";
-import { Pressable, View } from "react-native";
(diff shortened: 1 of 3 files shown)


src/components/sds/Icon/index.tsx
```diff
@@ -2453,13 +2453,13 @@ const IconComponent: React.FC<IconProps> = ({
   }
 
   const iconElement = (
-    <Pressable testID={testID} onPress={onPress}>
```
danilosilvackl, line 2456: just in case, switching `Pressabe` to `TouchableOpacity` do not work as well?

danilosilvackl (approved):
just need to check if need to remove the tests or not, other than that lgtm

PR #387, opened 2025-09-19
[BUG] Bad scan parameters

aristidesstaffieri:
Filters out tokens with no issuer in order to be able to scan other tokens from search results.

### DIFF
--- a/src/hooks/useTokenLookup.ts
+++ b/src/hooks/useTokenLookup.ts
@@ -247,11 +247,7 @@ export const useTokenLookup = ({
       try {
         const addressList = formattedRecords
           .filter((record) => record.issuer)
-          .map((token) =>
-            token.issuer
-              ? `${token.tokenCode}-${token.issuer}`
-              : token.tokenCode,
-          );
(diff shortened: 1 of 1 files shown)


src/hooks/useTokenLookup.ts
```diff
@@ -245,9 +245,13 @@ export const useTokenLookup = ({
 
     if (formattedRecords.length > 0 && isMainnet(network)) {
       try {
-        const addressList = formattedRecords.map((token) =>
-          token.issuer ? `${token.tokenCode}-${token.issuer}` : token.tokenCode,
-        );
+        const addressList = formattedRecords
+          .filter((record) => record.issuer)
+          .map((token) =>
+            token.issuer
```
CassioMG, line 251: do we still need this line if we are filtering right above it?

aristidesstaffieri: ah no we dont, removed in 36ea1ba6c5c4ed61d4af78eb4d5a5a12451c1f0b

CassioMG (approved):
LGTM

Issue #281, opened 2025-08-19
Delete `City` and `Carrier` user properties from Amplitude


Keeeeeeeks:
For mobile users, can we delete the following user props from User Profiles, and purge all historical tracking of these props:
- `Carrier`
- `City`

Reasoning:
- At this time, I don't believe we need to collect these data points

danilosilvackl:
@Keeeeeeeks @CassioMG after a extensive research on the amplitude documentations and after some tests I figured some stuff: 
- We can't purge historical data
- We can keep the SDK from tracking the `carrier`, but we can't hide the field from appearing on the platform UI, but we can filter it out on dashboards
- The only way to keep Amplitude from tracking the City is to disable the ip address tracking, but that would imply in not having region or country tracked as well. see [Location](https://amplitude.com/docs/sdks/analytics/react-native/react-native-sdk#location) for more information
- we could send the region or country fields ourselves, but then we would have to leverage another package or come up with our own implementation to track those fields then send to amplitude

Keeeeeeeks:
- Freighter BE does collect country (but not city/state) when a user makes a request
- Sentry deduces country from IP address
- Looks like Cloudflare collects the country data for Freighter BE: `cf-ipcountry`

PR #389, opened 2025-09-19
[Android] App must support 16 KB memory page sizes

### LINKED ISSUE #356: [Android] App must support 16 KB memory page sizes
We're getting [this issue from Google Play Console](https://play.google.com/console/u/0/developers/7937367420705076161/app/4973869174659263578/policy-center/issues/4988095825467727584/details), which should be resolved by `Oct 31, 2025` otherwise they won't let us release app updates:

Image

danilosilvackl:
This issue was mostly solved by updating some of those libs. 
then patching one of those 
and lastly adding a `postinstall` script to rebuild one of those libs using a newer version of the `NDK`
NDK version was upgraded from 27.1.99999 to 28.2.999999

I kept track of the work on the GH Issue #356 

But here some good info if we run on the same problem with another lib:
- how to check which lib is _unaligned_ [elf_alignment](https://developer.android.com/guide/practices/page-sizes#elf-alignment)
  - for 16kb page size issue we just need to make sure that `arm64-v8a` and `x86_64` architectures are _aligned_
- updating the packaging if needed ([gradle plugin update](https://developer.android.com/guide/practices/page-sizes#update-packaging))
- how to update the NDK and recompile your libs: [compile for elf alignment](https://developer.android.com/guide/practices/page-sizes#compile-16-kb-alignment)
- how to create a emulator that supports 16kb page size and test using it: [check here](https://developer.android.com/guide/practices/page-sizes#test)
  - if your app isn't 16kb elf compliant it will show you a very aggressive popup when the app opens, so it helps to tell you when to run the script of the first bullet

> [!CAUTION]
> We should review the relevant libs updates in the near future, if they solve it we can update them, if not look into alternatives

Libs affected: 
- [react-native-vision-camera](https://github.com/mrousavy/react-native-vision-camera) (updated)
- [react-native-fast-opencv](https://github.com/lukaszkurantdev/react-native-fast-opencv) (patched)
- [react-native-skia](https://github.com/Shopify/react-native-skia) (updated)
- [react-native-scrypt](https://github.com/Crypho/react-native-scrypt) (manually rebuilt)
- [react-native-keychain](https://github.com/oblador/react-native-keychain) (updated)

 
TODO:
- [ ] record smoke test on android
- [ ] record smoke test on iOS

ps: I already tested the most common flows, adding a token, removing, send and swap but did not record yet

closes #356 
kinda solves #281


scripts/build-scrypt-16kb-aligned.js
```diff
@@ -0,0 +1,227 @@
...
+
+// Build for each architecture
+const architectures = [
+  { name: "arm64-v8a", prefix: "aarch64-linux-android", api: "21" },
+  { name: "armeabi-v7a", prefix: "armv7a-linux-androideabi", api: "21" },
+  { name: "x86", prefix: "i686-linux-android", api: "21" },
+  { name: "x86_64", prefix: "x86_64-linux-android", api: "21" },
+];
+
+for (const arch of architectures) {
```
aristidesstaffieri, line 150: what about extracting this into something like `buildArch` and building the different architectures in parallel? It could be nice if we run this in CI.

scripts/setup-scrypt-16kb.js
```diff
@@ -0,0 +1,62 @@
...
+
+// 3. Try to build 16KB-aligned prebuilt libraries
+console.log("=> Building 16KB-aligned prebuilt libraries...");
+try {
+  const buildScript = path.join(__dirname, "build-scrypt-16kb-aligned.js");
+  execSync(`node "${buildScript}"`, { stdio: "inherit" });
+  console.log("=> 16KB page size alignment setup complete!");
+} catch (error) {
+  // Check if we're in CI environment
+  if (process.env.CI || process.env.GITHUB_ACTIONS) {
```
aristidesstaffieri, line 52: do we still want to exit if this fails in CI? It seemed like we would need this in all future Android builds.

PR #427, opened 2025-09-25
Exponential backoff and max retries for Horizon 504

leofelix077:
### What

[TODO: Short statement about what is changing.]

### Why

[TODO: Why this change is being made. Include any context required to understand
the why.]

### Known limitations

[TODO or N/A]

#### PR structure

- [ ] This PR does not mix refactoring changes with feature changes (break it
      down into smaller PRs if not).
- [ ] This PR has reasonably narrow scope (break it down into smaller PRs if
      not).
- [ ] This PR includes relevant before and after screenshots/videos highlighting
      these changes.

#### Testing

- [ ] These changes have been tested and confirmed to work as intended on
      Android.

#### Release

- [ ] This PR adds JSDocs to new funcionalities.
- [ ] I've checked with the product team if we should add metrics to these
      changes.
- [ ] I've shared relevant before and after screenshots/videos highlighting
      these changes with the design team and they've approved the changes.

### DIFF
--- a/__tests__/services/stellar.test.ts
+++ b/__tests__/services/stellar.test.ts
@@ -2,41 +2,9 @@
  * Tests for stellar service, focusing on submitTx retry logic with exponential backoff
  * This test focuses on the core retry functionality and constants
  */
-import {
-  SUBMIT_BACKOFF_MAX_ATTEMPTS,
-  BASE_BACKOFF_SEC,
-} from "services/stellar";
+import { BASE_BACKOFF_SEC } from "services/stellar";
 
 describe("stellar service - submitTx retry logic", () => {
-  it("should have correct constants for retry configuration", () => {
-    expect(SUBMIT_BACKOFF_MAX_ATTEMPTS).toBe(5);
-    expect(BASE_BACKOFF_SEC).toBe(1000);
-  });
-
-  it("should calculate exponential backoff delays correctly", () => {
-    // Test the exponential backoff calculation: 2^(attempt-1) * BASE_BACKOFF_SEC
-    const delays = [];
-    for (let attempt = 1; attempt <= SUBMIT_BACKOFF_MAX_ATTEMPTS; attempt++) {
-      const delay = 2 ** (attempt - 1) * BASE_BACKOFF_SEC;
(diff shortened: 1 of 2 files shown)


__tests__/services/stellar.test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { BASE_BACKOFF_SEC } from "services/stellar";
+
+describe("stellar service - submitTx retry logic", () => {
+  it("should implement correct delay timing", async () => {
+    jest.useFakeTimers();
+
+    const delays: number[] = [];
+
+    // Simulate the delay logic from submitTx
+    const simulateDelay = (attempt: number) => {
```
aristidesstaffieri, line 14: it would be nice if this was actually a reference to the real backoff logic in the implementation so that the tests would catch regressions there if it changed in the future.

__tests__/services/stellar.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    jest.useRealTimers();
+  });
+
+  it("should verify retry condition logic", () => {
+    // Test the retry condition that matches submitTx implementation
+    // This tests the logic: isHorizonError(e) && e.response.status === 504
+
+    // Mock the isHorizonError function behavior
+    const isHorizonError = (error: any) =>
+      !!(error && error.response && typeof error.response.status === "number");
```
piyalbasu, line 48: why can't we directly test the `isHorizonError` helper in `services/stellar`? Seems like this test is mostly verifying that the mock works, and not that the helper in `services/stellar` is working

__tests__/services/stellar.test.ts
leofelix077: true. I updated now to generate based on the delays and check for the values directly

__tests__/services/stellar.test.ts
leofelix077: we can. I adjusted it now

__tests__/services/stellar.test.ts
leofelix077: oh think it got lost to the void. NOW it's there

PR #437, opened 2025-09-30
[FEATURE] scan swaps

aristidesstaffieri:
Closes stellar/freighter#209
Closes https://github.com/stellar/freighter-mobile/issues/439

### What
New reference for [warnings & transaction detail in swaps](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=8180-16579&t=dnpNYFIWwMYR4hSi-11).
Adds token/transaction scanning to the swap flow.
Adds malicious/suspicious asset and transaction warning elements in the swap flow.
Adds transaction details trigger and sheet in the swap flow.
Updates PathList to match design for representing paths.
Adds new tests for Swap flow, tests scan related states.

### Why
Planned feature as part of the Blockaid epic of work

### Known limitations
N/A

#### PR structure

- [x] This PR does not mix refactoring changes with feature changes (break it
      down into smaller PRs if not).
- [x] This PR has reasonably narrow scope (break it down into smaller PRs if
      not).
- [x] This PR includes relevant before and after screenshots/videos highlighting
      these changes.

#### Testing

- [x] These changes have been tested and confirmed to work as intended on
      Android.

#### Release

- [x] This PR adds JSDocs to new funcionalities.
- [x] I've checked with the product team if we should add metrics to these
      changes.
- [x] I've shared relevant before and after screenshots/videos highlighting
      these changes with the design team and they've approved the changes.

Updates:
simulator_screenshot_B8669705-75B7-4450-BC52-F7E6F30F2153

simulator_screenshot_671BCD85-EB1D-4280-ACA7-3212FA0B8BC8

simulator_screenshot_82E2E31B-AF6D-4367-BB2F-B992C1AEDCB0

Screenshot_1759249984

Screenshot_1759250289

### DIFF
--- a/src/components/screens/SwapScreen/components/SwapReviewBottomSheet.tsx
+++ b/src/components/screens/SwapScreen/components/SwapReviewBottomSheet.tsx
@@ -82,6 +82,8 @@ const SwapReviewBottomSheet: React.FC<SwapReviewBottomSheetProps> = ({
 
   // Use sourceAmount as-is since it's already in the correct format for display
 
+  // Use sourceAmount as-is since it's already in the correct format for display
+
   const currentConversionRate =
     pathResult?.conversionRate ||
     calculateSwapRate(
--- a/src/services/blockaid/api.ts
+++ b/src/services/blockaid/api.ts
@@ -49,7 +49,6 @@ export const scanToken = async (
     const scanResult = response.data.data as Blockaid.TokenScanResponse;
 
     analytics.track(AnalyticsEvent.BLOCKAID_TOKEN_SCAN, {
-      response: scanResult,
       tokenCode,
       network,
     });
@@ -88,7 +87,6 @@ export const scanBulkTokens = async (
     const scanResult = response.data.data as Blockaid.TokenBulkScanResponse;
 
     analytics.track(AnalyticsEvent.BLOCKAID_BULK_TOKEN_SCAN, {
-      response: scanResult,
       addressList,
       network,
     });
@@ -123,7 +121,6 @@ export const scanSite = async (
     const scanResult = response.data.data as Blockaid.SiteScanResponse;
 
     analytics.track(AnalyticsEvent.BLOCKAID_SITE_SCAN, {
-      response: scanResult,
       url,
       network,
     });
@@ -162,7 +159,6 @@ export const scanTransaction = async (
       .data as Blockaid.StellarTransactionScanResponse;
 
     analytics.track(AnalyticsEvent.BLOCKAID_TRANSACTION_SCAN, {
-      response: scanResult,
       xdr,
       url,
       network,


__tests__/components/screens/SwapScreen/SwapAmountScreen.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+import { NativeStackScreenProps } from "@react-navigation/native-stack";
+import SwapAmountScreen from "components/screens/SwapScreen/screens/SwapAmountScreen";
+import Icon from "components/sds/Icon";
+import { SWAP_ROUTES, SwapStackParamList } from "config/routes";
+import { renderWithProviders } from "helpers/testUtils";
+import { useBalancesList } from "hooks/useBalancesList";
+import React, { act } from "react";
+import { View } from "react-native";
+
+import { mockBalances } from "../../../../__mocks__/balances";
```
aristidesstaffieri, line 12: I haven't yet figured out how to get jest to alias this?

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -84,6 +90,12 @@ export const useSwapTransaction = ({
     if (!transactionXDR) {
       throw new Error("Failed to build swap transaction");
     }
+    try {
+      const scanResult = await scanTransaction(transactionXDR, "internal");
+      setTransactionScanResult(scanResult);
+    } catch (error) {
+      logger.error("SwapTransaction", "Transaction scan failed", error);
```
aristidesstaffieri, line 97: if we fail to scan the transaction we log the error, but should this also trigger the "could not scan" state?

One problem we have with this view is that there is 1 banner/sheet to represent 3 scans so we could end up in a situation where we fail to scan the transaction but have a scan result for an asset.

Right now it chooses to show the transaction scan result first, then asset scan results if tx has none.

__mocks__/use-colors.ts
```diff
@@ -0,0 +1,34 @@
+export const mockUseColors = () => {
+  jest.mock("hooks/useColors", () => () => ({
```
piyalbasu, line 2: just curious - i noticed a lot of tests mock this hook. It seems like this hook just returns a static config map. What's the value in mocking it rather than just using the actual hook?

src/components/screens/SwapScreen/components/SwapReviewBottomSheet.tsx
```diff
@@ -155,16 +152,116 @@ const SwapReviewBottomSheet: React.FC<SwapReviewBottomSheetProps> = ({
     }
   };
 
+  const { isMalicious: isTxMalicious, isSuspicious: isTxSuspicious } =
```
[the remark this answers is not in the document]
aristidesstaffieri: 👍 yeah so the hook that gets the scan results [uses the store action](https://github.com/stellar/freighter-mobile/blob/main/src/ducks/balances.ts#L205) and if it fails the default values are used as a fallback.

From the components point of view we also have [unit tests that render it with no scan results.](https://github.com/stellar/freighter-mobile/pull/437/files#diff-f36cbf40e5338e5e50c4eb622c1a47d543efae3b29b646a222d3b3c87877108fR110)

CassioMG:
FYI, it looks like @danilosilvackl created the #439 issue on the `freighter` repo instead of `freighter-mobile` repo so I just transferred the issue to `freighter-mobile` and assigned to Aristides. It was quite easy, there was a "transfer issue" button which allowed me to transfer it in case it happens again in the future.

src/config/analyticsConfig.ts
```diff
@@ -37,6 +37,7 @@ export enum AnalyticsEvent {
   VIEW_SWAP_TIMEOUT = "loaded screen: swap timeout",
   VIEW_SWAP_SETTINGS = "loaded screen: swap settings",
   VIEW_SWAP_CONFIRM = "loaded screen: swap confirm",
+  VIEW_SWAP_TRANSACTION_DETAILS = "loaded screen: swap transaction details",
```
CassioMG, line 40: thanks for checking and adding the analytics for the swap transaction details 🚀 

should we also have one for send transaction details?

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
aristidesstaffieri: well so I think we really only need to scan both assets and transactions because of our UX flow.

We need to scan assets before we build a transaction in order to display the warning icon on the assets in the preview. Once a user configures a transaction then we can scan the transaction as well since the transaction could be flagged for non asset related reasons. 

To answer your question though, I believe that a transaction scan would include non-benign results in it for assets as well and they would be [in this map](https://github.com/blockaid-official/blockaid-client-node/blob/d6576b02ad059747eb5aaaccef43b265d1187225/src/resources/stellar/stellar.ts#L528) but it's hard to find a real swap to test this since there aren't a lot of pools with liquidity for flagged assets.

src/config/analyticsConfig.ts
aristidesstaffieri: ah yeah good call, I added this in 8d663c3f4b6a5392ddc2cbb34144c13e8199fc6b

PR #499, opened 2025-10-24
[BUG] Fixes banner usage in Collectible review

aristidesstaffieri:
### What
Fixes a bug introduced with the work to enable sending a collectible.
Extracts common logic between token and collectible review.
Adds test to cover common helpers and the banner variants.

### Why
CollectibleReview did not properly set the bannerText and variant, leading to a confusing screen in cases where the scan has a result.

### Known limitations
N/A

#### PR structure

#### Testing

- [] These changes have been tested and confirmed to work as intended on Android.
- [x] These changes have been tested and confirmed to work as intended on small iOS screens.
- [] These changes have been tested and confirmed to work as intended on small Android screens.

#### Release

### DIFF
--- a/src/components/screens/SendScreen/helpers/index.tsx
+++ b/src/components/screens/SendScreen/helpers/index.tsx
@@ -26,7 +26,7 @@ export function useTransactionSecurity(
     ) {
       const warnings = extractSecurityWarnings(transactionScanResult);
 
-      if (Array.isArray(warnings) && warnings.length > 0) {
+      if (warnings.length > 0) {
         return warnings;
       }
     }


src/components/screens/SendScreen/helpers/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  );
+
+  const transactionSecurityWarnings = useMemo(() => {
+    if (
+      transactionSecurityAssessment.isMalicious ||
+      transactionSecurityAssessment.isSuspicious
+    ) {
+      const warnings = extractSecurityWarnings(transactionScanResult);
+
+      if (Array.isArray(warnings) && warnings.length > 0) {
```
piyalbasu: Looks like `extractSecurityWarnings` is typed to return an Array. Should this type be updated if it can return anything other than an Array?

piyalbasu: If it's guarding against a type error or something similar, I think the helper should handle that so the caller has a reliable output

src/components/screens/SendScreen/helpers/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  onMemoMissingPress,
+}: UseSendBannerContentParams): BannerContent | undefined {
+  const { t } = useAppTranslation();
+
+  return useMemo(() => {
+    const shouldShowNoticeBanner =
+      isRequiredMemoMissing || isMalicious || isSuspicious;
+
+    if (!shouldShowNoticeBanner) {
+      return undefined;
```
piyalbasu: this is so NIT, but I feel `null` is perhaps a better return value as `null` generally implies that this is intentionally set as empty (which is what we're doing here) where as `undefined` generally implies we've yet to set a value

PR #514, opened 2025-10-29
Fix Send Fiat Amount

### LINKED ISSUE #480: [Send] Editing dollar amount doesn’t work
### What version are you using?

1.6.23 (31)

### What did you do?

Opened the Send amount and tried inputting a DOLLAR amount right away (you need to tap on the little “swap” button below the amount number to input a dollar amount).

I can’t always input the dollar amount.

### What did you expect to see?

The amount reflect what I am typing.

### What did you see instead?

Nothing happens.

Image

leofelix077:
Adjust templating of Fiat amount when switching from entering token amount
Closes #480 

https://github.com/user-attachments/assets/9feff34a-6e19-4246-a590-a53993796800

#### PR structure

#### Testing

- [ ] ~These changes have been tested and confirmed to work as intended on small iOS screens.~
- [ ] ~These changes have been tested and confirmed to work as intended on small Android screens.~

#### Release

### DIFF
--- a/src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
+++ b/src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
@@ -719,6 +719,7 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
               </Text>
               <TouchableOpacity
                 className="ml-2"
+                hitSlop={10}
                 onPress={() => setShowFiatAmount(!showFiatAmount)}
               >
                 <Icon.RefreshCcw03


src/hooks/useTokenFiatConverter.ts
```diff
@@ -15,12 +16,14 @@ interface UseTokenFiatConverterProps {
...
-  fiatAmount: string;
+  fiatAmount: string; // Internal value (dot notation)
+  fiatAmountDisplay: string; // Display value (locale-formatted)
   showFiatAmount: boolean;
   setShowFiatAmount: (show: boolean) => void;
   handleDisplayAmountChange: (key: string) => void;
   setTokenAmount: (amount: string) => void;
   setFiatAmount: (amount: string) => void;
+  updateFiatDisplay: (amount: string) => void;
 }
```
aristidesstaffieri: one thing that sticks out to me about this hook is that it uses multiple related but separate state variables to trigger a cascade of effects in order to reach the correct state.

`tokenAmount → fiatAmount → fiatAmountDisplay`

This works but leaves a path them for them to fall out of sync. What about refactoring this into a `useReducer` workflow that can provide the state changes atomically?

I think if you really prefer the semantics of `useState` instead we could consider deriving the display values from the real amounts directly(memo value) instead of storing them in state. This would reduce the number of renders needed for this update.

leofelix077:
@CassioMG just quick update: saw your comment, just will take a bit to get back to it due to the other https://github.com/stellar/freighter-mobile/issues/520 issue. but thanks a lot for the scenarios. will check them out and post videos of it as soon I can get to it

src/hooks/useTokenFiatConverter.ts
leofelix077: To be fair I had forgotten about the useReducer/reducers in general. it went to the back of my mind, as we don't use it anywhere else in the code and I hadn't seen it in a while, but I think it's a much better approach for manipulating multiple values at once

PR #600, opened 2025-12-03
[Collectibles] Hide Collectible

### LINKED ISSUE #502: Hide Collectibles
Users should be able to hide/unhide collectibles.

Note: for the "Successfully hid collectible" notification we should use our existing component instead of the now shown on Figma.

[Figma](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=6597-168742&t=jtTZA9y1VVeP4c3V-1):

Image

CassioMG:
### What

Closes #502 

This PR implements the "Hide Collectible" functionality.

It adds a `isHidden` property for the `Collectible` interface, persists the hidden collectibles in `local storage` and add new `functions to hide/unhide a collectible` in addition to a `useFilteredCollectibles` hook which serves the `visibleCollectibles` and `hiddenCollectibles` lists separately so it can be easily accessible anywhere like in places where we have `CollectiblesGrid` being rendered.

Users should still be able to take any action on a hidden collectible as normal like sending it to another user, viewing it, saving it to the camera roll, etc.

This PR also fixes a bug where the tokens and collectibles lists were not scrollable when rendered from the Send flow.

### Why

So users are able to hide collectibles if they want to.

Note: please ignore the `Remove collectible` options from images and videos as it's there only for local development.

Screenshot 2025-12-02 at 17 36 01

Screenshot 2025-12-02 at 17 38 12

### Hiding and Showing a collectible

[hiding-showing-collectible.webm](https://github.com/user-attachments/assets/8449c6c6-ecb1-48a4-afbc-f5f598fd22c8)

### Sending a Hidden collectible

[sending-hidden-collectible.webm](https://github.com/user-attachments/assets/4e0b8945-e518-4af1-854c-2845a0fbb278)

### Tokens and Collectibles scroll bug fix on Send flow

[send-scroll-bug-fix.webm](https://github.com/user-attachments/assets/23660935-e921-478e-bd97-690403cf0d65)

### Known limitations

The hidden collectibles info is only stored locally, so if users delete the app or use another device they will need to hide collectibles all again. This is probably not a big issue during this initial adoption phase, but in the future we could revisit it and assess the work to save this info on the backend in case we think that makes sense.

#### PR structure

#### Testing

#### Release

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -125,7 +125,7 @@ jobs:
     timeout-minutes: 60
 
     steps:
-      - name: Display macOS Image Information
+      - name: Display macOS Image Information and Xcode versions
         run: |
           echo "=== macOS Runner Information ==="
           echo "Runner OS: ${{ runner.os }}"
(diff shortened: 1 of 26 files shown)


src/ducks/collectibles.ts
```diff
@@ -683,4 +724,108 @@ export const useCollectiblesStore = create<CollectiblesState>((set, get) => ({
...
+    tokenId,
+  }: {
+    publicKey: string;
+    network: string;
+    contractId: string;
+    tokenId: string;
+  }) => {
+    try {
+      // Add to hidden collectibles storage
+      await addHiddenCollectibleToStorage({
```
piyalbasu: related to my comment here: https://github.com/stellar/freighter-mobile/pull/600/files#r2586427355

If the `publicKey` or `publicKey[network]` is not found, `addHiddenCollectibleToStorage` will silently return, but your code will continue and `updatedCollections` down below will reflect something different than what is in storage

src/helpers/collectibles.ts
```diff
@@ -291,3 +333,191 @@ export const transformBackendCollections = async (
...
+  tokenId: string;
+}): Promise<void> => {
+  const { network, publicKey, contractId, tokenId } = params;
+
+  try {
+    const storage = await getHiddenCollectiblesStorage();
+
+    // Check if the user has any hidden collectibles for this network
+    if (!storage[publicKey] || !storage[publicKey][network]) {
+      return;
```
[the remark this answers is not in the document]
CassioMG: makes sense, just fixed it here: https://github.com/stellar/freighter-mobile/pull/600/commits/d7798ae91d02abea7ab595a598391e2323ab9616

src/ducks/collectibles.ts
piyalbasu: Ah sorry, you're right - I misread what function was being called here. Disregard my comment

PR #613, opened 2025-12-10
[CHORE] 8347531

aristidesstaffieri:
### What

[TODO: Short statement about what is changing.]

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

#### PR structure

- [ ] This PR does not mix refactoring changes with feature changes (break it down into smaller PRs if not).
- [ ] This PR has reasonably narrow scope (break it down into smaller PRs if not).

#### Testing

#### Release


src/helpers/deviceSecurity.ts
```diff
@@ -0,0 +1,10 @@
+import JailMonkey from "jail-monkey";
+
+export const isDeviceJailbroken = (): boolean => {
+  try {
+    return JailMonkey.isJailBroken();
+  } catch (error) {
+    // Fail closed - treat errors as jailbroken device
+    return true;
```
CassioMG: are we confident on defaulting it to `true` here?

if anything happens with the lib which ends up resulting in an error then suddenly all users will be locked out of the app. Or if let's say some devices doesn't support it, then some users will be locked out and it may take some time until we realize this is happening and deploy a fix

src/navigators/RootNavigator.tsx
```diff
@@ -90,6 +93,15 @@ export const RootNavigator = () => {
     const initializeApp = async (
       postInitializeCallback?: () => void,
     ): Promise<void> => {
+      const deviceCompromised = isDeviceJailbroken();
+      setIsJailbroken(!deviceCompromised);
```
CassioMG: is this intended to set it like `!deviceCompromised` instead of `deviceCompromised` ?

on line 159 it's also using `!isJailbroken` instead of `isJailbroken`

if we remove the `!` on both places I believe it'd look a bit more intuitive. Wdyt?

src/navigators/RootNavigator.tsx
```diff
@@ -144,6 +156,10 @@ export const RootNavigator = () => {
     return <LoadingScreen />;
   }
 
+  if (!isJailbroken) {
+    return <SecurityBlockScreen />;
+  }
+
```
CassioMG: I'd suggest moving it to line 154 before the loading screen just to play safe so we assure there isn't any screen higher than `SecurityBlockScreen` in priority

aristidesstaffieri: updated in 11651690540da3fc6fa541bfb2202068d5737dba

src/helpers/deviceSecurity.ts
```diff
@@ -12,7 +12,7 @@ export const isDeviceJailbroken = (): boolean => {
     return isJailBroken;
   } catch (error) {
     const errorMessage = error instanceof Error ? error.message : String(error);
-    Sentry.captureMessage(`[isJailBroken] ${errorMessage}`, "error");
+    logger.error("isDeviceJailbroken", "isJailBroken", errorMessage);
```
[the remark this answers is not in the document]
aristidesstaffieri: 👍 updated in 647e9225d9bc8cbac4f64e6b0293a62d9f47e2f4

Issue #607, opened 2025-12-05
[QA 1.9.24] Interaction UX Updates


sdfcharles:
- once you’ve added a token you should be taken back to `Manage Tokens` instead of `Add a Token` again
- on `swap` (and other areas where you have multiple sheets overlaid) instead of one going down and the other popping up, can the second sheet just go over the first? that way when the second sheet is closed, the first sheet is there and doesnt have to reanimate in
- on `settings`, when i click save, the sheet goes down and then the keyboard goes down. can the keyboard go down with the sheet instead?

Issue #608, opened 2025-12-05
[QA 1.9.24] Investigate Token Icons


sdfcharles:
<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/fadf2b4a-566c-4609-97c9-c5e9800c77b0" />

<img width="758" height="1652" alt="Image" src="https://github.com/user-attachments/assets/81f57fda-a2c5-4b4b-8057-9fade6b8a46a" />

- when i add ARST, i just see a blank circle instead of `AR` as the placeholder or the actually token asset (on the extension, for ARST its stuck on a ghost loading state)
- when i swap XLM for BLND, the blend icon shows correctly in history, but when i swap BLND for XLM, the image is not shown, just the placeholder `BL`

---
<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/d7cc6e71-2e83-4b25-bca5-c2cb97536888" />

unrelated but small design tweak
- on `token details` can we update the refresh button to a text button like the design [here](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=8510-17726&t=ZqyZru4YB3pIQMit-1)?

PR #660, opened 2026-01-08
[QA 1.9.24] Fixes and improvements

CassioMG:
### What

Closes https://github.com/stellar/freighter-mobile/issues/607 and https://github.com/stellar/freighter-mobile/issues/608 which were created based on v1.9.24 QA findings.

#### After adding a token navigate back to `Manage Tokens` instead of `Add a Token` again:

https://github.com/user-attachments/assets/c093efe2-4288-40de-8078-18cb01a5e3b4

#### Upon saving transaction settings dismiss both the sheet and the keyboard at the same time:

https://github.com/user-attachments/assets/418f4283-6044-4f1b-84cc-0ff9ee498472

#### Correctly display both token icons on swap history items:
* [This is the commit](https://github.com/stellar/freighter-mobile/pull/660/changes/336018c6fe8634559d7194d3df673e1bf2fac202) that fixed it on this PR, but it turns out this has been already fixed on main on [this other recent commit](https://github.com/stellar/freighter-mobile/commit/547490301147b552317d8fc60ac1f6058a9c0093#diff-5cf0e7037333a2c78f226dad1e4d23775fa0cdac17a7c28371317d16b52055cfL79)

Screenshot 2026-01-08 at 15 34 09

#### Fallback to 2-letter icon when token icon image fails to load:

Screenshot 2026-01-08 at 16 06 57

### Why

So the app offers a better UX.

### Known limitations

In regards to the [point below](https://github.com/stellar/freighter-mobile/issues/607):

_- on swap (and other areas where you have multiple sheets overlaid) instead of one going down and the other popping up, can the second sheet just go over the first? that way when the second sheet is closed, the first sheet is there and doesnt have to reanimate in_

It is possible to implement it but it is not a trivial change given how the sheet modal works and how sheets are being currently nested. We'd need to do some refactor and re-test all places that use multiple sheets in order to make sure it won't add any unexpected bugs. I've created [this ticket](https://github.com/stellar/freighter-mobile/issues/661) as a follow-up.

#### PR structure

#### Testing

#### Release

### DIFF
--- a/__tests__/components/sds/Token.test.tsx
+++ b/__tests__/components/sds/Token.test.tsx
@@ -1,5 +1,8 @@
-import { render } from "@testing-library/react-native";
+/* eslint-disable no-promise-executor-return */
+/* eslint-disable @typescript-eslint/require-await */
+import { act, fireEvent, render, waitFor } from "@testing-library/react-native";
 import { Token } from "components/sds/Token";
+import { Text } from "components/sds/Typography";
 import React from "react";
 
 /**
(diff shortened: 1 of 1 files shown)


sdfcharles:
nice! changes look great, animation is smooth and the history looks a lot more rich now 🙏

src/components/sds/Token/index.tsx
```diff
@@ -334,6 +336,71 @@ const TokenImage = styled.Image`
...
+  const handleImageLoad = () => {
+    imageLoadedRef.current = true;
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = null;
+    }
+  };
+
+  // If we have a valid image, show the image
+  if (hasValidImage && !showFallback) {
```
aristidesstaffieri: wdyt about adding a unit test asserting the fallback behavior?

src/components/TokenIcon.tsx
```diff
@@ -136,29 +135,16 @@ export const TokenIcon: React.FC<TokenIconProps> = ({
...
       <TokenComponent
         variant="single"
         size={size}
         sourceOne={{
           altText: t("tokenIconAlt", { code: token.code }),
           backgroundColor,
+          image: imageUrl,
+          // Fallback: show SorobanTokenIcon for Soroban custom tokens
+          // in case the specific icon is not available
           renderContent: () => <SorobanTokenIcon />,
```
piyalbasu: Correct me if I'm wrong, but looking at the code for `<Token`, it looks like `renderContent` takes precedent over `image`. So won't this show `SorobanTokenIcon` every time and disregard `imageUrl`?

https://github.com/stellar/freighter-mobile/blob/main/src/components/sds/Token/index.tsx#L439

CassioMG: That's correct for the main branch but this PR changes that precedence. It now gives priority to `imageUrl` and fallback to `renderContent` if imageUrl doesn't exist or fail to load.

I've changed the precedence for simplification since I noticed all parent components that were using the `Token` component were already giving priority to `imageUrl` over `renderContent` so I figured it'd make sense to just change `Token` behavior to align with that so we could simplify the code.

src/components/sds/Token/index.tsx
```diff
@@ -334,6 +336,71 @@ const TokenImage = styled.Image`
...
+  const IMAGE_LOAD_TIMEOUT = 1000;
+
+  // Check if image is valid (non-empty string or valid ImageSourcePropType)
+  const hasValidImage =
+    source.image &&
+    (typeof source.image === "string"
+      ? source.image.trim().length > 0
+      : !!source.image);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
CassioMG: yeah, I was kind of in the fence on adding a loader here to prevent having lots of rows rendering the loader at the same time for a short period and potentially affect performance as it's an animation, and not loading the imageUrl should be an edge case as most tokens have it working. But I'll give it a try and let you know

aristidesstaffieri (approved):

PR #704, opened 2026-01-30
Improve token icon URL validation and caching

### LINKED ISSUE #700: Move token icon URL validation to tokenIcons duck
# Move token icon URL validation from TokenIcon component to tokenIcons duck

## Context

The `TokenIcon` component is used across the app (history list, history details, balances list, balance details, etc.) to render token icons. The current flow:

1. `tokenIcons` duck fetches and caches icon URLs (may be empty strings)
2. `TokenIcon` reads URLs from the store and passes them to `TokenComponent`
3. `TokenComponent` uses `ImageWithFallback` to validate each URL asynchronously during render
4. Each `ImageWithFallback` instance manages its own loading state, timeout, and fallback logic

**Additional context:**
- `cacheTokenListIcons` runs on every app launch (via `useFetchTokenIcons` hook in `TabNavigator`) after a 5s delay
- It caches icon URLs from verified token lists, which can include hundreds of tokens
- `refreshIcons` also runs on every app launch (if 24 hours have passed since last refresh)
- It processes all cached icons (could be hundreds from token lists + user's tokens) in batches to fetch updated URLs
- Both functions already perform network requests to fetch/cache URLs
- Most of these tokens may never be used by the user
- If we validate all icons in `cacheTokenListIcons` or `refreshIcons`, it could trigger hundreds of additional network requests on every app launch, causing performance issues and unnecessary battery/network usage

## Problem

We're seeing blank images or 2-letter icon for tokens even when:
- The token has a valid icon URL
- The token is native XLM (which has a local image file)

Image

Root cause: delegating async image validation to the `TokenIcon` component causes issues when many tokens render simultaneously:
- Race conditions from concurrent async validation
- Each component validates independently, causing duplicate work
- Even local images (XLM) can show blank due to async validation timing
- No shared validation state across components

Performance concern: if we validate all icons from token lists or during refresh on app launch, we'd validate hundreds of icons that may never be used, causing:
- Unnecessary network requests (doubling the work since we already fetch URLs)
- Performance degradation (even with batching, validating hundreds of icons takes time)
- Battery drain
- Poor user experience
- Network costs

## Goal

Move image URL validation responsibility from `TokenIcon` to the `tokenIcons` duck, which can:
- Validate URLs lazily in the background (only when actually needed)
- Cache validation results to avoid duplicate work
- Mark invalid URLs so `TokenIcon` can show fallback immediately
- Prioritize validation for tokens the user actually has (in balances) over token list icons
- Avoid validating icons in `cacheTokenListIcons` and `refreshIcons` (performance optimization)

This makes `TokenIcon` synchronous and removes async validation from the render path, while ensuring we only validate icons that are actually used, avoiding performance issues on app launch.

## Implementation Plan

*Note: This is an `AI-generated` implementation plan, please review it and adjust as needed.

### Phase 1: Extend tokenIcons duck with validation state

1. Update `Icon` interface in `src/ducks/tokenIcons.ts`:
   - Add `isValidated: boolean` (whether URL was validated)
   - Add `isValid: boolean | null` (null = not validated, true = valid, false = invalid)
   - Keep existing `imageUrl` and `network` fields

2. Add validation function:
   - Create `validateIconUrl(url: string): Promise<boolean>`
   - Use React Native's `Image.prefetch()` or a HEAD request
   - Implement timeout (2-3 seconds)
   - Return `true` if image loads successfully, `false` otherwise

3. Update `fetchIconUrl`:
   - After fetching URL, validate it in the background (don't block)
   - Store with `isValidated: true` and `isValid: <result>`
   - If invalid, set `imageUrl: ""` or keep URL but mark `isValid: false`

4. Add lazy validation trigger:
   - Create `validateIconOnAccess(identifier: string)` function
   - Triggers validation when an icon is accessed but not yet validated
   - Non-blocking: returns immediately, updates store when validation completes

### Phase 2: Migrate existing cache entries (Lazy Validation Strategy)

1. Migration strategy:
   - On app start, mark all existing cached icons as `isValidated: false`
   - Do not validate them immediately
   - Validate only when icons are actually accessed

2. Update `cacheTokenListIcons`:
   - Continue caching URLs from token lists (as it does now)
   - Mark all new icons as `isValidated: false`, `isValid: null`
   - Do not validate icons in `cacheTokenListIcons` (performance optimization)
   - This prevents validating hundreds of unused icons on every app launch

3. Validation priority (when icons are accessed):
   - Priority 1: Icons for tokens in user's balances (validate immediately on access)
   - Priority 2: Icons for tokens in token lists (validate in background, very small batches)
   - Priority 3: Icons that were never accessed (don't validate at all)

### Phase 3: Simplify TokenIcon component (Lazy Validation on Access)

1. Remove async validation logic:
   - Make `TokenIcon` fully synchronous
   - When reading icon from store:
     - If icon exists AND `isValid === true`, use `imageUrl`
     - If icon exists BUT `isValidated === false`, trigger background validation (non-blocking) and show fallback immediately
     - If icon doesn't exist, `isValid === false`, or `imageUrl === ""`, show fallback immediately
   - Once validation completes, component will re-render with validated URL (if valid)

2. Handle special cases:
   - Native XLM: continue using `logos.stellar` directly (no validation needed)
   - Local images (like USDC): mark as pre-validated in duck (`isValidated: true`, `isValid: true`)

3. Add validation trigger:
   - When `TokenIcon` reads an unvalidated icon, call `validateIconOnAccess(identifier)`
   - This triggers background validation without blocking the UI

### Phase 4: Simplify TokenComponent/ImageWithFallback

1. Simplify `ImageWithFallback` in `src/components/sds/Token/index.tsx`:
   - Remove timeout logic (validation already done in duck)
   - Remove loading state/spinner (images are pre-validated)
   - Keep minimal error handling as safety net
   - If image fails to load unexpectedly, show fallback

2. Alternative: Consider removing `ImageWithFallback` entirely:
   - If all URLs are pre-validated, use React Native `Image` directly
   - Keep minimal error handling

### Phase 5: Update refresh logic (Lazy Validation Strategy)

1. Update `refreshIcons`:
   - Continue fetching new URLs for cached icons (as it does now)
   - Mark all refreshed icons as `isValidated: false`, `isValid: null`
   - Do not validate icons in `refreshIcons` (performance optimization)
   - This prevents validating hundreds of icons that may not be used
   - Validation will happen lazily when icons are accessed
   - The function already processes icons in batches to avoid overwhelming the network; adding validation would double the network work

2. Update `fetchBalancesIcons`:
   - After fetching icons for balances, validate them immediately (these are high priority)
   - These are tokens the user actually has, so validation is worth the cost
   - Mark as `isValidated: true` with validation result

3. Update `cacheTokenListIcons`:
   - Continue caching URLs from token lists (as it does now)
   - Mark all new icons as `isValidated: false`, `isValid: null`
   - Do not validate icons in `cacheTokenListIcons` (performance optimization)
   - This prevents validating hundreds of unused icons on every app launch

### Phase 6: Handle edge cases

1. Network failures:
   - If validation fails due to network issues, mark as "pending validation"
   - Retry validation later (maybe on next access)
   - Don't show blank; show fallback until validated

2. URL changes:
   - If token's icon URL changes, re-validate
   - Invalidate old cache entries

3. Local images:
   - Pre-validate local images (XLM, USDC) immediately
   - Mark as `isValid: true` without network check

## Benefits

- Performance: validation happens once per URL, not per render
- Efficiency: only validates icons that are actually used, not hundreds of unused token list icons
- Network optimization: avoids doubling network work in `refreshIcons` and `cacheTokenListIcons`
- Reliability: eliminates race conditions with shared validation state
- UX: immediate fallback for invalid URLs, no blank images
- Simplicity: `TokenIcon` becomes synchronous and easier to reason about
- Scalability: lazy validation handles many tokens efficiently without impacting app launch

## Considerations

- Initial load: first-time tokens may show fallback until validation completes (acceptable)
- Storage: validation state slightly increases store size (acceptable trade-off)
- Migration: need to handle existing cache entries gracefully
- Testing: thoroughly test validation logic, edge cases, and migration path
- Performance: lazy validation ensures no performance impact on app launch from `cacheTokenListIcons` and `refreshIcons`

## Acceptance Criteria

- [ ] `tokenIcons` duck validates URLs lazily (only when accessed)
- [ ] `cacheTokenListIcons` does not validate icons (performance optimization)
- [ ] `refreshIcons` does not validate icons (performance optimization)
- [ ] `TokenIcon` component is synchronous (no async validation)
- [ ] No blank images shown for tokens with valid URLs
- [ ] XLM and USDC tokens always shows correctly (local image)
- [ ] Invalid URLs show fallback immediately
- [ ] Existing cached icons are migrated properly
- [ ] Icons for user's balances are validated with priority
- [ ] Network failures are handled gracefully
- [ ] Tests cover validation logic and edge cases
- [ ] No performance degradation on app launch

## Files to Modify

- `src/ducks/tokenIcons.ts` - Add validation logic with lazy validation
- `src/components/TokenIcon.tsx` - Simplify to be synchronous, trigger lazy validation
- `src/components/sds/Token/index.tsx` - Simplify `ImageWithFallback`
- `src/hooks/useFetchTokenIcons.ts` - May need updates for validation triggers

## Related Files (for reference)

- `src/helpers/getIconUrl.ts`
- `src/helpers/getIconUrlFromIssuer.ts`
- `src/helpers/getIconUrlFromTokensLists.ts`
- `src/ducks/verifiedTokens.ts`

leofelix077:
Closes #700 

https://github.com/user-attachments/assets/c7c53355-cee3-45b1-85cf-1be6126443ad

https://github.com/user-attachments/assets/51f69c3c-ee1f-4732-b8b8-ed168b4e2b2f

https://github.com/user-attachments/assets/3acc226c-d249-4ce4-ba3e-fd70eb48ff9c

#### PR structure

#### Testing

#### Release

### DIFF
--- a/src/components/TokenIcon.tsx
+++ b/src/components/TokenIcon.tsx
@@ -182,8 +182,13 @@ const TokenIconWithStore: React.FC<{
     (state) => state.validateIconOnAccess,
   );
 
-  const failedTokenCodes = useTokenIconsStore(
-    (state) => state.failedTokenCodes,
+  // Only select this specific token's failure status to avoid re-rendering
+  // all instances whenever any token fails validation
+  const isTokenFailed = useTokenIconsStore(
+    React.useCallback(
+      (state) => !!state.failedTokenCodes[tokenIdentifier],
+      [tokenIdentifier],
+    ),
   );
 
   const hasIconToValidate =
@@ -192,8 +197,7 @@ const TokenIconWithStore: React.FC<{
     !!icon.imageUrl &&
     icon.isValid !== false;
 
-  const shouldValidateIcon =
-    !iconUrl && hasIconToValidate && !failedTokenCodes[tokenIdentifier];
+  const shouldValidateIcon = !iconUrl && hasIconToValidate && !isTokenFailed;
 
   const lastValidImageUrl = icon?.lastValidImageUrl;
 
@@ -208,7 +212,7 @@ const TokenIconWithStore: React.FC<{
     shouldValidateIcon,
     tokenIdentifier,
     validateIconOnAccess,
-    failedTokenCodes,
+    isTokenFailed,
   ]);
 
   // When validation explicitly failed and no prior valid URL exists, pass
--- a/src/components/sds/Token/index.tsx
+++ b/src/components/sds/Token/index.tsx
@@ -404,8 +404,8 @@ const TokenLoader: React.FC = () => {
 //
 //   idle     → No URL available.   Show: fallback only.
 //   loading  → URL present, image fetching.  Show: fallback + loader overlay.
-//   loaded   → onLoadEnd confirmed.  Show: image only.
-//   fallback → Load failed / timed out.  Show: fallback only.
+//   loaded   → onLoad callback fired (success).  Show: image only.
+//   fallback → onError fired or timeout.  Show: fallback only.
 //              <TokenImage> is removed from the tree to prevent React Native
 //              Image from retrying a failed URL autonomously.
 //
@@ -471,8 +471,8 @@ function iconLoadReducer(
 }
 
 // Shared style for hiding the image element while it loads (prevents a white
-// flash before onLoadEnd fires, while keeping the element in the tree so that
-// onLoadEnd / onError callbacks are active).
+// flash before onLoad fires, while keeping the element in the tree so that
+// onLoad / onError callbacks are active).
 const hiddenImageStyle = StyleSheet.create({
   hidden: { opacity: 0 },
 });
@@ -585,7 +585,7 @@ const ImageWithFallback: React.FC<{
   // -------------------------------------------------------------------------
   useEffect(() => {
     // Local assets: always transition to loading so <TokenImage> renders;
-    // onLoadEnd fires immediately and advances to loaded.
+    // onLoad fires immediately and advances to loaded.
     if (isLocalAsset) {
       if (stateRef.current.phase !== IconPhase.LOADED) {
         dispatch({ type: IconActionType.START_LOADING, url: LOCAL_ASSET_URL });
@@ -621,7 +621,7 @@ const ImageWithFallback: React.FC<{
       return undefined;
     }
 
-    // Hard timeout: if onLoadEnd hasn't fired within the window, show fallback.
+    // Hard timeout: if onLoad hasn't fired within the window, show fallback.
     const timeoutId = setTimeout(() => {
       if (stateRef.current.phase === IconPhase.LOADING) {
         dispatch({ type: IconActionType.TIMEOUT });
--- a/src/ducks/tokenIcons.ts
+++ b/src/ducks/tokenIcons.ts
@@ -622,6 +622,13 @@ export const useTokenIconsStore = create<TokenIconsState>()(
     {
       name: TOKEN_ICONS_STORAGE_KEY,
       storage: createJSONStorage(() => AsyncStorage),
+      partialize: (state) => ({
+        // Only persist icons and lastRefreshed; failedTokenCodes is transient
+        // and should reset with each app session to allow recovery from
+        // temporary validation failures
+        icons: state.icons,
+        lastRefreshed: state.lastRefreshed,
+      }),
     },
   ),
 );


CassioMG:
@leofelix077 I've found 2 issues while testing on latest iOS Dev build:

1 - Icons kept blinking in loop after app update, but after some time apparently it fixed by itself and never happened again. I wonder if this behavior happened as a result from migrating from old cache structure to new cache structure. This one seems less critical as it stopped happening after a while:

https://github.com/user-attachments/assets/30976747-c7f9-49f1-b02d-13ef0e2e5622

2 - It appears the ARST icon fallback works correctly for the big icon version on balances list or tokens list in swap. But the smaller version on history rows does not load or take a lot of time to load. This seems unexpected behavior as I'd expect both the balances icons and history row icons to load from the same cache, so not sure why it behaves differently on history row compared to other places. 
2.1 - It also appears that each screen that have the small history icon loads it separately. Which means even after 1 screen loads the small icon the other screens/bottomsheets it still show the loading state forever for ARST.

Could we fix it so the fallback works quickly on the history rows?

https://github.com/user-attachments/assets/6aaa6649-9087-4d9d-97dd-9cff1cda7cd3

src/helpers/__tests__/validateIconUrl.test.ts
```diff
@@ -0,0 +1,285 @@
+import { validateIconUrl } from "helpers/validateIconUrl";
+import { Image } from "react-native";
+
```
CassioMG: can you please move this test tile to the main __tests__ folder? thanks

CassioMG:
@leofelix077 the USDC icon (which is special cased locally) is not appearing for me in any screen now after deleting and re-installing the app. I've quit and re-launch the app a few times but it's not fixing it. The loading of other icons appear to be working smoothly, so it's probably something related to the USDC special casing.

IMG_7663
IMG_7661
IMG_7660
IMG_7662

PR #774, opened 2026-03-18
Display better fee breakdown for soroban transactions

### LINKED ISSUE #768: Improve minResourceFee display in Send flow
Similar to https://github.com/stellar/freighter/issues/2475

in the app, we also don't display all the fees information for soroban transactions. The data is somewhat confusing, and not clear what are the total fees

Image
Image
Image

Image

leofelix077:
Already verified UI with @sdfcharles on build **1.13.25 (1774353202)**

Screenshot 2026-03-24 at 12 15 13
Screenshot 2026-03-24 at 11 50 24
Screenshot 2026-03-24 at 11 50 29
Screenshot 2026-03-24 at 11 50 35
Screenshot 2026-03-24 at 11 50 44

#### PR structure

#### Testing

#### Release


CassioMG:
## PR Review: Display better fee breakdown for soroban transactions (Mobile)

Cross-reference: This review covers this PR alongside the extension counterpart at stellar/freighter#2649. A comparison section is included at the bottom.

---

### Individual Review of Mobile PR

**Overall**: Well-structured feature with good separation of concerns. The `FeeBreakdownBottomSheet` component is clean, the Zustand store changes are logical, and the early fee estimation with amount `"0"` is a nice UX touch.

#### Things I Like
- **Auto-simulation with amount `"0"`** for early fee estimation — lets users see the fee breakdown before entering an amount. This is a UX advantage over the extension.
- **`extractErrorMessage` helper** — properly handles plain-object API errors that would otherwise render as `[object Object]`.
- **Request ID race-condition guard** — correctly prevents stale async responses from overwriting newer state.
- **`lastAutoSimulatedKey` ref pattern** — clean way to prevent re-triggering simulation loops.
- **`isSorobanContext` prop** on `FeeBreakdownBottomSheet` — ensures correct Soroban description text even before simulation completes (loading state), which is more correct than inferring from data.
- Good test coverage for the new store fields and simulation response shape changes.

#### Issues & Suggestions

**1. Potential fragility in `TransactionAmountScreen` auto-simulation effect** ⚠️
The `useEffect` at ~line 640 depends on `[selectedBalance, recipientAddress, isBuilding, prepareTransaction]`. When `prepareTransaction` runs, it sets `isBuilding: true` → then `false`, which re-triggers the effect. The `lastAutoSimulatedKey` ref guards against this, but `prepareTransaction` is recreated on every render (it's a `useCallback` with many deps including `tokenAmount`). If `tokenAmount` changes, `prepareTransaction` identity changes, `lastAutoSimulatedKey` still matches so it won't re-fire — but it's fragile. Consider removing `isBuilding` from the dependency array since the ref already handles deduplication, or memoizing the key comparison more defensively.

**2. `totalFeeXlm` is computed in two places** 🔁
Both `FeeBreakdownBottomSheet` and `SendReviewBottomSheet` independently compute `totalFeeXlm` with the same `BigNumber(inclusionFee).plus(resourceFee)` logic. Consider computing and exposing `totalFeeXlm` directly from `useTransactionBuilderStore` (either as a derived getter or a stored field) to keep it DRY and avoid divergence.

**3. `isSoroban` detection is duplicated across layers** 🔁
The "is this a Soroban transaction?" check is done in:
- `transactionBuilder.ts` (`buildTransaction` — checks `contractId` / `isContractId`)
- `TransactionSettingsBottomSheet` (`isCollectibleTransfer || isCustomToken || isSorobanRecipient`)
- `SendReviewBottomSheet` (`isSorobanTransaction(selectedBalance, recipientAddress)`)
- `TransactionAmountScreen` (same helper)

These use slightly different logic. The settings sheet no longer uses the `isSorobanTransaction` helper from `helpers/soroban` and instead inlines its own version — but the inlined version (`isCollectibleTransfer || isCustomToken || isSorobanRecipient`) may not exactly match `isSorobanTransaction()` in all edge cases. Worth verifying these are equivalent or consolidating.

**4. `simulateContractTransfer` return type change is breaking** 📝
`simulateContractTransfer` now returns `{ preparedTransaction, minResourceFee }` instead of a raw XDR string. All internal callers are updated, but if there are any other consumers (e.g. in swap flows or other features), they would break silently. The extension PR avoids this by adding optional fields to an existing interface instead. Worth double-checking there are no other callers.

**5. Fee estimation build skips Blockaid scan silently**
In `TransactionAmountScreen.prepareTransaction`, when `feeEstimationAmount` is provided, the function returns early before the Blockaid scan. This is intentional (comment says "dummy amount"), but it means if the user opens the review without changing the amount, the scan result could be stale or missing. Worth confirming the scan always runs on the "real" build before review.

**6. Cancel button logic change appears unrelated** 📝
```tsx
// Before
tertiary={(isSuspicious && !isUnableToScan) || isExpectedToFail}
destructive={isMalicious && !shouldUseRowLayout}

// After  
tertiary={isSuspicious || isUnableToScan || isExpectedToFail}
destructive={isMalicious}
```
This changes behavior: previously `isUnableToScan` excluded the tertiary style, now it enables it. And `destructive` no longer checks `shouldUseRowLayout`. Was this intentional? It seems unrelated to the fee breakdown feature — might be worth calling out in the PR description or splitting into a separate commit.

**7. Minor: `300ms` setTimeout for auto-simulation**
The debounce in `TransactionAmountScreen` uses a raw `setTimeout` with 300ms. The cleanup function handles `clearTimeout`, so it's safe — just noting it's a manual debounce.

---

### Cross-Platform Comparison: Mobile (#774) vs Extension (#2649)

Both PRs implement the same feature (Soroban fee breakdown: Inclusion Fee + Resource Fee + Total Fee) across the two Freighter platforms. Here's how they differ:

| Aspect | Mobile (#774) | Extension (#2649) |
|--------|---------------|-------------------|
| **State management** | Zustand store (`useTransactionBuilderStore`) with `isSoroban`, `sorobanResourceFeeXlm`, `sorobanInclusionFeeXlm` | Redux + `useRequest` pattern (`simulationState` with `RequestState` enum) |
| **Fee data location** | Stored directly in Zustand store; computed in components via `BigNumber` | `SimulateTxData` interface (shared type) carries `inclusionFee?` and `resourceFee?` |
| **Soroban detection (FeePane)** | Dual system: `isSoroban` from store + `isSorobanContext` prop from parent context | Inferred from `simulationState.data?.resourceFee` presence |
| **Fee breakdown component** | `FeeBreakdownBottomSheet` — NativeWind/Tailwind, uses `@gorhom/bottom-sheet` | `FeesPane` — SCSS-styled, integrated into `MultiPaneSlider` |
| **Auto-simulation trigger** | `useEffect` with `lastAutoSimulatedKey` ref; uses amount `"0"` for early estimation | `useEffect` watching `destination` + `asset` with `simulationDataRef` for dedup |
| **Fee estimation approach** | Simulates with amount `"0"` to get fees before user enters amount; `skipAmountValidation` flag | Simulates with the real amount; auto-triggers when destination changes |
| **Base fee inflation prevention** | Inclusion fee derived from `params.transactionFee \|\| MIN_TRANSACTION_FEE` in the store | `lastInclusionFeeRef` + explicit `saveTransactionFee` reset before re-simulation |
| **Total fee computation** | Computed independently in both `FeeBreakdownBottomSheet` and `SendReviewBottomSheet` (duplicated) | Computed in `simulateTx` → `recommendedFee = baseFee + minResourceFee`; single source of truth |
| **EditSettings fee label** | "Inclusion fee" for Soroban, "Transaction fee" for classic; info button opens FeeBreakdownBottomSheet | "Inclusion Fee" for Soroban, "Transaction Fee" for classic; info button opens FeesPane |
| **Info icon on review screen** | Only shown for Soroban transactions | Always shown (classic and Soroban) |
| **Simulation return type** | **Breaking change**: `simulateContractTransfer` and `simulateCollectibleTransfer` now return `{ preparedTransaction, minResourceFee }` instead of raw XDR string | Unchanged — `useSimulateTxData` hook returns data via existing pattern |
| **Error handling** | New `extractErrorMessage` helper for plain-object API errors | Existing patterns |
| **Test coverage** | Unit tests updated for store + service changes; no E2E tests | **358-line E2E test suite** (6 test cases including re-simulation) |
| **i18n** | EN + PT translations | EN + PT translations (matching keys/descriptions) |
| **Loading state** | `ActivityIndicator` spinner in bottom sheet and fee row | "Calculating..." text in FeesPane and fee display |

#### Key Architectural Differences Worth Aligning

1. **Total fee computation**: The extension computes the total fee in one place (`recommendedFee` in the simulation hook). The mobile computes it in two components independently. Consider centralizing this in the Zustand store — either as a computed getter or a stored `totalFeeXlm` field — to match the extension's DRY approach.

2. **Fee estimation with amount 0**: The mobile introduces `feeEstimationAmount` and `skipAmountValidation` to simulate with amount 0 before the user enters a value. The extension doesn't do this — it only simulates when the destination changes (using the real amount). This gives mobile users earlier fee visibility, but adds complexity (modified `prepareTransaction` signature, `skipAmountValidation` in validation, skipping Blockaid scan). Worth evaluating whether the extension should adopt this too, or whether the added complexity outweighs the benefit.

3. **Soroban detection strategy**: The mobile uses `isSorobanContext` (derived from send context) + `isSoroban` (from store) — two separate signals. The extension's `FeesPane` infers it from `resourceFee` presence, which is simpler but shows classic text during loading. The mobile approach is more correct during loading. Consider aligning on one strategy.

4. **Breaking vs non-breaking simulation API**: The mobile changes `simulateContractTransfer`'s return type from `string` to `{ preparedTransaction, minResourceFee }` (breaking). The extension adds optional fields to `SimulateTxData` (non-breaking additive). The extension approach is safer — if there are any other consumers of `simulateContractTransfer` or `simulateCollectibleTransfer`, they would break silently with the mobile change.

5. **Review screen info icon visibility**: The mobile only shows the fee info icon for Soroban transactions; the extension always shows it. Worth deciding on a consistent approach across platforms — showing it always (extension approach) is arguably better for feature discoverability.

6. **Base fee inflation guard**: The extension's `lastInclusionFeeRef` + explicit Redux dispatch approach is more defensive than the mobile's reliance on `params.transactionFee || MIN_TRANSACTION_FEE`. The extension had to solve this explicitly because Redux stores the total after simulation — the mobile's Zustand store separates inclusion and resource fees, which naturally avoids this. Both work, but for different reasons.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/transactionService.ts
```diff
@@ -718,9 +730,8 @@ export const simulateContractTransfer = async ({
   }
 
   try {
-    // Note: If destination is already muxed (from buildPaymentTransaction),
-    // it will be passed through here. The memo parameter is kept for backward compatibility
-    // but for CAP-0067, memo should be embedded in the muxed address.
+    // Follow the extension pattern: use /simulate-token-transfer which
+    // builds and simulates the transaction on the backend.
     const result = await simulateTokenTransfer({
```
CassioMG, line 737: **Important — displayed inclusion fee may diverge from what's actually submitted.**

`simulateTokenTransfer({...})` is called without forwarding `fee`, even though `SimulateTokenTransferParams.fee?: string` exists in `backend.ts`. The backend rebuilds the tx from `params` with whatever default fee it picks. Meanwhile `transactionBuilder.ts` sets `sorobanInclusionFeeXlm = params.transactionFee || MIN_TRANSACTION_FEE` and the new breakdown UI shows that to the user.

Net effect on the real-amount Soroban path: the inclusion-fee slider in `TransactionSettingsBottomSheet` appears to affect the fee but may not. A user who bumps it 100x will see the breakdown rise without the on-chain fee actually changing.

**Suggested fix:** either thread `fee: xlmToStroop(params.transactionFee).toString()` through this call, or — safer — derive `sorobanInclusionFeeXlm` from the returned `preparedTransaction` XDR (`TransactionBuilder.fromXDR(...).fee` minus `minResourceFee`) so what the user sees matches what the network gets. The latter also covers the early fee-estimation path.

__tests__/hooks/useInitialRecommendedFee.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+    });
+  });
+
+  afterEach(() => {
+    act(() => {
+      useTransactionSettingsStore.getState().resetSettings();
+    });
+  });
+
+  it("does not overwrite a manually changed send fee when the recommended fee changes", () => {
```
CassioMG, line 19: **Important — this test only covers the single-mount case; the bug is in the multi-mount case actually used in the app.**

`useInitialRecommendedFee` keeps `hasManuallyChangedRef` per-hook-instance (a `useRef`). The hook is mounted simultaneously in 4 places (`TransactionSettingsBottomSheet`, `TransactionAmountScreen`, `SendCollectibleReview`, `SwapAmountScreen`), but `markAsManuallyChanged()` is only called from inside `TransactionSettingsBottomSheet`. When `recommendedFee` next changes (e.g., a network-fee-stats refresh), the *other* three hook instances still have `hasManuallyChangedRef.current === false` and will overwrite the user's manual fee back to `recommendedFee`.

This test passes because it renders a single instance. Consider adding a test that mounts the hook from two places, marks one as manually changed, then triggers a `recommendedFee` rerender — and assert the user's value survives.

**Suggested fix to the hook:** hoist the "manually changed" flag into `useTransactionSettingsStore` / `useSwapSettingsStore` (e.g., `feeManuallyChanged: boolean` + `markFeeManuallyChanged()` action) so all instances share the flag.

src/ducks/transactionBuilder.ts
```diff
@@ -295,8 +375,17 @@ export const useTransactionBuilderStore = create<TransactionBuilderState>(
...
+      // show outdated data while the new build is in progress)
+      // Collectibles are always Soroban transactions.
+      set({
+        isBuilding: true,
+        error: null,
+        requestId: newRequestId,
+        isSoroban: true,
+        sorobanResourceFeeXlm: null,
+        sorobanInclusionFeeXlm: null,
+      });
```
CassioMG, line 398: **Important — `buildSwapTransaction` is missing the same Soroban-field reset.**

`buildTransaction` (line ~167) and now `buildSendCollectibleTransaction` (this hunk) reset `isSoroban` / `sorobanResourceFeeXlm` / `sorobanInclusionFeeXlm` at the start of each build cycle. `buildSwapTransaction` only resets `isBuilding` / `error` / `requestId` and silently leaves the Soroban fields stale.

Latent today (no swap screen reads these fields), but it breaks the store invariant "these fields match the transaction currently in `transactionXDR`." Any future shared component that reads `isSoroban` after a swap build will silently misbehave.

**Suggested fix:** add `isSoroban: false, sorobanResourceFeeXlm: null, sorobanInclusionFeeXlm: null` to the `set(...)` call inside `buildSwapTransaction`.

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -622,9 +632,43 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+    const currentKey = `${selectedBalance?.id}|${recipientAddress}`;
+
+    if (
+      !isBuilding &&
+      lastAutoSimulatedKey.current !== currentKey &&
+      isSorobanTransaction(selectedBalance, recipientAddress) &&
+      recipientAddress &&
+      selectedBalance
+    ) {
+      lastAutoSimulatedKey.current = currentKey;
```
CassioMG: **Important — eager mutation of `lastAutoSimulatedKey.current` weakens the debounce.**

The ref is assigned the new `currentKey` *before* the 300ms timer fires. If the user is editing the recipient address and the effect re-runs with a new `currentKey` before the timer fires, the prior timer is cleared (good), but `lastAutoSimulatedKey.current` is already advanced to the latest intermediate key. If the user later types and lands on a key matching that intermediate value (e.g., backspaces), the dedupe check thinks it's already been simulated and silently skips.

**Suggested fix:** move the `lastAutoSimulatedKey.current = currentKey;` assignment *into* the `setTimeout` callback (right before `prepareTransaction(false, "0")`), so it only marks the key as simulated once the simulate actually fires.

src/ducks/transactionBuilder.ts
```diff
@@ -16,6 +22,24 @@ import {
 } from "services/transactionService";
 import { create } from "zustand";
 
+/**
+ * Extracts a human-readable error message from any thrown value.
+ * Handles native Error instances, ApiError plain objects (from apiFactory),
+ * and arbitrary values. Prevents "[object Object]" from reaching the UI.
+ */
+const extractErrorMessage = (error: unknown): string => {
```
CassioMG, line 30: **Minor — duplicated logic, candidate for extraction.**

`extractErrorMessage` is a useful helper but the same `error instanceof Error ? error.message : String(error)` pattern (and the object-with-message variant) already exists in several services across the codebase. Consider promoting this to `src/helpers/errors.ts` so it can be reused (and so the `"[object Object]"` defense applies everywhere). Low priority — fine to defer to a follow-up.

src/components/FeeBreakdownBottomSheet.tsx
```diff
@@ -0,0 +1,160 @@
...
+              <ActivityIndicator
+                size="small"
+                color={themeColors.text.secondary}
+              />
+            ) : (
+              <Text md primary>
+                {hasBuildError
+                  ? "—"
+                  : formatTokenForDisplay(
+                      sorobanInclusionFeeXlm ?? transactionFee,
```
CassioMG, line 106: **Minor — readability.**

`sorobanInclusionFeeXlm ?? transactionFee` falls back to the classic `transactionFee` for the inclusion-fee row when simulation hasn't populated the field yet. Combined with the `hasBuildError` em-dash check above, the reader has to track three states (building / error / pre-sim / post-sim) to follow the rendering ladder. A one-line comment explaining the fallback ("pre-simulation we show the user-selected fee; post-simulation the simulated value") would help maintainability.

__tests__/ducks/transactionBuilder.test.ts
```diff
@@ -134,6 +141,11 @@ describe("transactionBuilder Duck", () => {
     expect(state.error).toBeNull();
     expect(transactionService.buildPaymentTransaction).toHaveBeenCalled();
     expect(transactionService.simulateContractTransfer).toHaveBeenCalled();
+    // Soroban fee fields should be populated from the simulation result
+    expect(state.sorobanResourceFeeXlm).not.toBeNull();
+    expect(typeof state.sorobanResourceFeeXlm).toBe("string");
+    expect(state.sorobanInclusionFeeXlm).not.toBeNull();
+    expect(typeof state.sorobanInclusionFeeXlm).toBe("string");
   });
```
CassioMG, line 175: **Minor — coverage gap on the classic path.**

This Soroban test now asserts `sorobanResourceFeeXlm` / `sorobanInclusionFeeXlm` are populated. The companion classic-transaction test (`it("should build a standard transaction successfully")` around line 93, outside this hunk) doesn't assert `state.isSoroban === false` after build. Worth adding so a future regression where every send is falsely tagged as Soroban gets caught.

CassioMG (commented):
Code review covering correctness, conventions, and tests. Strengths: BigNumber arithmetic throughout, good `requestId` race-guarding in `transactionBuilder`, conventions respected (no `Platform.OS`, no hex literals, absolute imports, en/pt parity). Findings posted inline below.

**Top concern (Important):** the displayed inclusion fee for Soroban contract transfers is decoupled from what the wallet actually submits — the user's slider may have no effect on real-amount sends, but the UI now suggests it does. Worth verifying empirically before merge.

No Critical issues — on-chain correctness is preserved (submitted XDR always comes from the backend's `preparedTransaction`).

CassioMG (approved):

PR #821, opened 2026-04-15
enable sending to federated address

### LINKED ISSUE #807: Enable sending to federated address
### What version are you using?

v1.14.25

### What did you do?

Tried sending to a LOBSTR federated address like `gcewqz*lobstr.co`.

### What did you expect to see?

Success send, like we have on extension.

### What did you see instead?

Error after tapping "Review" on Send Amount screen.

Slack: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1775692549230539

#### Error when trying to send XLM to `gcewqz*lobstr.co`

Image

#### Error when trying to send non-XLM to `gcewqz*lobstr.co`

Image

Image

leofelix077:
Closes #807 

# tested on iOS Device - Build 1.13.25 (1776460485)

----------

https://github.com/user-attachments/assets/ead6bd2e-d74e-4a9f-8051-543558257504

-----

<div>
Screenshot 2026-04-17 at 18 23 57
Screenshot 2026-04-17 at 18 24 00
</div>

<div>
Screenshot 2026-04-17 at 18 24 08
Screenshot 2026-04-17 at 18 24 12
</div>

<div>
Screenshot 2026-04-17 at 18 24 26
Screenshot 2026-04-17 at 18 24 32
</div>

#### PR structure

#### Testing

#### Release


.github/workflows/android-e2e.yml
```diff
@@ -231,46 +231,51 @@ jobs:
       # - SwitchToTestnet: Switch network to testnet
       # - SendClassicTokenMainnet: Send XLM on mainnet (small amount: 0.000001 XLM)
       # - SwapClassicTokenMainnet: Swap tokens on mainnet (small amount: 0.000001 XLM)
+      # - SendFederatedAddress: Send XLM to a federated address on mainnet (small amount: 0.000001 XLM)
```
CassioMG: can we also assert sending `USDC`? maybe we could do both in a single `SendFederatedAddress` flow

.github/workflows/ios-e2e.yml
```diff
@@ -228,46 +228,51 @@ jobs:
       # - SwitchToTestnet: Switch network to testnet
       # - SendClassicTokenMainnet: Send XLM on mainnet (small amount: 0.000001 XLM)
       # - SwapClassicTokenMainnet: Swap tokens on mainnet (small amount: 0.000001 XLM)
+      # - SendFederatedAddress: Send XLM to a federated address on mainnet (small amount: 0.000001 XLM)
```
CassioMG: same

src/ducks/sendRecipient.ts
```diff
@@ -61,23 +63,25 @@ export const useSendRecipientStore = create<SendStore>((set, get) => ({
...
-        : [];
+      const parsedData: (string | { address: string; name?: string })[] =
+        storedAddresses ? JSON.parse(storedAddresses) : [];
 
       // Get current active account public key
       const activePublicKey = await getActiveAccountPublicKey();
 
       // Transform to the Contact format, filtering out the current account
-      const contactList: Contact[] = parsedAddresses
+      // Supports both old format (string[]) and new format ({address, name}[])
```
CassioMG: does this new format goes in line with @piyalbasu's Contact Book feature?

__tests__/components/screens/SendScreen/SendSearchContacts.test.tsx
```diff
@@ -77,6 +77,7 @@ const getSendStoreMock = (overrides = {}) =>
     searchAddress: mockSearchAddress,
     addRecentAddress: mockAddRecentAddress,
     setDestinationAddress: mockSetDestinationAddress,
+    clearSearchState: jest.fn(),
```
CassioMG: **Important - No unit tests for federation resolution logic**

This `clearSearchState: jest.fn()` stub is the only test change in the PR. There are no tests covering the federation resolution path itself in `src/ducks/sendRecipient.ts` (the `searchAddress` branch at lines ~163-194, where `Federation.Server.resolve` is called):
- Happy path (valid `account_id` returned)
- `Federation.Server.resolve` rejection (DNS/HTTP failure, timeout, TLS fail)
- Resolved address equals the active account (`sendToSelfFederation`)
- Malformed `account_id` from the server (non-G or empty string)

For money-moving code that depends on an external third-party resolution step, regression risk is high and a single live E2E flow isn't sufficient coverage. Please add a `sendRecipient.test.ts` duck test (or extend this file) that mocks `Federation.Server.resolve` for each of the cases above.

src/ducks/sendRecipient.ts
```diff
@@ -229,7 +234,8 @@ export const useSendRecipientStore = create<SendStore>((set, get) => ({
 
       const result: Contact = {
         id: `search-${Date.now()}`,
-        address: searchTerm,
+        address: resolvedAddress,
```
CassioMG, line 285: **Important — Resolved `account_id` is not validated before use**

`resolvedAddress` here comes straight from `fedRecord.account_id` assigned at line 172 above. That value is whatever the federation server returned — it isn't checked to be a valid Ed25519 public key. A malformed or malicious response (missing, non-G prefix, wrong length, contract address) will propagate into `getAccount()` / `destinationAddress` and surface later as a confusing Horizon lookup error.

Please validate right after the assignment on line 172:
```ts
if (!StrKey.isValidEd25519PublicKey(fedRecord.account_id)) {
  set({ isSearching: false, searchError: t("sendRecipient.error.federationNotFound") });
  return;
}
resolvedAddress = fedRecord.account_id;
```
(or reuse `isValidStellarAddress`).

CassioMG, line 285: **Important — No explicit timeout on `Federation.Server.resolve`**

The resolve call on line 171 uses SDK defaults, which can hang for a long time on a slow or malicious federation server. The 300ms debounce in `SendSearchContacts` reduces the *number* of calls but does nothing to bound the *duration* of each one.

Please wrap the call with an explicit timeout (e.g. `Promise.race` against a ~10s timer, or the SDK's `timeout` option if exposed) and surface the existing `federationNotFound` / network error when the timer wins. This keeps the UI responsive and prevents a hostile federation server from holding the send flow hostage while the user waits.

e2e/flows/transactions/SendFederatedAddress.yaml
```diff
@@ -0,0 +1,130 @@
+appId: org.stellar.freighterdev
+tags:
+  - transactions
+  - send
+  - federation
+  - mainnet
```
CassioMG, line 6: **Important — E2E flow runs on mainnet against a real third-party federation server**

Tagged `mainnet` with a hard-coded dependency on `lobstr.co`'s federation endpoint (see line 68: `gcewqz*lobstr.co`). This means:
- Every run spends real XLM; even 0.000001 adds up across runs and accounts.
- CI becomes flaky when LOBSTR is down, rate-limits, or changes its federation response shape.
- We take a cross-team dependency outside our control for a core test.

The flow is already tagged `mainnet` — worth confirming it's excluded from the default CI pipeline. Even better: add a testnet variant with either a controlled/owned federation endpoint or a mocked response (via a local server or Maestro's `http.mock`). Reserve the live LOBSTR flow for manual/opt-in pre-release verification.

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -90,35 +94,67 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
+   * @param {string} [contactName] - The federation address if applicable
    */
   const handleContactPress = useCallback(
-    (contactAddress: string) => {
+    (contactAddress: string, contactName?: string) => {
       if (recentAddresses.some((c) => c.address === contactAddress)) {
         analytics.track(AnalyticsEvent.SEND_PAYMENT_RECENT_ADDRESS);
       }
+
+      const isFederation = !!contactName && isFederationAddress(contactName);
```
CassioMG, line 169: **Important — Federation detected by regex on `Contact.name`, not by actual resolution**

`isFederation` here is computed purely by regex-matching the contact's `name` field. Any recent contact whose label happens to match the federation pattern — e.g. a user types `alice*foo.com` as a human label for a plain G... contact — will be treated as a federation address in the UI without ever having been resolved.

Recommend: store an explicit discriminator on `Contact`, e.g. a dedicated `federationAddress?: string` field alongside `address`, or a `type: 'federation' | 'address'` tag. Set it only when resolution actually produced a federation mapping (i.e. in `searchAddress`), and gate this branch on that field instead of the regex. The current regex-on-name approach is fragile and will misclassify user-chosen labels.

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -20,12 +20,13 @@ import {
 import { useQRDataStore } from "ducks/qrData";
 import { useSendRecipientStore } from "ducks/sendRecipient";
 import { useTransactionSettingsStore } from "ducks/transactionSettings";
+import { isFederationAddress } from "helpers/stellar";
```
CassioMG, line 26: **Minor — Federation regex accepts unicode / homoglyphs**

The imported `isFederationAddress` (defined in `src/helpers/stellar.ts:19`) uses:
```ts
/^[^*@]+\*[^*@]+(\.[^*@]+)+$/
```
This accepts arbitrary unicode in both the local part and the domain — including whitespace, control characters, and homoglyph characters (e.g. Cyrillic `о` in `lobstr.co` vs. Latin `o`). A homoglyph domain will parse as a valid federation address and resolve via an attacker-controlled server without the user noticing.

Suggestions in increasing order of strictness:
1. Reject non-printable characters.
2. Normalize via URL/IDN (punycode) rules and show the punycoded form to the user when non-ASCII characters are present.
3. Restrict to ASCII entirely for the MVP and revisit once IDN handling is designed end-to-end.

(Filing here rather than on `stellar.ts:19` directly since that file isn't in this PR's diff and GitHub only allows inline comments on modified files.)

CassioMG:
Have we cross-checked this implementation with extension?

src/ducks/sendRecipient.ts
CassioMG, line 217: **Important — Federation server `memo` / `memo_type` are silently dropped**

Per [SEP-0002](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md), a federation server can return a memo requirement alongside the resolved account. Many exchanges and custodians (LOBSTR, Binance, Coinbase, Kraken, etc.) route funds to pooled Stellar accounts where the memo is the **only** way to identify the individual customer.

### What the server actually returns

```json
{
  "stellar_address": "alice*exchange.com",
  "account_id": "GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ",
  "memo_type": "id",
  "memo": "1234567"
}
```

### What we currently do (bug)

At line 172 above:

```ts
const fedRecord = await Federation.Server.resolve(searchTerm);
resolvedAddress = fedRecord.account_id;   // ← only account_id is read
fedAddress = searchTerm;
```

`fedRecord.memo` and `fedRecord.memo_type` are discarded and never reach `transactionSettings`, so by the time the tx is built, this routing instruction is gone.

### Suggested fix

Carry the memo through to the transaction builder and pre-populate the tx memo (locked from user edits):

```ts
const fedRecord = await Federation.Server.resolve(searchTerm);
resolvedAddress = fedRecord.account_id;
fedAddress = searchTerm;

// SEP-0002: honor server-mandated memo
if (fedRecord.memo_type && fedRecord.memo) {
  set({
    federationMemo: { type: fedRecord.memo_type, value: fedRecord.memo },
  });
}
```

Then in `transactionSettings` / `SendReviewBottomSheet`, when `federationMemo` is present: lock the memo field to the returned value and surface a clear label like _"Required by recipient (federation)"_.

### Next steps

If wiring this end-to-end through the Send pipeline (store → amount → review → submit) is out of scope (or too complex) to fit in this PR, I'd suggest filing a ticket for grooming on the next sprint. This shouldn't block shipping basic federation resolution, but it must be tracked/assessed during backlog grooming.

### Note on Freighter extension

The browser extension appears to have the same bug. All three federation-resolution call sites only read `fedResp.account_id` and discard `memo` / `memo_type`:

- [`extension/src/popup/components/send/SendTo/hooks/useSendToData.tsx`](https://github.com/stellar/freighter/blob/main/extension/src/popup/components/send/SendTo/hooks/useSendToData.tsx) (send flow)
- `extension/src/popup/helpers/account.ts`
- `extension/src/popup/views/SignTransaction/index.tsx`

Worth either fixing there in parallel or filing a GH issue on [stellar/freighter](https://github.com/stellar/freighter/issues) so both clients converge on correct SEP-0002 behavior.

e2e/flows/transactions/SendFederatedAddress.yaml
leofelix077: this is intended. we have other E2E tests running on mainnet. and flakiness is inherent to the e2e tests, unless the endpoints get mocked, which then would defeat the purpose of testing e2e

src/ducks/sendRecipient.ts
leofelix077: it allows the name to be saved as the "title", so federated address on first line, and the G address on subtitle. should work with contact book, but pending testing

leofelix077:
@CassioMG  I opened a separate issue for the extension for the memo type inference bug listed on the analysis https://github.com/stellar/freighter/issues/2717 / https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0002.md

src/services/transactionService.ts
```diff
@@ -382,7 +384,24 @@ export const buildPaymentTransaction = async (
...
+            transactionBuilder.addMemo(Memo.hash(hashBytes));
+          } else {
+            transactionBuilder.addMemo(Memo.text(memo));
+          }
+        } catch {
+          transactionBuilder.addMemo(Memo.text(memo));
+        }
+      } else {
+        transactionBuilder.addMemo(Memo.text(memo));
+      }
```
CassioMG: **Runtime crash**

`Memo.id(memo)` validates the numeric range (0..2^64−1) and will still throw at transaction-build time for, e.g., a 25-digit string of digits even though it passes `/^\d+$/`.

**Crash path**: a malformed federation `memo` with `memo_type:"id"` reaches `Memo.id(memo)` and throws — there's no `try/catch`, unlike the `hash` branch.

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -90,35 +95,93 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
+   * @param {string} [contactName] - The federation address if applicable
    */
   const handleContactPress = useCallback(
-    (contactAddress: string) => {
+    async (contactAddress: string, contactName?: string) => {
       if (recentAddresses.some((c) => c.address === contactAddress)) {
         analytics.track(AnalyticsEvent.SEND_PAYMENT_RECENT_ADDRESS);
       }
+
+      const isFederation = !!contactName;
```
CassioMG: `Contact.name` is a free-form label; using its presence as the federation discriminator will misclassify Contact Book entries**

`isFederation = !!contactName` infers federation-ness from a field that the in-flight Contact Book feature plans to use as a free-form per-contact label (e.g. `"Mom's wallet"`). As soon as a non-federation contact has any `name`:

- The `isFederation && isRecentContact` branch below will call `searchAddress("Mom's wallet")`, which fails `isValidStellarAddress` and surfaces `searchError` — the user sees an error on a previously working contact.
- The `saveFederationAddress(contactName)` call below stores a non-federation label as a federation address in `transactionSettings`, and any federation memo from a prior resolution leaks onto the new send.

**Suggested fix**: add a discriminator on `Contact` (e.g. `type: "federation" | "address"`, or a dedicated `federationAddress?: string`), set only after a successful `Federation.Server.resolve`, and use that here. This will be load-bearing the moment the Contact Book PR lands; safer to ship the discriminator with this PR than to retrofit later.

src/ducks/sendRecipient.ts
```diff
@@ -12,10 +12,16 @@ import { getAccount } from "services/stellar";
...
+export enum ContactType {
+  Federation = "federation",
+  Address = "address",
+}
+
 interface Contact {
   id: string;
   address: string;
   name?: string;
+  type: ContactType;
```
CassioMG: it looks like this new `type` property was added here and is being set in several places in the duck but it is not being consumed anywhere 🤔

leofelix077: Fixed. Removed ContactType enum and type property - verified it was write-only with no business logic consumption. Commit 7918bc30.

src/services/transactionService.ts
```diff
@@ -394,7 +396,29 @@ export const buildPaymentTransaction = async (
...
```
CassioMG: **Silent memo-type downgrade**

When `Memo.id(memo)` validation fails (non-numeric, > 2^64−1), the catch falls through to `Memo.text(memo)` at line 406. Same shape for `memo_type:"hash"` with invalid base64 (lines 414, 417) and `memo_type:"return"` (bare `else` at line 420).

For a federation handle that points to a custodial/exchange address with `memo_type:"id"`, the receiving exchange routes deposits by memo type. A silent downgrade to `text` means the transaction succeeds on-chain but the funds aren't routed to the user's sub-account — they land in the omnibus address, and recovery is manual support / KYC / possibly weeks. I'd rather have the wallet fail-loud than rewrite the memo type.

**Sketch** (not prescriptive): `logger.warn("buildPaymentTransaction", "Federation memo failed validation", { memoType, error })` to capture a Sentry breadcrumb, then surface an `Alert` — something like *"We couldn't process the memo returned by the federation server. The recipient may not credit this transfer correctly. Please verify the address or contact the recipient."* — and abort the send.

**For reference**, the extension's parallel PR https://github.com/stellar/freighter/pull/2744 ships a `federationMemo.ts` helper that validates `id` / `hash` / `text` against SEP-0002, throws on failure, and surfaces a generic user-facing error while Sentry-capturing the diagnostic detail. Not saying we should port it wholesale — just one already-designed pattern in case it fits.

CassioMG (approved):

PR #844, opened 2026-04-27
Fix Android crash on activity restoration

CassioMG:
### What

In `MainActivity.onCreate`, pass `null` to `super.onCreate(...)` instead of forwarding `savedInstanceState`. This prevents Android from attempting to restore `react-native-screens` fragments from the saved state bundle.

Closes https://github.com/orgs/stellar/projects/58/views/2?pane=issue&itemId=179721892

#### The crash

Play Console > Crashes and ANRs has been reporting a top crash for months:

> `org.stellar.freighterwallet.MainActivity.onCreate` — `androidx.fragment.app.Fragment$InstantiationException`

Baseline (1.17.26): **406 events / 118 users**, ~20% of Android crashes. The label is misleading though — Android wraps the real exception. Sentry has the unwrapped trace ([FREIGHTER-MOBILE-4G](https://stellarorg.sentry.io/issues/FREIGHTER-MOBILE-4G)):

> `IllegalStateException: ScreenStack fragments should never be restored. Follow instructions from https://github.com/software-mansion/react-native-screens/issues/17#issuecomment-424704067 to properly configure your main activity.`

`react-native-screens` deliberately throws from `ScreenStackFragment`'s no-arg constructor (see [`ScreenStackFragment.kt:87-91`](https://github.com/software-mansion/react-native-screens/blob/main/android/src/main/java/com/swmansion/rnscreens/ScreenStackFragment.kt#L87-L91)). When Android tears down `MainActivity` (process death, OS reclaim, config change while backgrounded), it serializes the live screen fragments into `savedInstanceState`. On relaunch, `FragmentManager.restoreSaveStateInternal` calls the no-arg constructor → boom, every time. JS-side React Navigation state is what should drive screen reconstruction, not Android's fragment manager.

The fix here is the canonical workaround documented by the `react-native-screens` maintainer in the linked issue: pass `null` to `super.onCreate` so Android never tries to restore the fragments.

#### Why this was nearly invisible in Sentry

Play Console: 406 events / 118 users.
Sentry FREIGHTER-MOBILE-4G: 40 events / 8 users.

Sentry undercounts this category by ~10× because of how `@sentry/react-native` initializes today. We only call `Sentry.init()` from JS (in `src/config/sentryConfig.ts`); there is no native `SentryAndroid.init` in `MainApplication.onCreate`. Per Sentry's docs:

> "the SDK has a current limitation of not capturing native crashes that occur prior to the `init` method being called on the JS layer"

This crash *is* `MainActivity.onCreate` — strictly before the JS bridge boots, before the bundle evaluates, before `Sentry.init()` runs. So the native SDK isn't listening yet. Play Console captures it from the OS; Sentry doesn't.

### Known limitations

- Hard to reproduce locally without forcing process death. Real validation will come from Play Console crash rate on the release after this merges.
- The Sentry issue (FREIGHTER-MOBILE-4G) is intentionally **not** auto-resolved on merge — we'll resolve manually after confirming the Play Console crash rate drops on the next release. A scheduled remote agent will DM the author on May 11 with the verification checklist.

#### PR structure

#### Testing

#### Release

### DIFF
--- a/android/app/src/main/java/com/freightermobile/MainActivity.kt
+++ b/android/app/src/main/java/com/freightermobile/MainActivity.kt
@@ -25,10 +25,13 @@ class MainActivity : ReactActivity() {
       DefaultReactActivityDelegate(this, mainComponentName, fabricEnabled)
 
   /**
-   * react-native-screens package requires this override to properly work on Android devices.
+   * Pass null to super.onCreate to prevent Android from restoring react-native-screens
+   * fragments — they intentionally throw on restoration (process death, config change in
+   * background) and crash the app on relaunch.
+   * https://github.com/software-mansion/react-native-screens/issues/17#issuecomment-424704067
    */
   override fun onCreate(savedInstanceState: Bundle?) {
     RNBootSplash.init(this, R.style.BootTheme)
-    super.onCreate(savedInstanceState)
+    super.onCreate(null)
   }
 }


android/app/src/main/java/com/freightermobile/MainActivity.kt
```diff
@@ -25,10 +25,13 @@ class MainActivity : ReactActivity() {
...
-   * react-native-screens package requires this override to properly work on Android devices.
+   * Pass null to super.onCreate to prevent Android from restoring react-native-screens
+   * fragments — they intentionally throw on restoration (process death, config change in
+   * background) and crash the app on relaunch.
+   * https://github.com/software-mansion/react-native-screens/issues/17#issuecomment-424704067
    */
   override fun onCreate(savedInstanceState: Bundle?) {
     RNBootSplash.init(this, R.style.BootTheme)
-    super.onCreate(savedInstanceState)
+    super.onCreate(null)
```
piyalbasu, line 35: i'm just curious - what does this do now if we're just passing null?

piyalbasu:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

android/app/src/main/java/com/freightermobile/MainActivity.kt
CassioMG: Good question. When Android kills our app to free memory and the user reopens it, Android tries to "restore" the previous screens. Our navigation library (`react-native-screens`) doesn't support that — it deliberately errors out, which is what's crashing the app.

Passing `null` here tells Android: "don't restore anything, start fresh." The app then opens like a normal cold launch and React Navigation rebuilds the screens from our JavaScript state.

Users don't lose anything they care about — all our app state (wallets, balances, settings) lives in JavaScript, not in this Android-level mechanism. The only thing we would technically give up is auto-restoring scroll positions on **native views**, which we don't use anyway on RN.

This is the standard fix recommended by the `react-native-screens` maintainer.

piyalbasu: Ah got it. Thanks for the explanation!

piyalbasu (approved):

PR #856, opened 2026-05-06
Linearize Send Flow

### LINKED ISSUE #809: Linearize Send flow from Home screen
Currently, the Send flow from the Home screen drop the user on the amount screen where they can choose token, amount, and recipient in any order. This can be confusing (e.g., entering an amount before realizing the wrong token is selected).

Proposed flow from Home screen: Send → token selection → amount → recipient → review. Each step gives the user one thing to do at a time.

When starting from the Token Details screen, keep the current behavior (navigate straight to Amount with the token pre-selected).

This should be coordinated with the [equivalent extension change](https://github.com/stellar/freighter/issues/2686) to maintain a consistent cross-platform experience.

Context: https://stellarfoundation.slack.com/archives/C03347FNAHK/p1775602561479619

leofelix077:
Closes #809 

# iOS Build 1.18.26 (1778530225)

This PR linearizes the mobile Send flow so the user moves more cleanly from recipient selection to amount entry to review/processing, with less back-and-forth navigation.
It also tightens recipient handling, keyboard behavior, and state handoff across the send screens so the flow is more reliable
The new coverage now includes scenarios like searching/pasting recipients, choosing from recents or my wallets, and switching between token and collectible send paths.
It also covers amount entry and formatting, review sheet behavior, and navigation/state transitions when moving between recipient, amount, and review screens.

Side adjustments: Added a small change to code-style. as there was a misleading rule previously

the code-style tried to enforce e.g.
`<Button onPress={() => void handleAsyncAction(id)} />`, conflicting with the eslint rule that disables no-floating-promises

handleAsyncAction (no void and no explicit await) is a valid construct in the codebase

-------

General smoke test for android, with some input validations

https://github.com/user-attachments/assets/a578f312-9ce2-4db2-b6c1-1b4ee22f2d5b

-----

Normal send flow, changing inputs, keyboard typing, deleting, and auto focus
Swapping from fiat to token
Input cursor and highlight while typing
Send flow with recents and account names
Send collectible flow 

https://github.com/user-attachments/assets/a973d153-f661-474f-8679-7d3056935df7

-----
Sending to unknown address and recents

https://github.com/user-attachments/assets/509b234e-5a52-43b9-a1af-201f118ae1f7

#### PR structure

#### Testing

- [x] These changes have been tested and confirmed to work as intended on small iOS screens.
- [x] These changes have been tested and confirmed to work as intended on small Android screens.

#### Release


src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -365,6 +400,44 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+      }
+
+      Array.from({ length: currentValue.length }).forEach(() => {
+        handleDisplayAmountChange("");
+      });
+
+      text.split("").forEach((char) => handleDisplayAmountChange(char));
+    },
+    [amountInputValue, handleDisplayAmountChange],
+  );
```
leofelix077, line 498: **[Critical] Multi-character paste/autofill silently drops digits**

`handleNativeAmountChange` fans an N-char `onChangeText` payload into N synchronous calls of `handleDisplayAmountChange(char)`. All of those dispatches share the same render's closure of `currentDisplay`, so the reducer keeps overwriting the same single character — only the **last** key survives.

Trace pasting `"123"` into `"0"`:
1. `("", "0")` → `"0"`
2. `("1", "0")` (stale) → `"1"`
3. `("2", "0")` (stale) → `"2"`
4. `("3", "0")` (stale) → `"3"`

Final `tokenAmount = "3"`. Same for autofill, iOS suggestion bar, and the e2e `inputText: "0.000001"` in `SendClassicTokenMainnet.yaml:102` (would attempt ~1 XLM instead of 0.000001 XLM).

**Fix**: drop the per-char fan-out — either dispatch a single action that folds the entire string in the reducer, or have the reducer read `state.tokenAmountDisplayRaw` instead of the closure-captured payload. Add a regression test that goes through the **real** `useTokenFiatConverter` reducer (the current test mocks it).

— *Found by: Claude (C1) + Superpowers (Crit#1)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -259,14 +287,21 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+    });
   };
 
   const navigateToSelectContactScreen = () => {
-    navigation.navigate(SEND_PAYMENT_ROUTES.SEND_SEARCH_CONTACTS_SCREEN);
+    navigation.push(SEND_PAYMENT_ROUTES.SEND_SEARCH_CONTACTS_SCREEN, {
+      returnToAmount: true,
+      transition: "slide_from_bottom",
+    });
   };
```
leofelix077, line 337: **[Critical] `navigation.push` everywhere can stack duplicate Amount/Token/Contact screens**

`navigateToSelectTokenScreen` and `navigateToSelectContactScreen` both call `navigation.push(...)`, and the return path uses `navigation.goBack()`. Oscillating Amount → Contact → Amount → Token → Amount produces a stack of `[Amount, Contact, Amount, Token, Amount]` — five frames the user thinks are one.

Consequences:
- Each duplicate Amount re-runs the entry effect (`saveRecipientAddress("")`, `saveRecipientName("")`), which can wipe the just-selected recipient.
- Hardware back from outermost Amount lands on Token, not Home.
- Unmount cleanup at lines 183–200 fires per-instance, leaving outer instances' transaction-settings store inconsistent.

**Fix**: use `navigation.navigate(...)` (or guard with `popTo(...)` when the target instance already exists). Honor `transition` via `withTransitionOverride` on the existing screen. Add a test that simulates the oscillation and asserts a stable stack length.

— *Found by: Superpowers (Crit#2)*

src/components/screens/SendScreen/screens/TransactionTokenScreen.tsx
```diff
@@ -16,19 +17,30 @@ type TransactionTokenScreenProps = NativeStackScreenProps<
...
+  const { t } = useAppTranslation();
   const { saveSelectedTokenId, saveSelectedCollectibleDetails } =
     useTransactionSettingsStore();
   const { account } = useGetActiveAccount();
   const { network } = useAuthenticationStore();
   const publicKey = account?.publicKey;
 
+  useLayoutEffect(() => {
+    navigation.setOptions({ title: t("transactionTokenScreen.title") });
+  }, [navigation, t]);
```
leofelix077: **[High] Redundant `navigation.setOptions({ title })` — wrong key for native-stack**

```tsx
useLayoutEffect(() => {
  navigation.setOptions({ title: t("transactionTokenScreen.title") });
}, [navigation, t]);
```

Two issues:
1. `SendPaymentNavigator.tsx:42-49` already configures `headerTitle` for this screen via `withTransitionOverride`. This effect is duplicate work.
2. `@react-navigation/native-stack` reads the header label from `headerTitle`, not `title` (which is only a fallback). Setting `title` here doesn't update `CustomNavigationHeader`. On i18n language flips this effect is the only thing that re-runs, so the header may appear stale.

**Fix**: delete the `useLayoutEffect`. Rely on the navigator's `headerTitle: t(...)` re-rendering when `useAppTranslation` returns a new `t`.

— *Found by: Claude (H1)*

src/config/routes.ts
```diff
@@ -227,15 +227,26 @@ export type AddFundsStackParamList = {
 };
 
 export type SendPaymentStackParamList = {
-  [SEND_PAYMENT_ROUTES.SEND_SEARCH_CONTACTS_SCREEN]: undefined;
+  [SEND_PAYMENT_ROUTES.SEND_SEARCH_CONTACTS_SCREEN]:
+    | {
+        returnToAmount?: boolean;
+        transition?: "slide_from_bottom" | "slide_from_right" | "default";
```
leofelix077: **[High] `transition` is a string union — should be an enum**

```ts
transition?: "slide_from_bottom" | "slide_from_right" | "default";
```

Issues:
- Project rule (`code-style.md` / `freighter-lint-rules`): prefer enums over finite string-union literals.
- This file already exports `ScreenTransition = NativeStackNavigationOptions["animation"]` (line 10) and uses it on `RootStackParamList` — the new `SendPaymentStackParamList` entries duplicate the literals inline.
- `withTransitionOverride` accepts the broader animation type, so the param shape is artificially narrower than the implementation.
- A typo (`"slide_from_buttom"`) would be silently accepted.

**Fix**:
```ts
export enum ScreenTransition {
  SlideFromBottom = "slide_from_bottom",
  SlideFromRight = "slide_from_right",
  Default = "default",
}
```
Reuse in both routes.ts entries (lines 233 and 243) and at the call sites in `TransactionAmountScreen.tsx`.

— *Found by: Claude (H2) + Copilot (#4) + Superpowers (Imp#6)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -882,99 +957,197 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+          spendableBalance.toString(),
+          selectedBalance?.tokenCode,
+        ),
+      })
+    : null;
+
+  // For very long values, stack into two rows: conversion on top, swap+available below.
+  const shouldSplitSecondaryAmounts =
+    !!availableAmountText &&
+    secondaryConversionAmount.length + availableAmountText.length > 34;
```
leofelix077: **[High] Magic number `34` — and it's measured in characters**

```ts
const shouldSplitSecondaryAmounts =
  !!availableAmountText &&
  secondaryConversionAmount.length + availableAmountText.length > 34;
```

`34` has no semantic name and is fragile across locales (PT `"disponível"` vs EN `"available"`, plus large numbers with grouping separators, plus `"USDC"` vs `"yXLM"` token codes — these flip the threshold).

**Fix**: extract `SECONDARY_AMOUNTS_INLINE_CHAR_LIMIT = 34` (in `config/constants.ts`) with a comment explaining what was measured. Better: replace with an `onLayout`-based check so it's pixel-accurate on small screens.

— *Found by: Claude (H3) + Copilot (#5) + Superpowers (Imp#8)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -882,99 +957,197 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+                  style={{
+                    position: "absolute",
+                    top: 0,
+                    right: 0,
+                    bottom: 0,
+                    left: 0,
+                    opacity: 0.01,
+                  }}
+                  caretHidden
+                />
```
leofelix077: **[High] Hidden `TextInput` is inaccessible**

```tsx
<TextInput
  ref={amountInputRef}
  testID="amount-text-input"
  keyboardType="decimal-pad"
  showSoftInputOnFocus
  autoFocus
  value={amountInputValue}
  onChangeText={handleNativeAmountChange}
  style={{ position: "absolute", ..., opacity: 0.01 }}
  caretHidden
/>
```

- No `accessibilityLabel`, no `accessibilityHint`. Screen-reader users land on a near-invisible field with no announced purpose.
- The visible `Display` / `HighlightedAmountDisplay` is decorative — VoiceOver/TalkBack will announce the empty hidden field with the locale's "edit text" default.
- `caretHidden` + `opacity: 0.01` makes selection / cursor controls unusable; double-tap-to-edit gestures don't work for AT users.
- The `TouchableOpacity` wrapper is also unlabeled.

**Fix**: add `accessibilityLabel` / `accessibilityHint` (with EN/PT keys), and set `accessibilityElementsHidden` / `importantForAccessibility="no-hide-descendants"` on the visible `Display` so AT doesn't read it twice. Verify with VoiceOver and TalkBack before merging.

— *Found by: Claude (H4)*

src/components/screens/SendScreen/components/TokensCollectiblesInline.tsx
```diff
@@ -0,0 +1,160 @@
...
+  onCollectiblePress,
+  showSpendableAmount = false,
+  feeContext = TransactionContext.Send,
+  balanceRowTestIDPrefix,
+}) => {
+  const { t } = useAppTranslation();
+  const { themeColors } = useColors();
+  const isLoading = useCollectiblesStore((state) => state.isLoading);
+  const error = useCollectiblesStore((state) => state.error);
+  const { visibleCollectibles } = useFilteredCollectibles();
```
leofelix077, line 50: **[High] New component ships with zero unit tests**

`TokensCollectiblesInline` (160 lines, four rendering branches: loading, error, empty, list-with-collections, plus tap handlers for tokens and collectibles) has no `__tests__/...test.tsx` file. The PR adds tests for `SendSearchContacts` and `TransactionAmountScreen` but skips this new component, which is the entire content of `TransactionTokenScreen` now.

**Fix**: add `__tests__/components/TokensCollectiblesInline.test.tsx` covering all four states and that `onTokenPress` / `onCollectiblePress` forward the right payload.

— *Found by: Claude (H5)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -130,34 +160,53 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
       recentAddresses,
       setDestinationAddress,
       saveRecipientAddress,
+      saveRecipientName,
       navigation,
+      route.params,
       selectedCollectibleDetails,
+      selectedTokenId,
     ],
   );
```
leofelix077, line 326: **[High] Duplicated linear-flow navigation logic**

`handleContactPress` (lines 162–211) and the inline `myWallets.map(...) onPress` (lines 322–351) both:
- set destination, save recipient address+name
- branch on collectibles vs `returnToAmount` vs forward navigation

…with subtle differences (the wallet row doesn't track `AnalyticsEvent.SEND_PAYMENT_RECENT_ADDRESS`, the wallet row doesn't treat the wallet name as a federation address, etc.). Two-and-a-half copies of the same branching that must stay in sync, and they already drift on federation-label handling and `recipientLabel` empty/undefined nuance.

**Fix**: extract a single `proceedAfterRecipient(address: string, name?: string)` helper colocated with the screen, and call it from both. Pass an analytics flag for the different events.

— *Found by: Claude (H6) + Superpowers (Imp#11)*

src/components/TokensCollectiblesInline.tsx
```diff
@@ -0,0 +1,160 @@
...
+              </TouchableOpacity>
+            ))}
+          </View>
+        ))}
+      </View>
+    );
+  }
+
+  return (
+    <ScrollView
```
leofelix077, line 130: **[High] Lost virtualization on inline token/collectible picker**

`ScrollView` + `.map()` for collectibles (and `BalancesList` with `disableInnerScrolling`) eagerly renders all rows. On wallets with many assets/NFTs this increases mount time and memory, and can cause jank. The previous `RecentContactsList` also lost its `FlatList` for the same reason.

**Fix**: use `SectionList` / `FlatList` / `FlashList` for both token and collectible sections, or keep per-tab virtualized lists. If kept as-is, document the bounds (recents cap, etc.).

— *Found by: Copilot (#2) + Claude (L8) + Superpowers (Min#15)*

src/components/layout/ScrollableKeyboardView.tsx
```diff
@@ -14,7 +14,7 @@ const StyledKeyboardAvoidingView = styled(KeyboardAvoidingView).attrs(
   (props: KeyboardAvoidingViewProps) => ({
     behavior: Platform.select({
       ios: "padding",
-      android: undefined,
+      android: "padding",
```
leofelix077, line 17: **[High] Global Android `KeyboardAvoidingView` change is too broad**

Switching Android `behavior` from `undefined` to `"padding"` affects every screen using `ScrollableKeyboardView` (onboarding/import/settings/add token/add collectible/send). Android typically uses `windowSoftInputMode=adjustResize` (manifest) which already moves content up — combining it with `behavior:"padding"` commonly produces double-padding or layout jumps on rotation. Two consecutive PR commits (`6a18093`, `ceaf9354`) iterating on this hint at empirical tuning.

**Fix**: keep the shared default unchanged; opt into Android padding only on screens that need it (e.g. via `keyboardAvoidingViewProps` on `BaseLayout`):
```tsx
behavior: Platform.select({ ios: "padding", android: undefined })
```
QA spot-check on input-heavy screens: sign-in/sign-up forms, recovery phrase, add-token/collectible — any screen with an input near the bottom.

— *Found by: Copilot (#1) + Claude (KAV) + Superpowers (Imp#9)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -130,34 +160,53 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
    * @param {string} contactAddress - The selected contact address
    */
   const handleContactPress = useCallback(
-    (contactAddress: string) => {
+    (contactAddress: string, contactName?: string) => {
       if (recentAddresses.some((c) => c.address === contactAddress)) {
         analytics.track(AnalyticsEvent.SEND_PAYMENT_RECENT_ADDRESS);
       }
+      const federationLabel =
```
leofelix077: **[High] Federation label propagation uses the wrong source field**

Federation detection here is based on `contactName`, but search suggestions pass federation data via `address` and often have no `name`. `federationLabel` is therefore usually dropped, making downstream recipient display metadata inconsistent. Also: if a contact's display "name" happens to look like `user*domain.com`, it's treated as a federation address rather than a name — conflating two concepts.

**Fix**: derive federation label from `contactAddress` first, then fall back to `contactName`:
```tsx
const federationLabel = isFederationAddress(contactAddress)
  ? contactAddress
  : contactName && isFederationAddress(contactName)
    ? contactName
    : "";
```
Better still: track federation explicitly on the contact record rather than re-detecting from `name`.

— *Found by: Copilot (#3) + Superpowers (Min#19)*

src/navigators/SendPaymentNavigator.tsx
```diff
@@ -27,14 +27,26 @@ export const SendPaymentStackNavigator = () => {
...
         component={SendSearchContacts}
-        options={getScreenBottomNavigateOptions(t("sendPaymentScreen.title"))}
+        options={({ route }) =>
+          withTransitionOverride(
+            {
+              headerTitle: t("sendPaymentScreen.title"),
+            },
+            route,
+          )
+        }
```
leofelix077, line 78: **[High] `SendSearchContacts` lost the close-X header button and slide-from-bottom transition**

Previous version used `getScreenBottomNavigateOptions(t("sendPaymentScreen.title"))` which set `headerLeft: () => <CustomHeaderButton icon={Icon.X} />`, the custom `header` component, `animation: "slide_from_bottom"`, and `animationDuration: 300`. The new options object is just `{ headerTitle }` — no `headerLeft`, no animation defaults — so the header falls back to a default back arrow and system slide-from-right.

**Fix**: either restore the X close button (and pick the animation explicitly via `withTransitionOverride`), or — if losing the X is intentional in the linear flow — confirm with design and document in the PR description.

— *Found by: Superpowers (Imp#4) + Claude (L1)*

e2e/flows/transactions/SendClassicTokenMainnet.yaml
```diff
@@ -104,11 +101,16 @@ tags:
 # Confirm the transaction
 - tapOn:
```
leofelix077, line 114: **[High] `inputText: "0.000001"` on mainnet is brittle and (per Crit#1) likely wrong**

Same root cause as the multi-character paste bug — Maestro's `inputText` issues a single text-change with the full string, which goes through `handleNativeAmountChange` and gets reduced to the last digit. Without a fix this would attempt to send roughly **1 XLM instead of 0.000001 XLM** on mainnet. Even if Maestro's current device interaction happens to dispatch keys with re-renders between (masking the bug), the contract is fragile across Maestro versions.

The "retry confirm if sheet still visible" runFlow at lines 112–119 is also a smell: needing a retry usually means the first confirm fired mid-animation. Worth diagnosing rather than papering over.

**Fix**: fix the underlying paste bug, then either keep `inputText` (now safe) or revert to per-key taps using stable testIDs. Investigate and remove the retry-confirm hack once root-caused.

— *Found by: Superpowers (Imp#5) + Claude (L15)*

src/components/screens/SendScreen/components/HighlightedAmountDisplay.tsx
```diff
@@ -100,6 +100,8 @@ interface HighlightedAmountDisplayProps {
   highlightColor: string;
   normalColor: string;
   secondaryColor: string;
+  align?: "left" | "center";
+  size?: "xs" | "lg" | "xl";
```
leofelix077: **[High] `HighlightedAmountDisplay` adds string-union props instead of enums**

New `align?: "left" | "center"` and `size?: "xs" | "lg" | "xl"` props are string unions. Per the project's enum-over-string-union rule (and matching how Display sizes are typed elsewhere in the SDS), these should be a defined enum / shared type.

**Fix**:
```ts
enum AmountAlignment { Left = "left", Center = "center" }
```
…and reuse the existing Display size enum/type instead of a hand-rolled subset.

— *Found by: Superpowers (Imp#7)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -147,13 +153,20 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
       saveRecipientAddress(routeRecipientAddress);
     }
-  }, [routeRecipientAddress, saveRecipientAddress]);
+    saveRecipientName(routeRecipientName ?? "");
+  }, [
+    routeRecipientAddress,
+    routeRecipientName,
+    saveRecipientAddress,
+    saveRecipientName,
+  ]);
```
leofelix077, line 173: **[High] Mount effect clears `recipientAddress`, then a separate effect immediately re-applies it from route params**

Two effects on mount: lines 138–157 clear (depending on `tokenId`), lines 159–169 restore (depending on `routeRecipientAddress/Name`). On initial render this works (clear → restore), but on a subsequent `route.params` update via `setParams`, the first effect doesn't re-run if `tokenId` doesn't change, and the second runs with both old/new transitions, briefly showing the cleared state. Combined with the duplicate-Amount-instance bug (Crit#2), the timing window is non-obvious.

**Fix**: derive recipient from route params **and** stored value in a single effect, and avoid clearing in the entry-effect — let the unmount cleanup at lines 183–200 own the reset. Or treat AmountScreen as fully controlled and read recipient from route params on render with no effect at all.

— *Found by: Superpowers (Imp#10)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -235,18 +287,74 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
+                  </Text>
+                </View>
+                {myWallets.map((wallet) => (
+                  <ContactRow
+                    key={wallet.id}
+                    testID={`my-wallet-row-${wallet.id}`}
+                    address={wallet.publicKey}
+                    name={wallet.name}
+                    onPress={() => {
+                      setDestinationAddress(wallet.publicKey);
```
leofelix077: **[Medium] Wallet row triggers an unnecessary Horizon roundtrip**

`setDestinationAddress(wallet.publicKey)` (in `sendRecipient.ts:258-296`) always fires `getAccount(address, network)` to refresh `isDestinationFunded`. For a wallet from `allAccounts` we already have the funded status implicitly — it's one of the user's own accounts. Wasted RPC per tap, plus a brief `isDestinationFunded: null` flash that the unfunded-notice logic can react to.

**Fix**: skip the funded fetch when the address is in `allAccounts`, or set `isDestinationFunded: true` immediately and let the Horizon call confirm.

— *Found by: Claude (M1)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -130,34 +160,53 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
     [
       recentAddresses,
       setDestinationAddress,
       saveRecipientAddress,
+      saveRecipientName,
       navigation,
+      route.params,
       selectedCollectibleDetails,
+      selectedTokenId,
     ],
```
leofelix077, line 325: **[Medium] `handleContactPress` deps include the entire `route.params`**

```ts
[..., route.params, ...]
```
Only `route.params?.returnToAmount` is read. With React Navigation 7, `route.params` reference can be unstable across renders, recreating the callback on every parent re-render and wasting `useCallback`'s purpose.

**Fix**:
```ts
const returnToAmount = route.params?.returnToAmount;
…
[..., returnToAmount, ...]
```

— *Found by: Claude (M2)*

src/components/screens/SendScreen/screens/TransactionProcessingScreen.tsx
```diff
@@ -103,7 +103,10 @@ const TransactionProcessingScreen: React.FC<
       setStatus(TransactionStatus.FAILED);
     } else if (transactionHash) {
       setStatus(TransactionStatus.SENT);
-      addRecentAddress(recipientAddress);
+      addRecentAddress(
```
leofelix077: **[Medium] Persisting a self-owned wallet name in recents creates duplicate UI**

```ts
addRecentAddress(recipientAddress, recipientName || federationAddress || undefined);
```

If the user just sent to their own "Savings" wallet, that name gets persisted to recents. Next time `SendSearchContacts` opens, the same address is hidden from recents (filtered as `activePublicKey`) — but only while that account is the *active* one. Switching active account makes "Savings" appear in both recents and "My Wallets". Duplication and source-of-truth ambiguity.

**Fix**: in this screen, if `recipientAddress` is in `useAuthenticationStore().allAccounts`, skip `addRecentAddress` entirely. Alternatively dedupe in `loadRecentAddresses` against `allAccounts`.

— *Found by: Claude (M3)*

src/ducks/sendRecipient.ts
```diff
@@ -293,6 +295,16 @@ export const useSendRecipientStore = create<SendStore>((set, get) => ({
...
 
+  prepareForSearch: () => {
+    set({
+      searchResults: [],
+      searchError: null,
+      isValidDestination: false,
+      isDestinationFunded: null,
+      isSearching: true,
+    });
+  },
```
leofelix077, line 360: **[Medium] `prepareForSearch` and `searchAddress` duplicate the same init block**

Both immediately set `isSearching: true`, `searchError: null`, `isValidDestination: false`, `isDestinationFunded: null`, `searchResults: []`. The intent is instant UI feedback during the debounce window, but:
- A `prepareForSearch` followed by `searchAddress` resolution then quick keystroke replays the reset twice.
- Shared init means the next refactor of one will silently diverge from the other.

**Fix**: factor a private `_resetSearchState()` helper inside the duck and call it from both.

— *Found by: Claude (M4)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -239,6 +255,17 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+  const focusAmountInput = useCallback(() => {
+    amountInputRef.current?.focus();
+
+    // iOS can occasionally ignore focus on fully hidden inputs; retry on next tick.
+    setTimeout(() => {
+      if (!amountInputRef.current?.isFocused()) {
+        amountInputRef.current?.focus();
+      }
+    }, 0);
+  }, []);
```
leofelix077, line 281: **[Medium] `setTimeout(..., 0)` for iOS focus retry has no cleanup**

```ts
const focusAmountInput = useCallback(() => {
  amountInputRef.current?.focus();
  setTimeout(() => {
    if (!amountInputRef.current?.isFocused()) amountInputRef.current?.focus();
  }, 0);
}, []);
```

`focusAmountInput` is called from `handleConfirmTransactionSettings`, `handleCancelTransactionSettings`, and `handleCancelReview` — each invocation queues another timer with no cancel handle. If the screen unmounts in the same tick, the retry runs on a null/stale ref.

**Fix**: store the timer in a ref and `clearTimeout` on unmount, or use `requestAnimationFrame` with an existence check. Even simpler: wrap the second focus in `InteractionManager.runAfterInteractions` to wait for the bottom-sheet animation properly.

— *Found by: Claude (M5) + Superpowers (Min#13)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -91,6 +104,14 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
   const shouldShowUnfundedNotice =
     !isCollectibleSend && !isContractTokenSend && !isContractDestination;
 
+  const myWallets: Account[] = (allAccounts ?? []).filter(
+    (acc) => acc.publicKey !== activeAccount?.publicKey,
+  );
```
leofelix077: **[Medium] `myWallets` filter recomputes every render (and `allAccounts ?? []` may be unnecessary)**

```ts
const myWallets: Account[] = (allAccounts ?? []).filter(
  (acc) => acc.publicKey !== activeAccount?.publicKey,
);
```

Runs on every keystroke (each `setAddress` re-renders), and the resulting array is fed into `myWallets.map`. Cheap individually, but it's in the keystroke hot path. Also: is `allAccounts` actually optional? If `useAuthenticationStore` always returns at least `[]`, drop the `??`.

**Fix**:
```ts
const myWallets = useMemo(
  () => (allAccounts ?? []).filter((a) => a.publicKey !== activeAccount?.publicKey),
  [allAccounts, activeAccount?.publicKey],
);
```

— *Found by: Claude (M6 + L4)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -235,18 +287,74 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
             onContactPress={handleContactPress}
           />
         ) : (
-          recentAddresses.length > 0 && (
-            <RecentContactsList
-              transactions={recentAddresses}
-              onContactPress={handleContactPress}
-            />
-          )
+          <ScrollView showsVerticalScrollIndicator={false}>
```
leofelix077: **[Medium] `ScrollView` lacks `keyboardShouldPersistTaps`**

```tsx
<ScrollView showsVerticalScrollIndicator={false}>
```

When the soft keyboard is up (it is — the input has focus), tapping a contact/wallet row first dismisses the keyboard and only fires `onPress` on the second tap. The previous `RecentContactsList` had this issue inside its `KeyboardAvoidingView` + `FlatList`, but the new direct `ScrollView` makes it explicit.

**Fix**: `<ScrollView keyboardShouldPersistTaps="handled">`.

— *Found by: Claude (M7)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -365,6 +400,44 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+        );
+        return;
+      }
+
+      Array.from({ length: currentValue.length }).forEach(() => {
+        handleDisplayAmountChange("");
+      });
+
+      text.split("").forEach((char) => handleDisplayAmountChange(char));
+    },
```
leofelix077, line 496: **[Medium] Pasting non-numeric characters silently no-ops**

`handleNativeAmountChange("12abc34")` → after the loop, only the last digit can win (when Crit#1 is fixed) or nothing useful happens. From the user's perspective: "I pasted something and nothing happened." Common when copying from a fiat app that includes a currency symbol or thousand separators.

**Fix**: while fixing Crit#1, sanitize `text` to digits + locale decimal separator before folding. Optionally show a toast if the paste was non-numeric.

— *Found by: Claude (M9)*

src/components/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -183,8 +183,7 @@ export const HomeScreen: React.FC<HomeScreenProps> = React.memo(
 
     const handleSendPress = useCallback(() => {
       navigation.navigate(ROOT_NAVIGATOR_ROUTES.SEND_PAYMENT_STACK, {
-        screen: SEND_PAYMENT_ROUTES.TRANSACTION_AMOUNT_SCREEN,
-        params: { tokenId: NATIVE_TOKEN_CODE },
+        screen: SEND_PAYMENT_ROUTES.TRANSACTION_TOKEN_SCREEN,
       });
     }, [navigation]);
```
leofelix077, line 188: **[Low] HomeScreen no longer requests slide-from-bottom into the send flow**

```ts
navigation.navigate(ROOT_NAVIGATOR_ROUTES.SEND_PAYMENT_STACK, {
  screen: SEND_PAYMENT_ROUTES.TRANSACTION_TOKEN_SCREEN,
});
```

Without `params.transition`, the SendPayment stack uses whatever default the root navigator gives it. The previous `getScreenBottomNavigateOptions` provided a `slide_from_bottom` default. If the goal is to keep that animation, pass `params: { transition: ScreenTransition.SlideFromBottom }`. Verify on both platforms.

— *Found by: Claude (L2)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -882,99 +957,197 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+                  autoFocus
+                  value={amountInputValue}
+                  onChangeText={handleNativeAmountChange}
+                  style={{
+                    position: "absolute",
+                    top: 0,
+                    right: 0,
+                    bottom: 0,
+                    left: 0,
+                    opacity: 0.01,
```
leofelix077: **[Low] `opacity: 0.01` is an undocumented workaround**

Known iOS workaround to keep the hidden TextInput active while invisible (`opacity: 0` stops events on some RN versions), but undocumented here.

**Fix**: extract `HIDDEN_INPUT_OPACITY = 0.01` as a named constant with a 1-line rationale comment. Same treatment for any other ad-hoc magic numbers in inline styles (e.g. `maxWidth: "55%"` further down).

— *Found by: Claude (L3) + Superpowers (Imp#8)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -91,6 +104,14 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
   const shouldShowUnfundedNotice =
     !isCollectibleSend && !isContractTokenSend && !isContractDestination;
 
+  const myWallets: Account[] = (allAccounts ?? []).filter(
+    (acc) => acc.publicKey !== activeAccount?.publicKey,
+  );
+
+  const debouncedSearchAddress = useDebounce((searchTerm: string) => {
+    searchAddress(searchTerm);
+  }, DEFAULT_DEBOUNCE_DELAY);
```
leofelix077, line 166: **[Low] Floating promise in debounced search**

```ts
const debouncedSearchAddress = useDebounce((searchTerm: string) => {
  searchAddress(searchTerm);
}, DEFAULT_DEBOUNCE_DELAY);
```

`searchAddress` returns `Promise<void>` but isn't awaited. Errors are caught inside the duck so this is functionally fine, but per `code-style.md` the project flags floating promises in review.

**Fix**: `void searchAddress(searchTerm)` (or `await` if you want to chain).

— *Found by: Claude (L5)*

src/components/screens/SendScreen/components/SendReviewBottomSheet.tsx
```diff
@@ -388,9 +390,14 @@ const SendReviewBottomSheet: React.FC<SendReviewBottomSheetProps> = ({
...
-                {slicedAddress}
+              <Text xl medium numberOfLines={1}>
+                {recipientName || slicedAddress}
               </Text>
+              {recipientName && (
+                <Text md medium secondary numberOfLines={1}>
+                  {slicedAddress}
+                </Text>
+              )}
             </View>
```
leofelix077: **[Low] `recipientName` truncated with `numberOfLines={1}`**

Long wallet names like `"Treasury - Quarterly Disbursement Account"` get cropped, hiding part of what the user is reviewing. Acceptable, but consider showing the address on a second line and the name with `numberOfLines={2}`.

— *Found by: Claude (L7)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -882,99 +957,197 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+            </Text>
+
+            <View className="flex-row items-center justify-between">
+              <TouchableOpacity
+                className="flex-1 relative"
+                onPress={focusAmountInput}
+                onPressIn={focusAmountInput}
+                activeOpacity={1}
+                testID="send-amount-focus-trigger"
+              >
```
leofelix077, line 1120: **[Low] `onPressIn` AND `onPress` both call `focusAmountInput`**

```tsx
<TouchableOpacity
  onPress={focusAmountInput}
  onPressIn={focusAmountInput}
  activeOpacity={1}
  …
>
```

Calling both is intentional (so the keyboard pops up before the press-out delay), but it's worth a comment — the duplication looks accidental otherwise. Also: `onPressIn` fires for any touch start including swipes; on small devices a stray scroll-start could bring up the keyboard prematurely.

— *Found by: Claude (L9)*

__tests__/components/screens/SendScreen/SendSearchContacts.test.tsx
```diff
@@ -1,12 +1,20 @@
 import { NavigationContainer, RouteProp } from "@react-navigation/native";
```
leofelix077, line 1: **[Low] No test for the federation-name routing branch**

The new `handleContactPress` branch where `contactName` itself is a federation address (`isFederationAddress(contactName)` true) is novel logic with no test. Easy regression target.

— *Found by: Claude (L11)*

e2e/flows/transactions/SendClassicTokenMainnet.yaml
```diff
@@ -104,11 +101,16 @@ tags:
...
-    visible:
-      id: send-processing-screen
-    timeout: 30000
+
+# Retry confirm once if the sheet is still present
+- runFlow:
+    when:
+      visible:
+        id: send-review-confirm-button
+    commands:
```
leofelix077: **[Low] Retry-once on the confirm button**

```yaml
- runFlow:
    when:
      visible:
        id: send-review-confirm-button
    commands:
      - tapOn:
          id: send-review-confirm-button
```

A retry-once suggests the new flow has an intermittent dismissal issue (likely first confirm fired mid-animation). If known, drop a comment with a Slack/JIRA link. If unknown, file a flake to investigate — see Imp#5 for the same concern.

— *Found by: Claude (L15)*

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -130,34 +160,53 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
+          );
+        }
+      } else if (route.params?.returnToAmount) {
         navigation.goBack();
+      } else {
+        navigation.navigate(SEND_PAYMENT_ROUTES.TRANSACTION_AMOUNT_SCREEN, {
+          tokenId: selectedTokenId || NATIVE_TOKEN_CODE,
+          recipientAddress: contactAddress,
+          recipientName: recipientLabel || undefined,
+        });
```
leofelix077: **[Low] `typeof navigation.popTo === "function"` runtime guard is unusual**

`popTo` is typed on `NativeStackNavigationProp`, so the `typeof === "function"` check is almost always true at runtime. If this is defensive against an older RN-nav version, leave a comment; otherwise drop the branch and trust the type. Same fallback `navigation.navigate(...)` is repeated in the wallet-row branch (lines 327–338) — same simplification.

— *Found by: Superpowers (Min#16)*

src/ducks/sendRecipient.ts
```diff
@@ -293,6 +295,16 @@ export const useSendRecipientStore = create<SendStore>((set, get) => ({
...
   },
 
+  prepareForSearch: () => {
+    set({
+      searchResults: [],
+      searchError: null,
+      isValidDestination: false,
+      isDestinationFunded: null,
+      isSearching: true,
+    });
```
leofelix077: **[High] Recipient names for recents are never persisted** *(comment relocated — original target at line 117 isn't in this PR's diff)*

`addRecentAddress(address, name)` (around line 96-117) stores `name` only on the in-memory `recentAddresses` array. The persistence step maps to `contact.address` only:
```ts
const addressesOnly = updatedAddresses.map((contact) => contact.address);
await dataStorage.setItem(STORAGE_KEYS.RECENT_ADDRESSES, JSON.stringify(addressesOnly));
```
…and `loadRecentAddresses` rehydrates without a name. The new `recipientName` plumbing in `TransactionProcessingScreen` decorates the in-memory copy that survives until the next app launch, then disappears.

The headline UX win — "we remember the name you sent to" — silently degrades after every app restart.

**Fix**: persist as `Array<{ address, name? }>` (versioning the storage key for migration). Update `loadRecentAddresses` to read either shape (`string | { address, name? }`) for backward compatibility.

— *Found by: Superpowers (Imp#3) + Claude (L13)*

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -172,13 +185,15 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+      saveRecipientName("");
       resetSendRecipient();
       resetSettings();
     },
     [
       resetTransaction,
       saveSelectedTokenId,
       saveRecipientAddress,
+      saveRecipientName,
       resetSendRecipient,
```
leofelix077: **[High] Unmount cleanup unconditionally calls `resetSettings()` + `resetSendRecipient()`**

If the user has the AmountScreen unmount mid-flow (deep link to another stack, sheet trick, etc.), `resetSettings()` clobbers user-entered fee/timeout/memo even when the flow may resume. Combined with the duplicate-Amount-instance issue (Crit#2), an inner Amount's unmount can wipe settings the outer Amount still needs.

**Fix**: scope the reset to truly terminal events (transaction submitted, user pressed Close in processing). Use `useFocusEffect` plus a `didSubmit` ref to decide, instead of unmount cleanup.

— *Found by: Superpowers (Imp#12)*

__tests__/components/screens/SendScreen/screens/TransactionAmountScreen.test.tsx
```diff
@@ -1566,3 +1576,196 @@ describe("TransactionAmountScreen - Address Change Scenarios", () => {
...
+
+    fireEvent.changeText(getByTestId("amount-text-input"), "");
+    expect(mockHandleDisplayAmountChange).toHaveBeenCalledWith("");
+  });
+
+  it("calls handleDisplayAmountChange for each pasted character", () => {
+    const { getByTestId } = render(
+      <TransactionAmountScreen navigation={mockNavigation} route={mockRoute} />,
+    );
+
```
leofelix077, line 1904: **[Medium] This test mocks `useTokenFiatConverter` — it cannot catch the paste bug**

```ts
fireEvent.changeText(getByTestId("amount-text-input"), "123");
expect(mockHandleDisplayAmountChange).toHaveBeenCalledTimes(4);
expect(mockHandleDisplayAmountChange).toHaveBeenNthCalledWith(1, "");
expect(mockHandleDisplayAmountChange).toHaveBeenNthCalledWith(2, "1");
…
```

It only asserts the call sequence into the mock — not what the screen actually displays. It misses Crit#1 entirely.

**Fix**: add an unmocked variant that exercises the real reducer (or a thin stub mirroring it) and asserts the resulting visible amount. Seed it with `"0.000001"` so the e2e mainnet value is also covered.

— *Found by: Claude (M8) + Superpowers (Min#17)*

src/components/TokensCollectiblesInline.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const { themeColors } = useColors();
+  const isLoading = useCollectiblesStore((state) => state.isLoading);
+  const error = useCollectiblesStore((state) => state.error);
+  const { visibleCollectibles } = useFilteredCollectibles();
+
+  let collectiblesContent: React.ReactNode;
+
+  if (isLoading) {
+    collectiblesContent = (
+      <View className="items-center justify-center py-6">
```
leofelix077: **[Low] Loading state is asymmetric**

`isLoading` only governs collectibles. `BalancesList` has its own internal loading state — fine — but a future reader sees this `if (isLoading)` branch and may assume balances are also loading.

**Fix**: a one-line comment, or move the spinner above `BalancesList` to reflect the actual loading split.

— *Found by: Claude (L12)*

src/components/screens/SendScreen/screens/TransactionProcessingScreen.tsx
```diff
@@ -238,7 +243,7 @@ const TransactionProcessingScreen: React.FC<
         </View>
```
leofelix077, line 253: **[Low] `recipientName` is loaded but never displayed on the processing screen**

`recipientName` is used here only to seed `addRecentAddress`. The truncated address is shown to the user instead. This is internally consistent, but inconsistent with `SendReviewBottomSheet` which **does** show the name.

**Fix**: consider displaying the name on the processing screen too for visual continuity, or document why the divergence is intentional.

— *Found by: Superpowers (Min#18)*

src/i18n/locales/pt/translations.json
```diff
@@ -745,15 +746,16 @@
   },
   "transactionAmountScreen": {
     "title": "Enviar",
+    "availableBalance": "{{amount}} disponível",
     "percentageButtons": {
       "twentyFive": "25%",
       "fifty": "50%",
       "seventyFive": "75%",
       "max": "Máx"
```
leofelix077, line 755: **[Low] PT phrasing for `availableBalance` (singular vs plural)**

`"availableBalance": "{{amount}} disponível"` works for `XLM` but reads awkwardly for fiat-style amounts in Portuguese (`"3 disponível"`). PT also lacks a plural variant — same issue.

**Fix**: verify wording with a PT-native designer; consider an i18n plural form (`disponível`/`disponíveis`).

— *Found by: Claude (L10)*

src/config/routes.ts
```diff
@@ -227,15 +227,26 @@ export type AddFundsStackParamList = {
...
+    | undefined;
   [SEND_PAYMENT_ROUTES.SEND_COLLECTIBLE_REVIEW]: {
     tokenId: string;
     collectionAddress: string;
   };
-  [SEND_PAYMENT_ROUTES.TRANSACTION_TOKEN_SCREEN]: undefined;
+  [SEND_PAYMENT_ROUTES.TRANSACTION_TOKEN_SCREEN]:
+    | {
+        returnToAmount?: boolean;
+        transition?: "slide_from_bottom" | "slide_from_right" | "default";
```
leofelix077: **[Low] `transition` types in `routes.ts` and `withTransitionOverride` don't agree**

`routes.ts` types `transition` as `"slide_from_bottom" | "slide_from_right" | "default"`. `withTransitionOverride` (in `src/helpers/navigationOptions.tsx:69-77`) accepts the broader `NativeStackNavigationOptions["animation"]` (also includes `"fade"`, `"flip"`, `"none"`, etc.). A future caller passing `transition: "fade"` would type-check at the helper but not in the routes param shape.

**Fix**: align both to the same `ScreenTransition` enum (see the H2 comment on line 233), or widen this param.

— *Found by: Claude (L14)*

src/components/layout/ScrollableKeyboardView.tsx
leofelix077: Tested on Android. fixes the behavior of button not showing on Send Screen. No regression found testing on android device

src/i18n/locales/pt/translations.json
leofelix077: intended

__tests__/components/screens/SendScreen/components/RecentContactsList.test.tsx
```diff
@@ -22,8 +22,8 @@ jest.mock("hooks/useColors", () => ({
   __esModule: true,
   default: () => ({
     themeColors: {
-      foreground: {
-        primary: "#000000",
+      text: {
```
leofelix077, line 25: Two findings from @sdfcharles.

Should be minor changes that will not affect the final code, but will adjust soon and publish another version 

1 to adjust the spacing between the search bar and the recents / first header
and 1 to rework a bit the collectibles to have a horizontal rule and numbering on the right side, to have a more distinct view 

Send
image (1)

leofelix077: https://github.com/user-attachments/assets/61cc5fb9-c1ab-42b5-bf44-303635a76428

__tests__/components/screens/SendScreen/components/RecentContactsList.test.tsx
```diff
@@ -22,8 +22,8 @@ jest.mock("hooks/useColors", () => ({
...
+        secondary: "#000000",
```
leofelix077, line 26: Also coming from discussion with @sdfcharles, to linearize the send collectible

https://github.com/user-attachments/assets/b7d7126b-1f07-4a02-bfe4-d593a1e035df

https://github.com/user-attachments/assets/1fce4d16-767d-4250-be9d-cbbcd1b300b0

__tests__/components/screens/SendScreen/components/RecentContactsList.test.tsx
```diff
@@ -22,8 +22,8 @@ jest.mock("hooks/useColors", () => ({
...
       },
```
leofelix077, line 27: Also adjusting the flows when coming from the details screens

https://github.com/user-attachments/assets/90b296a6-0844-4bff-a6ed-8d18a878d60b

src/components/screens/SendScreen/SendSearchContacts.tsx
```diff
@@ -118,10 +180,67 @@ const SendSearchContacts: React.FC<SendSearchContactsProps> = ({
...
+          SEND_PAYMENT_ROUTES.SEND_COLLECTIBLE_REVIEW,
+          selectedCollectibleDetails,
+        );
+      } else {
+        navigation.navigate(SEND_PAYMENT_ROUTES.TRANSACTION_AMOUNT_SCREEN, {
+          tokenId: selectedTokenId || NATIVE_TOKEN_CODE,
+          recipientAddress: contactAddress,
+          recipientName: name,
+        });
+      }
```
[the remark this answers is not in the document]
leofelix077: intended

CassioMG:
Found a few issues:

1 - Weird list overlap when I use an account with many collectibles

https://github.com/user-attachments/assets/64217092-0970-420c-8744-6e8a1e388049

2 - App is crashing for me after selecting the token (happens with any token apparently, including XLM), I can't proceed from that screen. Same issue happens from the Token Details screen. Reinstalled the app and didn't fix it.

https://github.com/user-attachments/assets/f71fb69c-f897-4952-9001-f6f18c7fb478

Screenshot 2026-05-15 at 15 09 32

leofelix077:
@CassioMG for the long list of tokens / collectibles it should be fixed now 

but for the crash, I never seen it happen (nor @sdfcharles. he also tested on the device). I'm wondering what makes your address be empty / undefined to make it crash. the e2e tests were also passing 

I pushed changes to safeguard the address truncation, but might need your help to take a look again. If it's not something corrupted or stale on your storage (as iOS doesn't clear all data through reinstalls). If you can test it again and let me know. thanks in advance. If it keeps happening would be good to have a dump of your local auth duck (or a slice of it) to check what kind of data and format you are having, to understand the crash better

Screenshot 2026-05-19 at 10 03 53
Screenshot 2026-05-19 at 10 03 58

CassioMG:
@leofelix077 FYI, I think the crash was related to my local env, after doing another fresh install it worked

CassioMG:
@leofelix077 thanks for fixing the bug related to the long list of tokens/collectibles.

I've found a few more issues while testing:

1 - Duplicate token amount lines appearing while editing the dollar amount. This seems the same issue we've [found on extension](https://github.com/stellar/freighter/pull/2764#issuecomment-4464048818) a few days ago. What appears to happen here is that the "token available balance" value which usually appears aligned on the right side (right below the dropdown) is conflicting with the "token amount to send" value which usually appears aligned on the left side. This probably only happens for big string amounts. I'd suggest applying some UI constraints here to prevent this from happening, or keep things flexible, but ensure that:
- the "token available balance" always appear below the dropdown in all cases instead of jumping to the other side
- the "refresh-cw-03" icon always appear right next to the "amount to send" value instead of breaking the line which results in the icon being placed on the left side of the "token available balance" 
double-amount-lines-1
double-amount-lines-2

2 - It seems the keyboard doesn't always get dismissed when you tap on "Review send". This results in a bug where the user can't see or scroll the full content of the review bottom-sheet. It appears this bug is triggered when you edit something on transaction settings and then tap on "Review send" (first video). Another thing I've noticed is that once this bug is "activated" it always happens, for all tokens, not mattering if you are starting the send flow from the Home Send button or the Token Details screen (quick AQUA sample on the second video). It seems you have to switch tabs to History or Discover screen and only then go back to Home tab to "clear out" this bug otherwise you'll always see this issue on the "Review send" step.

https://github.com/user-attachments/assets/30bf873b-0bb0-4489-89bd-b433389c8286

https://github.com/user-attachments/assets/a4cc80b2-b8b7-49e7-9e16-750c23630cf2

3 - All screens with an "X" button should slide from the bottom. E.g. in the second 8 on the video below it navigates to a "X" screen which slides from the right instead of sliding from the bottom. Then when you tap the "X" button it slides down as expected so it appears we only need to fix the navigation when entering on the screen. This happens on both Send token and Send collectible flows.

https://github.com/user-attachments/assets/f3492bc8-f2f4-4576-99a0-9604c6d3e595

4 - Some UI tweaks when comparing [with Figma](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=11315-128443&t=MWDHIxXL0osTytjl-1):
 - The bottom margin of the Headers on the addresses lists (e.g. "Suggestions", "Recents", My Wallets") seems a little tight
 - The bottom of the address list is being cut (probably safe-area related)
 - The title of the Send amount screen should be "Send" instead of "Sending". Same title copy fix needs to be applied for Collectibles.

send-screen-title
bottom-is-cut-and-tight-margins

5 - When you start a "Send Collectible" flow and use the QR Scan to input the address you land on the Send token amount screen with XLM selected instead of the Send collectible review screen

https://github.com/user-attachments/assets/35863e61-eb4e-4aad-a5a7-f2320d84efa4

docs/best-practices/code-style.md
```diff
@@ -62,6 +62,29 @@ handler. See `anti-patterns.md` for examples. Note:
...
+// Correct — block body returns undefined; .catch() reports failures
+<Button
+  onPress={() => {
+    handleAsyncAction(id).catch((error) => {
+      logger.warn("Component", "Action failed", error);
+    });
+  }}
+/>
+```
+
```
CassioMG, line 87: could you please mention on the PR description that this best-practice file is being updated? Thanks

src/components/layout/ScrollableKeyboardView.tsx
CassioMG, line 17: this will change Android keyboard behavior in all screens with inputs, have we double checked to see if it's working consistently everywhere including the Discover tab? thanks

src/i18n/locales/en/translations.json
```diff
@@ -780,16 +781,17 @@
     }
   },
   "transactionAmountScreen": {
-    "title": "Send",
+    "title": "Sending",
+    "availableBalance": "{{amount}}",
```
CassioMG: it seems like we are not translating any copy here, it's just the amount

src/i18n/locales/pt/translations.json
```diff
@@ -744,16 +745,17 @@
     }
   },
   "transactionAmountScreen": {
-    "title": "Enviar",
+    "title": "Enviando",
+    "availableBalance": "{{amount}}",
```
CassioMG: same here

src/components/layout/ScrollableKeyboardView.tsx
CassioMG: just saw your [old comment here](https://github.com/stellar/freighter-mobile/pull/856/changes#r3219693158), seems like you have covered it already

docs/best-practices/code-style.md
```diff
@@ -62,6 +62,29 @@ handler. See `anti-patterns.md` for examples. Note:
 (`eslint.config.mjs`) — this is a reviewer-enforced convention, not a
 linter-enforced rule.
 
+**Async event handlers**: When a component prop expects a `void`-returning
```
leofelix077, line 65: Added this small change to code-style. as there was a misleading rule previously

the code-style tried to enforce 

`<Button onPress={() => void handleAsyncAction(id)} />`, conflicting with the eslint rule that disables no-floating-promises

handleAsyncAction (no void and no explicit await) is a valid construct in the codebase

leofelix077:
@CassioMG Applied the small UI tweaks to match closer the figma designs latest version, the keyboard dismiss, and fixing the navigation when scanning QR code on collectibles

regarding the  first point, I'm not sure what you mean. I also left a [comment](https://github.com/stellar/freighter/pull/2764#issuecomment-4548234384) on the extension PR regarding it 




IMG_5446
IMG_5447
IMG_5448

https://github.com/user-attachments/assets/c385a33c-9dc7-46f9-b20d-e2633688e2e0


https://github.com/user-attachments/assets/87071167-30a2-4521-9195-42ce5901ce2a


https://github.com/user-attachments/assets/2c6c23bb-86f7-464e-ab63-b212f8db239d

leofelix077:
UI Has also been adjusted to have the total balance above the token dropdown. Thread is centralized on the extension thread https://github.com/stellar/freighter/pull/2764#issuecomment-4566444093

e.g. 

Simulator Screenshot - iPhone 17 Pro - 2026-05-28 at 13 22 28

Simulator Screenshot - iPhone 17 Pro - 2026-05-28 at 13 22 48

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -137,31 +144,36 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+    // as a wallet/contact nickname.
+    if (routeRecipientName && isFederationAddress(routeRecipientName)) {
+      saveFederationAddress(routeRecipientName);
+      saveRecipientName("");
     } else {
-      saveSelectedTokenId("");
+      saveFederationAddress("");
+      saveRecipientName(routeRecipientName ?? "");
     }
+    saveMemoType("");
```
CassioMG: **Critical fund-loss bug on federation sends — `Memo.id` silently downgraded to `Memo.text`.**

`SendSearchContacts.handleContactPress` writes the federation memo + memo_type into `transactionSettings` immediately before `navigation.navigate(TRANSACTION_AMOUNT_SCREEN, ...)`. This mount effect's final line (`saveMemoType("")`) then unconditionally wipes `transactionMemoType` while leaving `transactionMemo` intact — putting the store in an inconsistent state.

By the time `buildTransaction` runs, the build site reads from `useTransactionSettingsStore.getState()` and the store has e.g. `transactionMemo: "12345"`, `transactionMemoType: ""`. In [`transactionService.ts:407-431`](https://github.com/stellar/freighter-mobile/blob/efafeca9f1132b6b4f534783b926671ae508e47a/src/services/transactionService.ts#L407-L431), an empty `memoType` falls through to `else { transactionBuilder.addMemo(Memo.text(memo)) }`. A federation address that returned `memo_type: "id"` (the standard for exchange deposit memos — Kraken, Binance, Coinbase, Bitfinex, etc.) is silently downgraded to `Memo.text`. Funds land at the exchange's omnibus address but are **not credited to the user's sub-account.**

**Affected entry paths** — every primary route to AmountScreen with a federation recipient:
- HomeScreen → TokenScreen → SearchContacts → AmountScreen (forward `navigate`)
- TokenDetailsScreen → SearchContacts → AmountScreen (`dismissToPreviousScreen: true` is set but `transition` is not, so the overlay branch in [SendSearchContacts.tsx#L225-L227](https://github.com/stellar/freighter-mobile/blob/efafeca9f1132b6b4f534783b926671ae508e47a/src/components/screens/SendScreen/SendSearchContacts.tsx#L225-L227) is false → forward-navigates → fresh mount)
- QR scanner → AmountScreen (separately also drops `memoType` upstream)

**Invisible to the user**: [`SendReviewBottomSheet.tsx#L303-L307`](https://github.com/stellar/freighter-mobile/blob/efafeca9f1132b6b4f534783b926671ae508e47a/src/components/screens/SendScreen/components/SendReviewBottomSheet.tsx#L303-L307) shows the memo value but not its type, so the user sees `"12345"` and confirms.

**Not caught by e2e**: `SendFederatedAddress.yaml` uses LOBSTR's federation server, which returns no memo on resolve.

**Suggested minimal fix** — remove this line. `handleContactPress` already sets memo + memoType atomically for both federation and non-federation contacts. Stale state from a previous incomplete send is already handled by `handleCloseSendFlow` / `handleProcessingScreenClose` / navigator-level `closeSendFlow`.

```suggestion
```

After applying the suggestion, also remove `saveMemoType` from the destructure on line 134 and from the deps array on line 174 to keep ESLint happy.

**Please add a regression test**: pre-populate `useTransactionSettingsStore` with `transactionMemo: "12345"`, `transactionMemoType: "id"`, mount `TransactionAmountScreen` with `route.params = { tokenId, recipientAddress, recipientName: "alice*kraken.com" }`, then assert `useTransactionSettingsStore.getState().transactionMemoType === "id"` after mount, and that `buildTransaction` is called with `transactionMemoType: "id"`. A second test with `memoType: "hash"` is also worth adding.

CassioMG:
@leofelix077 thanks for making the last UI changes/fixes. It appears there is [bug related to the Send (federation)](https://github.com/stellar/freighter-mobile/pull/856#discussion_r3326496502) flow where the memo type is always being reset on the SendAmount rendering. The fix seems simple but I left the AI verbose description there which I think could be helpful for the agent to get the whole context/severity of it. Could you also please double check to see if there isn't any other part broken related to the Send-federation flow? Thanks

src/components/screens/SendScreen/components/TokensCollectiblesInline.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }: {
+    collectionAddress: string;
+    tokenId: string;
+  }) => void;
+  showSpendableAmount?: boolean;
+  feeContext?: TransactionContext;
+  balanceRowTestIDPrefix?: string;
+}
+
+export const TokensCollectiblesInline: React.FC<
```
CassioMG, line 35: is this a shared component? If not, we could probably move to the folder where it's being used (e.g. SendScreen/components)

src/components/screens/SendScreen/components/HighlightedAmountDisplay.tsx
CassioMG: did this HighlightedAmountDisplay become a dead-component?

src/components/screens/SendScreen/components/TokensCollectiblesInline.tsx
leofelix077: currently not reused. moved it to SendScreen 👌

src/components/screens/SendScreen/components/HighlightedAmountDisplay.tsx
leofelix077: yep. also removed it now. it was only used there for the fiat / token input highlighting

CassioMG:
@leofelix077 I'm still seeing the bug in which the keyboard doesn't get dismissed after saving a memo on settings:

https://github.com/user-attachments/assets/e638a7e2-4eaf-439a-b670-a6579ae52312

leofelix077:
@CassioMG the issue is fixed now on 1.18.26 (1780495145). can you verify? I also added a delay to open the settings screen to allow for the keyboard dismiss from number input -> settings modal

https://github.com/user-attachments/assets/9a615902-2926-4eff-b821-52658d9b6c50

CassioMG:
@leofelix077 about the [wait for keyboard dismiss](https://github.com/stellar/freighter-mobile/pull/856/changes/243b04330b84c58407b645b6d4501c210e0c2c25) strategy, would you be willing to try [this approach instead](https://github.com/stellar/freighter-mobile/pull/879/changes/eb64f7d7690f6d2e08e6ae9e86ace7163d11c7e9#diff-ff46dc62340a3bbf2a9854f2e89ae1bc842939ff79b875c66346e103d3e6aa1dR110)? I think we could extract that function to a helper and try here on your PR, if that works then we don't need to rely on timeouts and will also have less duplicate code. Wdyt? If so, please check on both Android and iOS as I've only tested on iOS so far on my PR. Thanks

leofelix077:
@CassioMG much better approach. I forgot about the native events for keyboard. Will try it out and update it in a bit

leofelix077:
@CassioMG adjusted and retested with the native event handler on build 1.18.26 (1780503276)

https://github.com/user-attachments/assets/095b8bd4-5d43-40dc-bdd1-b139c049c3f8

CassioMG (approved):

PR #872, opened 2026-05-15
[WIP] Swap to New Token

CassioMG:
Design doc reviewed ✅ 

Implementation in progress...


docs/swap-to-new-token-design.md
```diff
@@ -0,0 +1,711 @@
...
+  returns AQUA).
+- **>1 results** — the `C…` substring incidentally appears in TOML metadata
+  (e.g. listed as a related contract or in a free-form description) of one or
+  more Classic assets, so stellar.expert returns each of those.
+
+Our handling is uniform across all three: we apply the same classic-only filter
+via [`helpers/balances.ts:getTokenType`](../src/helpers/balances.ts) that we use
+for any other search term. Classic records that come back pass through; Soroban
+contract tokens are filtered out. So a SAC paste typically resolves to its
+wrapped classic (and is swappable); a pure Soroban paste yields nothing; an
```
JakeUrban, line 172: Ideally someone who pastes a contract token address gets a message explaining that contract tokens are not yet supported.

docs/swap-to-new-token-design.md
```diff
@@ -0,0 +1,711 @@
...
+  (passes through `search`, `sort`, `order`, `limit`, `cursor`).
+- Keep returning the same `SearchTokenResponse` shape (already updated on this
+  `cg-swap-to-new-token` branch), so the frontend easily migrates by swapping
+  the service base URL.
+
+Until the proxy lands we can call stellar.expert directly via the existing
+[`services/stellarExpert.ts`](../src/services/stellarExpert.ts) — adding one new
+method that routes per-network the same way `searchToken` already does (via
+`getApiStellarExpertUrl(network)`), so testnet users get testnet trending data
+and mainnet users get mainnet:
```
JakeUrban, line 228: I forgot that we still call stellar expert directly. We should move all requests to third-party APIs to our backend so we have a single API for all data.

leofelix077 (approved):
don't have anything to add directly. At least from reading it and the diagrams it makes sense to me 

only I think general questions but more related to the implemenation on how it's gonna look like with the integration to the new flows and the different modals being added, navigation-wise and displaying of clear info to the user

e.g. when they add a trustline then their max spendable balance goes down, if we can/should show a toast to the user "Trustline added. 0.5 XLM will be reserved from total". or a more detail message inside the "This will add a trustline" banner

CassioMG:
> don't have anything to add directly. At least from reading it and the diagrams it makes sense to me
> 
> only I think general questions but more related to the implemenation on how it's gonna look like with the integration to the new flows and the different modals being added, navigation-wise and displaying of clear info to the user
> 
> e.g. when they add a trustline then their max spendable balance goes down, if we can/should show a toast to the user "Trustline added. 0.5 XLM will be reserved from total". or a more detail message inside the "This will add a trustline" banner

@leofelix077 thanks for reviewing. Good suggestion on making the 0.5 XLM reserve amount more clear on the UI, I've just synced with @minkyeongshin and we are [updating the banner](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=11697-19791&t=njLgo6jJJUom2ckJ-1) to expose it

docs/swap-to-new-token-design.md
```diff
@@ -0,0 +1,718 @@
...
+  returns AQUA).
+- **>1 results** — the `C…` substring incidentally appears in TOML metadata
+  (e.g. listed as a related contract or in a free-form description) of one or
+  more Classic assets, so stellar.expert returns each of those.
+
+Our handling is uniform across all three: we apply the same classic-only filter
+via [`helpers/balances.ts:getTokenType`](../src/helpers/balances.ts) that we use
+for any other search term. Classic records that come back pass through; Soroban
+contract tokens are filtered out. So a SAC paste typically resolves to its
+wrapped classic (and is swappable); a pure Soroban paste yields nothing; an
```
piyalbasu, line 172: > a pure Soroban paste yields nothing

Just to clarify - we just mean a C address that is not a SAC, right?

JakeUrban (commented):
Adding comments from an earlier review now. I'll find some time today to review the rest

docs/swap-to-new-token-design.md
CassioMG: correct

docs/swap-to-new-token-design.md
CassioMG: Good call. It's better informing the user than displaying zero results. I think we can display a short message like "Soroban contract tokens are not supported yet." in place of the [results list](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=11310-101382&t=PXV1pVIF9tH9elUN-1) after the request completes. Adding @minkyeongshin to the loop in case you have a copy/UI suggestion for this case. Thanks!

docs/swap-to-new-token-design.md
CassioMG: Yep, once we create the endpoint I was planning on replacing all stellar.expert's frontend direct calls with our endpoint instead

JakeUrban (approved):

docs/swap-to-new-token-design.md
```diff
@@ -0,0 +1,718 @@
...
+  Swap PR (not split out), so the screen ships consistent. The
+  `setDisplayAmountFromText` converter extension itself comes from whichever PR
+  lands first — this one or
+  [#856](https://github.com/stellar/freighter-mobile/pull/856) — per §6.2.
+- Coordinate the backend proxy GH issue as a separate stream; the frontend flow
+  can be initially merged using stellar.expert directly.
+
+## 14. Open questions
+
+1. **Trending price fallback** — when `/token-prices` lacks a token, do we show
```
piyalbasu: We should also consider a fallback for if stellar.expert is down. A user would not be able to see trending/popular tokens or be able to swap to a new asset (likely with a UI indication that the service is down), but they should still be able to swap to an asset they already have a trustline for (the existing flow)

piyalbasu (approved):
This is looking good to me, aside from the 1 comment I left.

If you wanted to, you could add this doc to the wallet-eng-monorepo, and expand this doc to include implementation details for extension (I imagine there'd be a lot of overlap between the 2). The benefit would be that there would be 1 doc that serves as the source of truth for both extension and mobile, helping keep implementation in sync between the 2

docs/swap-to-new-token-design.md
CassioMG: Good call, I'll make this fallback explicit on the doc.

CassioMG:
> This is looking good to me, aside from the 1 comment I left.
> 
> If you wanted to, you could add this doc to the wallet-eng-monorepo, and expand this doc to include implementation details for extension (I imagine there'd be a lot of overlap between the 2). The benefit would be that there would be 1 doc that serves as the source of truth for both extension and mobile, helping keep implementation in sync between the 2

@piyalbasu nice call, I'll add it to the monorepo and use the same doc to expand on the extension version once I start working on it - thanks!

docs/swap-to-new-token-design.md
CassioMG: I will be actually using this updated copy here for now: "Swapping with Soroban contract tokens is not supported yet." to be more specific about swapping, since users can still add and send custom tokens using Freighter

CassioMG: Synced with @minkyeongshin about it, we'll be using this copy "Soroban contract tokens aren't supported for swaps yet. Try searching for a Classic token instead." for cases where the search result return only Soroban contract tokens. This is the Figma design: https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=11919-40587&t=v7YhLC3pzUp7bR4f-1

PR #892, opened 2026-06-05
Support scanning Stellar address QR codes from account view

### LINKED ISSUE #761: Support scanning both WC and Stellar address QR codes

JakeUrban:
# Proposal: Support scanning Stellar address QR codes from account view

## Summary

Make the home screen QR scanner support scanning Stellar addresses (G, C, M, and federation addresses) in addition to WalletConnect URIs. When a Stellar address is scanned, route the user to the send flow with XLM pre-selected as the default token. Refactor the QR scanner architecture to centralize payload parsing and migrate all existing scanner hooks to the shared parser.

Also changes the wallet connect manual input component to only be visible when `isDev` is true.

Also removes the ability to view your address QR code from the QR code scanner. This was already something @minkyeongshin planned for, and it simplifies the implementation here.

## Design Reference

Figma: https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=12165-107592

## Structural Refactor: Centralized QR Payload Parser

**`parseQRPayload(data: string, currentPublicKey?: string): QRPayload`** — Pure synchronous function in `helpers/qrValidation.ts`:

```ts
type QRPayload =
  | { type: "stellar_address"; address: string }
  | { type: "wallet_connect"; uri: string }
  | { type: "invalid"; error: QRCodeError }
```

Priority: WC URI (unambiguous `wc:` prefix) → Stellar address → invalid. Self-send check (exact string match) included. Async validation (federation, muxed equivalence, destination status) handled by `searchAddress()`.

All three scanner hooks migrate to `parseQRPayload`.

### 3. Migrate existing hooks to `parseQRPayload`

**`useWalletConnectQrCodeScanner`**: Replace `isValidWalletConnectURI()` with `parseQRPayload()`. On `"wallet_connect"` → pair. On other → invalid WC URI toast.

**`useSendFlowQrCodeScanner`**: Replace `validateQRCodeWalletAddress()` with `parseQRPayload()`. On `"stellar_address"` → existing `searchAddress()` flow. On other → invalid address toast. Add useEffect for searchError: when `isProcessingQRScan && searchError !== null && !isSearching`, show error toast and reset processing flag.

### 4. Create `useHomeQrCodeScanner` hook

New hook at `src/hooks/useHomeQrCodeScanner.tsx`:

**Effect safety**: All useEffects are gated on an `isProcessingQRScan` flag that is only set to `true` when this hook's `handleQRCodeScanned` initiates a scan. This prevents stale `sendRecipientStore` state from triggering effects on mount or when other hooks use the same store. On mount, reset relevant store state via `resetSearch()` (or check the flag before acting).

**On scan**, calls `parseQRPayload(data, account.publicKey)`:
- `"stellar_address"` → sets `isProcessingQRScan = true`, calls `searchAddress(payload.address)`. useEffect:
  - **Success** (`isProcessingQRScan && isValidDestination && !isSearching && searchResults.length > 0`): saves recipient, `navigation.popToTop()` then navigates to `SEND_PAYMENT_STACK > TRANSACTION_AMOUNT_SCREEN` with `tokenId: NATIVE_TOKEN_CODE` and `recipientAddress: destinationAddress`. Tracks `analytics.trackQRScanSuccess(QRCodeSource.HOME_SCANNER)`. Resets `isProcessingQRScan`.
  - **Error** (`isProcessingQRScan && searchError !== null && !isSearching`): shows toast with `searchError`, resets `isProcessingQRScan`. Tracks `analytics.trackQRScanError(QRCodeSource.HOME_SCANNER, "address_validation_failed")`.
- `"wallet_connect"` → calls `walletKit.pair({ uri })`. Success: track success, `navigation.popToTop()`. Error: toast, track error.
- `"invalid"` → error toast with dedup (`lastErrorCode`), track error.

**Navigation on success**: Uses `navigation.popToTop()` before `navigation.navigate(SEND_PAYMENT_STACK, ...)` — same pattern as `useSendFlowQrCodeScanner`. This ensures back-navigation from the send screen returns to the home screen, not the camera.

**Presentation:**
- Title: `t("homeScanner.title")` = "Send funds"
- Scanner title: `t("homeScanner.scanQRCodeText")` = "Scan a QR code to send"
- `showManualInput: false`, `showHeaderRight: true`
- Header right → `navigation.replace(ACCOUNT_QR_CODE_SCREEN, { showTabs: true, showNavigationAsCloseButton: true, transition: "fade" })`
- `handleClose`: `navigation.popToTop()`

### 5. Update `useQRCodeScreenScanner` — enabled gates

Each hook receives an `enabled` boolean parameter (default `false`). When `enabled` is false, all useEffects early-return. Only the hook matching the active `source` is passed `enabled: true`. This prevents inactive hooks from triggering side effects when `useQRCodeScreenScanner` calls all three unconditionally.

```ts
const walletConnectResult = useWalletConnectQrCodeScanner(source === QRCodeSource.WALLET_CONNECT);
const sendFlowResult = useSendFlowQrCodeScanner(source === QRCodeSource.ADDRESS_INPUT);
const homeResult = useHomeQrCodeScanner(source === QRCodeSource.HOME_SCANNER);
```

### 6. Update home screen scan button

In `useHomeHeaders.tsx`, change to `source: QRCodeSource.HOME_SCANNER`.

### 7. Shared route param type

```ts
export type AccountQRCodeScreenParams = {
  showNavigationAsCloseButton?: boolean;
  showTabs?: boolean;
  transition?: ScreenTransition;
};
```

Referenced from both `RootStackParamList` and `AddFundsStackParamList`.

### 8. Tabs in AccountQRCodeScreen

When `showTabs` is true: set `headerShown: false` via `navigation.setOptions()` in useLayoutEffect, render custom header bar in component body.

Custom header: X (close → popToTop) left, "Receive"/"Scan" segmented control center. "Scan" → `navigation.replace(SCAN_QR_CODE_SCREEN, { source: HOME_SCANNER, transition: "fade" })`.

`useRightHeaderButton` called unconditionally but headerShown: false prevents rendering.

### 9. Fix QR error analytics argument order

`src/components/QRScanner.tsx` line 136: `trackQRScanError(error, context)` → `trackQRScanError(context, error)`.

### 10. i18n strings

English: `homeScanner.title` = "Send funds", `homeScanner.scanQRCodeText` = "Scan a QR code to send", `accountQRCodeScreen.tabs.receive` = "Receive", `accountQRCodeScreen.tabs.scan` = "Scan"

Portuguese: `homeScanner.title` = "Enviar fundos", `homeScanner.scanQRCodeText` = "Escaneie um código QR para enviar", `accountQRCodeScreen.tabs.receive` = "Receber", `accountQRCodeScreen.tabs.scan` = "Escanear"

### 11. Tests

- `parseQRPayload.test.ts` — All payload types, priority, self-send
- `useHomeQrCodeScanner.test.tsx` — Address success → popToTop + navigate with XLM; WC success; error paths; dedup; analytics; enabled=false no effects
- `useSendFlowQrCodeScanner.test.tsx` — Add error handling tests; enabled gate
- `useWalletConnectQrCodeScanner.test.tsx` — Migration verification; enabled gate
- AccountQRCodeScreen tabs render + navigation
- useHomeHeaders source change

## Files to Change

1. `src/config/constants.ts` — Add `HOME_SCANNER`
2. `src/config/routes.ts` — `AccountQRCodeScreenParams`, `showTabs`
3. `src/helpers/qrValidation.ts` — `parseQRPayload()` + `QRPayload` type
4. `src/hooks/useHomeQrCodeScanner.tsx` — New hook
5. `src/hooks/useQRCodeScreenScanner.tsx` — HOME_SCANNER case + enabled params
6. `src/hooks/useWalletConnectQrCodeScanner.tsx` — Migrate + enabled param
7. `src/hooks/useSendFlowQrCodeScanner.tsx` — Migrate + error handling + enabled param
8. `src/hooks/useHomeHeaders.tsx` — Change source
9. `src/components/screens/AccountQRCodeScreen/AccountQRCodeScreen.tsx` — Tabs
10. `src/components/QRScanner.tsx` — Fix analytics arg order
11. `src/i18n/locales/en/translations.json`
12. `src/i18n/locales/pt/translations.json`
13. Test files

## Edge Cases

- Self-send (G): parseQRPayload → error toast
- Self-send (muxed/federation): searchAddress() → searchError → error toast
- Federation failure: searchAddress() → searchError → error toast
- Destination status failure: searchAddress() → searchError → error toast
- Unfunded destinations: isDestinationFunded: false, send flow handles
- Stack buildup: navigation.replace() with fade for tab switches
- Back-navigation: popToTop() before navigate ensures correct back target
- Stale store state: isProcessingQRScan flag gates all effects
- Inactive hook effects: enabled parameter prevents side effects
- Rescan after error: Not supported (processedCodes limitation, accepted)

## Out of Scope

- Dev-only manual WC URI paste input
- QRScanner processedCodes reset on transient error
- Stable error codes in sendRecipientStore

---

Closes #761


JakeUrban (commented):
The figma link provided includes changes that were not intended to be in scope of this PR.

Functionally, this PR should enable the QR code scanner available from the account/home view to scan both wallet connect URI's in addition to stellar addresses (G, C, M, federated addresses). When addresses are scanned, users should be routed to the send flow, with the address pre-filled and XLM auto-selected.

Visually, very little should change. The only design changes that should be included in this PR are mocked in this [figma view](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=12045-40509&t=t9XMH1fq9L3Q8KvG-4), with an important caveat: 

Do not support the Receive/Scan tabs. In its place should be "Scan a QR code", styled identically to how "Send funds" is [here](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=12168-108051&t=t9XMH1fq9L3Q8KvG-4).

Additionally, the QR code scanner available from the "Send to" menu should be designed similarly to the account/home view scanner, as designed [here](https://www.figma.com/design/KwkHXQxbNmDllwermJtnRu/Freighter-Mobile?node-id=12168-108051&t=t9XMH1fq9L3Q8KvG-4). Note that this scanner view is send-specific.

What this PR includes that it shouldn't: 

- After selecting the account/home view QR code scanner, the "Scan QR Code" view includes another, but different, QR code scanner icon in the top right. Remove this. Note this also removes the view it routes to, which is out of scope

__tests__/components/screens/AccountQRCodeScreen/AccountQRCodeScreen.test.tsx
```diff
@@ -48,6 +48,8 @@ jest.mock("hooks/useAppTranslation", () => () => ({
       "accountQRCodeScreen.helperText": "Helper Text",
       "accountQRCodeScreen.bottomSheet.title": "Bottom Sheet Title",
       "accountQRCodeScreen.bottomSheet.description": "Bottom Sheet Description",
+      "accountQRCodeScreen.tabs.receive": "Receive",
+      "accountQRCodeScreen.tabs.scan": "Scan",
```
JakeUrban: We removed the tabs implementation in an earlier commit, so I believe these should be removed.

src/config/constants.ts
```diff
@@ -413,6 +413,8 @@ export enum QRCodeSource {
   ADDRESS_INPUT = "address_input",
   /** For scanning WalletConnect URIs */
```
JakeUrban: This PR updates the QR code scanner available in the account/home view to also support scanning addresses in addition to wallet connect URIs. If HOME_SCANNER handles both, why do we still need WALLET_CONNECT? Is WALLET_CONNECT used in a different view? If not, the walllet connect QR code scanner should be removed entirely.

src/hooks/useHomeQrCodeScanner.tsx
```diff
@@ -0,0 +1,259 @@
...
+  error: string;
+  showManualInput: false;
+  title: string;
+  scannerTitle: string;
+  context: QRCodeSource.HOME_SCANNER;
+}
+
+interface QRCodeScreenConfig {
+  showHeaderRight: false;
+}
```
JakeUrban: When is this config object used? If `showHeaderRight` is always false, remove it.

src/components/WalletConnectManualInputOverlay.tsx
```diff
@@ -1,109 +0,0 @@
-import { logos } from "assets/logos";
```
JakeUrban, line 1: Lets keep this component and use it when the application is in dev mode on the home scanner.

src/config/routes.ts
```diff
@@ -140,17 +140,16 @@ export const ALL_ROUTES_OBJECT = [
   SWAP_ROUTES,
 ] as const;
 
+export type AccountQRCodeScreenParams = undefined;
+
```
JakeUrban: We can assign undefined directly for the relevant key now.

src/hooks/useHomeQrCodeScanner.tsx
```diff
@@ -0,0 +1,258 @@
...
+    resetSendRecipient,
+  } = useSendRecipientStore();
+  const { saveRecipientAddress, saveMemo, saveMemoType } =
+    useTransactionSettingsStore();
+
+  const state: QRCodeScreenState = {
+    manualInput: "",
+    isConnecting: isConnectingWC || (isProcessingAddress && isSearching),
+    error: "",
+    showManualInput: false,
```
JakeUrban: This is currently never `true`, but per my other comment, we should make it true when in dev mode, and the manual input overlay should be present in that case.

piyalbasu:
### Possible double-navigation when scanning from the home QR scanner

**Not a merge blocker** — low likelihood, low blast radius (no funds move; the send screen still requires manual confirmation). Flagging for awareness + a cheap guard.

**TL;DR:** If the home scanner reads a Stellar address and a WalletConnect QR code back-to-back (e.g. off a payment page that shows both), the app can navigate twice — landing you on the send screen and then unexpectedly bouncing you back to home about a second later. It doesn't require deliberate fast scanning: a single page showing two QR codes plus normal camera drift can trigger it, and a federation address (which needs a network lookup) widens the window.

**Steps to reproduce:**
1. Open the home QR scanner.
2. Scan a valid Stellar address QR. Resolution is async and does not navigate immediately. (Use a federation address like `name*domain.com`, or throttle the network, to keep this in flight for >1s.)
3. While that resolution is still pending, let the camera read a different QR that is a WalletConnect URI (`wc:...`) — e.g. by drifting the camera across a page that shows both codes.
4. The address resolves and navigates you to the send screen.
5. ~1s later the WalletConnect pairing completes and pops you back off the send screen to home.

(The reverse order — WC first, address second — produces the same class of conflicting navigation.)

---

<details>

**Root cause:** `handleQRCodeScanned` has no entry guard, so a second scan of a *different* type is fully processed while the first is still in flight. The address path and the WalletConnect path then each issue navigation independently, with no awareness of the other.

The scanner's dedup only blocks re-scanning the *same* string plus a 1s debounce between any two scans — it does not block a different code from being processed mid-resolution:

https://github.com/stellar/freighter-mobile/blob/f5a30deb2d7f52559f92f0a7f81b725c6fdf40bd/src/components/QRScanner.tsx#L96-L124

`handleQRCodeScanned` switches on payload type with no `isProcessingAddress` / `isConnectingWC` guard at the top, so both states can become true at once:

https://github.com/stellar/freighter-mobile/blob/f5a30deb2d7f52559f92f0a7f81b725c6fdf40bd/src/hooks/useHomeQrCodeScanner.tsx#L115-L154

The WalletConnect success path schedules a delayed `handleClose()` (which is `navigation.popToTop()`):

https://github.com/stellar/freighter-mobile/blob/f5a30deb2d7f52559f92f0a7f81b725c6fdf40bd/src/hooks/useHomeQrCodeScanner.tsx#L107-L135

…while the address-resolution success effect guards only on `!enabled || !isProcessingAddress` (no `isConnectingWC` check) and calls `popToTop()` + `navigate(SEND_PAYMENT_STACK)`:

https://github.com/stellar/freighter-mobile/blob/f5a30deb2d7f52559f92f0a7f81b725c6fdf40bd/src/hooks/useHomeQrCodeScanner.tsx#L188-L218

Because `searchAddress()` is async (federation `.toml` resolution + Horizon account-existence check), the window between the address scan and its navigation is routinely >1s, which is all the debounce requires to allow the second scan through.

**Deterministic repro without a camera:** drive `handleQRCodeScanned` directly — call it with a valid address (leave the `searchAddress` mock pending), call it again with a `wc:` URI, resolve `walletKit.pair()`, then resolve the address search, and assert `navigation.popToTop` is called twice.

1. *Cheap guard:* early-return at the top of `handleQRCodeScanned` when `isProcessingAddress || isConnectingWC` is already true — ignore a new scan until the first settles.
2. *Root cause:* collapse `isProcessingAddress` and `isConnectingWC` into a single `status: "idle" | "searching_address" | "connecting_wc"` enum so "busy" is a mutually-exclusive state. This also removes a related latent issue: the WC success path never resets `isConnectingWC`, which can leave the connecting spinner stuck if `popToTop()` doesn't unmount the screen.

</details>

src/hooks/useHomeQrCodeScanner.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+        case "invalid":
+          if (lastErrorCode !== data) {
+            showToast({
+              variant: "error",
+              title:
+                payload.error === QRCodeError.SELF_SEND
+                  ? t("sendPaymentScreen.cannotSendToSelf")
+                  : t("sendPaymentScreen.invalidAddress"),
+              duration: 3000,
```
piyalbasu, line 164: NIT: let's const the duration length here since we're using the same value for both toasts

piyalbasu (commented):
From a code perspective, this lgtm other than the 2 small comments I posted. 

@CassioMG Do you have thoughts on the best way to test this? If I can confirm this works on my phone, happy to approve

piyalbasu (approved):
Okay, it took me way too long, but I was able to get this running on my device and confirm that scanning an address works!

PR #896, opened 2026-06-10
feat: encrypt tab screenshots at rest, remove blur

JakeUrban:
Tab screenshots in the Discover browser were stored as a single unencrypted JSON blob in AsyncStorage, blurred and compressed to 25% quality to limit exposure. This PR implements the approach from the [spike](https://github.com/stellar/wallet-eng-monorepo/issues/9): encrypt screenshots at rest so we can remove the blur and raise quality — a win on both security and UX simultaneously.

## What changed

**Encryption.** Each screenshot is now encrypted with AES-256-GCM before it touches disk. A random 256-bit data-encryption key (DEK) is generated on first use, persisted to Keychain under a new non-biometric config (`WHEN_UNLOCKED_THIS_DEVICE_ONLY`, no `accessControl`), and cached in memory for the session. The key never leaves the device and does not require the wallet to be unlocked or a biometric prompt to render thumbnails.

**Storage shape.** Images moved off AsyncStorage and onto the filesystem, one encrypted file per tab at `<CachesDir>/tab-screenshots/<tabId>.enc`. AsyncStorage now holds only metadata (tabId, tabUrl, timestamp, filename) — a few hundred bytes regardless of how many tabs are open. This eliminates the Android `CursorWindow` crash risk: the old single-blob design would have exceeded the ~2 MB SQLite row limit once blur was removed and image sizes grew.

**Blur removed, quality raised.** `react-native-fast-opencv` is removed. Capture quality goes from 0.25 to 0.5. Security comes from encryption, not from degrading the image.

**Cipher choice.** The spike benchmarked the existing `tweetnacl.secretbox` helpers on the actual Hermes VM (no JIT): ~12 ms per 30 KB screenshot, projecting to ~35–60 ms/image on an average Android device. Native AES-GCM via `react-native-quick-crypto` (JSI/Nitro) is ~0.03 ms — the same `crypto.subtle` surface the extension's security team vetted, now available on mobile.

**Wipe path.** The DEK is cleared in `clearTemporaryData` alongside the wallet's derived key, so it's wiped on the same logout path that wipes all other sensitive data. `clearAllScreenshots` now also deletes the `.enc` files.

**Upgrade migration.** `migrateOldScreenshots()` runs once on first load after upgrade: if it finds the old blob (detected by the presence of `uri` fields), it deletes the key and rotates the DEK. Thumbnails regenerate on the next browse session; `TabPreview` already handles missing screenshots with favicon/home fallbacks.

## Key technical decisions

- **Non-biometric Keychain config for the DEK** — a new `SCREENSHOT_KEYCHAIN_OPTIONS` alongside the existing two tiers. Biometric gating (`BIOMETRY_ANY_OR_DEVICE_PASSCODE`) on the existing `SECURE_KEYCHAIN_GET_OPTIONS` would prompt Face ID/passcode every time the Tabs Overview renders.
- **DEK is independent of the wallet derived key** — thumbnails should render on the lock screen without unlocking the wallet. A separate random DEK achieves this; it has no bearing on key material security.
- **No lazy per-tile decryption in the UI layer** — with native AES-GCM the cipher itself is free; the existing `loadScreenshots` parallel-loads all tabs when the overview is opened, which is effectively lazy enough. Avoids adding async loading state to `TabPreview`.
- **One `.enc` file per tab, not a zip or database** — simplest possible eviction (delete the file on tab close), no cross-file locking, O(1) writes instead of the old O(total screenshots) reserialize-everything pattern.

## Before merging

- [ ] `yarn install` and verify the build — confirm Nitro peer wiring under New Arch, check Android 16 KB page-size alignment (similar to the existing `setup-scrypt-16kb.js` scrypt patch)
- [ ] On-device CursorWindow validation — store a >2 MB AsyncStorage value on a low-end Android; confirm the crash; confirm the new design never approaches the limit
- [ ] On-device cipher benchmark — encrypt/decrypt a 60 KB blob 30× on a mid-range Android; expect sub-ms per operation
- [ ] User-acceptance check — validate 0.5-quality thumbnails are sharp enough on real devices before shipping
- [ ] Visual check — open Tabs Overview; confirm sharp, unblurred thumbnails
- [ ] Security check — inspect `<CachesDir>/tab-screenshots/*.enc` directly; confirm binary ciphertext, not a readable JPEG
- [ ] Wipe path — trigger logout with wipe; confirm `.enc` files deleted and DEK removed from Keychain
- [ ] Upgrade migration — install over a build with the old `browser_screenshots` blob; confirm old key cleared on first launch

<details>
<summary>Implementation details</summary>

### New file: `src/helpers/screenshotCrypto.ts`

Owns DEK lifecycle and AES-256-GCM operations:

- `getOrCreateScreenshotDek()` — memory cache → Keychain → generate fresh 32-byte random key
- `encryptScreenshot(dataUri)` — packs `[0x01 version | 12-byte IV | ciphertext+GCM tag]`, returns base64
- `decryptScreenshot(base64)` — unpacks and decrypts, returns `data:image/jpeg;base64,...`
- `clearScreenshotDek()` — clears cache and removes Keychain entry (wipe path)
- `resetScreenshotDek()` — clear + generate fresh (migration path)

### Changed: `src/helpers/screenshots.ts`

Interface change: `ScreenshotData` now extends `ScreenshotMeta` (the stored shape). `ScreenshotMeta` has `{ tabId, tabUrl, timestamp, file }` — no `uri`. `ScreenshotData` adds `uri` (in-memory only, never written to storage).

Function signature change: `saveScreenshot(tabId, tabUrl, plaintextUri)` replaces the old `saveScreenshot(ScreenshotData)` that accepted a pre-built object including the image bytes.

Removed: `getBlurredImage()` and all `react-native-fast-opencv` imports. Removed `OpenCV.clearBuffers()` from the `captureTabScreenshot` finally block.

Added: `migrateOldScreenshots()` — one-time upgrade path.

### Changed: `src/services/storage/keychainSecurityConfig.ts`

Added `SCREENSHOT_KEYCHAIN_OPTIONS`: `{ accessible: WHEN_UNLOCKED_THIS_DEVICE_ONLY }`. Modeled on the existing `INDEX_KEYCHAIN_OPTIONS` pattern; no `accessControl` so reads never trigger a biometric prompt.

### Changed: `src/config/constants.ts`

- `SCREENSHOT_QUALITY`: 0.25 → 0.5
- Added `SCREENSHOT_DEK_SERVICE: "screenshot_dek"` (Keychain service key)
- Added `SCREENSHOT_FILES_DIR: "tab-screenshots"` (subdirectory under `CachesDirectoryPath` — kept out of iCloud/iTunes backups, reclaimable by the OS)
- Removed `SCREENSHOT_BLUR_STRENGTH` object

### Changed: `src/services/storage/helpers.ts`

`clearTemporaryData` now includes `clearScreenshotDek()` in its `Promise.all`, alongside HASH_KEY, TEMPORARY_STORE, and DERIVED_KEY.

### Changed: `src/ducks/browserTabs.ts`

`loadScreenshots` calls `migrateOldScreenshots()` before the parallel screenshot load. The rehydration path that calls `loadScreenshots` runs it on the first launch after upgrade.

### Dependencies

- Added: `react-native-quick-crypto ^1.1.5`, `react-native-nitro-modules ^0.31.2`, `react-native-quick-base64 ^3.0.0`
- Removed: `react-native-fast-opencv 0.4.6`

`@dr.pogodin/react-native-fs` was already a dependency — no new modules needed for filesystem I/O.

</details>


src/config/constants.ts
```diff
@@ -340,20 +340,14 @@ export const BROWSER_CONSTANTS = {
   MAX_SCREENSHOTS_STORED: 30,
   MAX_ACTIVE_WEBVIEWS: 10, // Maximum number of active WebView instances
   SCREENSHOT_FORMAT: "jpg",
-  SCREENSHOT_QUALITY: 0.25,
+  SCREENSHOT_QUALITY: 0.5,
```
JakeUrban, line 343: We need to validate the quality is sufficient here with user-acceptance testing

package.json
```diff
@@ -124,8 +123,11 @@
     "react-native-keychain": "10.0.0",
     "react-native-level-fs": "2.0.0",
     "react-native-localize": "3.5.2",
+    "react-native-nitro-modules": "^0.31.2",
     "react-native-permissions": "5.4.2",
     "react-native-qrcode-svg": "6.3.15",
+    "react-native-quick-base64": "^2.1.0",
```
JakeUrban: **Dependency conflict:** `react-native-quick-crypto@1.1.5` declares a peer dependency of `react-native-quick-base64 >= 3.0.0`, but this pins `^2.1.0`. The lockfile confirms the problem: it now resolves both `react-native-quick-base64@2.2.2` (this entry) and `@3.0.0` (pulled in via `@craftzdog/react-native-buffer`). Two versions of the same native module in the tree is exactly the situation RN autolinking handles badly — it will link one copy (the root 2.2.2) while quick-crypto's JS expects the >=3 API.

Was `^2.1.0` chosen deliberately for compatibility with something else? If not, this should be `^3.0.0` so the tree dedupes to a single version. This is the kind of thing the "verify the build / Nitro peer wiring" checklist item would catch, but it's visible statically already.

src/helpers/screenshotCrypto.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Returns the screenshot data-encryption key (DEK), creating and persisting it on first use.
+ *
+ * Flow: memory cache → Keychain → generate fresh.
+ * The DEK is a random 256-bit key, independent of the wallet derived key.
+ * It is stored under a non-biometric Keychain entry so thumbnails render
+ * without prompting the user.
+ */
+export const getOrCreateScreenshotDek = async (): Promise<CryptoKey> => {
```
JakeUrban: **Race on first use:** if two `encryptScreenshot`/`decryptScreenshot` calls hit this concurrently before the cache is populated (e.g. parallel captures, or `loadScreenshots`' `Promise.all` on first launch), both miss `_dekCache`, both find nothing in Keychain, and both run `_generateAndStoreDek()` with *different* random keys. Files encrypted under whichever key loses the Keychain write become permanently undecryptable after restart.

The standard fix is to memoize the in-flight promise rather than the resolved key:

```ts
let _dekPromise: Promise<CryptoKey> | null = null;

export const getOrCreateScreenshotDek = (): Promise<CryptoKey> => {
  if (!_dekPromise) {
    _dekPromise = loadOrGenerateDek().catch((e) => {
      _dekPromise = null; // don't cache failures
      throw e;
    });
  }
  return _dekPromise;
};
```

Since the data is regenerable thumbnails the blast radius is small, but the fix is cheap and removes a whole class of "thumbnails randomly disappear after update" reports.

src/helpers/screenshotCrypto.ts
```diff
@@ -0,0 +1,165 @@
...
+        "raw",
+        dekBytes,
+        { name: "AES-GCM", length: 256 },
+        false,
+        ["encrypt", "decrypt"],
+      );
+      _dekCache = key;
+      return key;
+    }
+  } catch (error) {
```
JakeUrban: **Transient read error destroys the existing DEK:** this catch treats *any* Keychain read failure as "no key exists" and falls through to `_generateAndStoreDek()`, which overwrites the stored entry. Keychain/Keystore reads can fail transiently (Android keystore briefly unavailable, app woken in background before first unlock, etc.) — in that case a perfectly valid DEK gets rotated and every existing screenshot becomes undecryptable.

Consider only generating when the read *succeeds and returns no entry* (`stored === false`), and rethrowing (or returning a sentinel that makes encrypt/decrypt fail soft) on an actual error. `findTabScreenshot` already handles decrypt failures gracefully, so failing the single render is strictly better than rotating the key.

src/helpers/screenshotCrypto.ts
```diff
@@ -0,0 +1,165 @@
...
+    logger.error("screenshotCrypto", "Failed to clear DEK from Keychain:", error);
+  }
+};
+
+/**
+ * Clears the existing DEK and generates a fresh one.
+ * Used after migration wipes the old unencrypted screenshot blob.
+ */
+export const resetScreenshotDek = async (): Promise<CryptoKey> => {
```
JakeUrban, line 171: **Dead code / description mismatch:** `resetScreenshotDek` is exported but never called anywhere in the app — and the PR description says the migration "deletes the key and rotates the DEK", which this was presumably meant to implement. `migrateOldScreenshots()` only removes the AsyncStorage blob.

In practice rotation is a no-op on the upgrade path (the DEK can't exist before this feature ships), so I'd either wire this into `migrateOldScreenshots()` to match the stated design, or delete the function and fix the PR description. The last commit removed its assertions from the migration tests, which suggests the call itself was dropped unintentionally.

src/helpers/screenshots.ts
```diff
@@ -1,87 +1,116 @@
...
+
+    const filePath = encFilePath(meta.file);
+    if (!(await exists(filePath))) {
+      // Stale metadata — clean it up silently
+      metaMap.delete(tabId);
+      await AsyncStorage.setItem(
+        BROWSER_CONSTANTS.SCREENSHOT_STORAGE_KEY,
+        JSON.stringify(Object.fromEntries(metaMap)),
+      );
+      return null;
```
JakeUrban, line 86: **Concurrent read-modify-write of the whole metadata map:** `loadScreenshots` calls `findTabScreenshot` for every tab in parallel (`Promise.all`). If more than one tab has stale metadata — exactly the state every user is in right after the migration wipes files, or after an iCloud restore where the DEK didn't travel — each call rewrites the *entire* map from its own stale snapshot. Deletions can be lost or freshly-written entries from a concurrent `saveScreenshot` clobbered.

The simplest fix: don't write here at all — just `return null`. Stale entries are harmless (this function already returns null for them every time), and `cleanupScreenshots`/`pruneScreenshots` runs right after `loadScreenshots` and could own stale-file cleanup in one serialized pass. If you keep the inline cleanup, it needs to be serialized with the other map writers.

src/helpers/screenshots.ts
```diff
@@ -1,87 +1,116 @@
...
 /**
- * Retrieves all stored screenshots from persistent storage as a Map for O(1) lookups.
- * @returns Promise<Map<string, ScreenshotData>> - Map of tabId to screenshot data
+ * ScreenshotMeta plus the decrypted data URI (in-memory only, never persisted).
+ */
+export interface ScreenshotData extends ScreenshotMeta {
+  uri: string;
+}
+
+const screenshotsDir = (): string =>
```
JakeUrban, line 42: **Minor — consider `CachesDirectoryPath` (or marking the dir as excluded from backup):** on iOS, everything under Documents is backed up to iCloud/iTunes by default. These files are (a) regenerable caches and (b) encrypted under a `THIS_DEVICE_ONLY` key, so after a restore onto a new device they're undecryptable dead weight — the app handles that gracefully via the missing-file/decrypt-failure paths, but we're still shipping useless ciphertext into users' backups. Caches also lets the OS reclaim the space under disk pressure, which is the right behavior for thumbnails.

JakeUrban (commented):
## Review summary

The overall approach is sound: native AES-256-GCM via `react-native-quick-crypto`, one encrypted file per tab with metadata-only AsyncStorage, a random DEK in a non-biometric Keychain entry, and wipe/migration paths wired into the existing lifecycle. The versioned `[version | IV | ciphertext+tag]` envelope with a fresh random 12-byte IV per encryption is correct GCM usage, and the graceful-null fallback on any decrypt failure means none of the findings below can brick the UI — worst case is a regenerated thumbnail.

### Findings, highest impact first

1. **Dependency conflict (likely build/runtime breakage)** — `react-native-quick-base64` is pinned at `^2.1.0`, but `react-native-quick-crypto` requires `>=3.0.0` as a peer. The lockfile now resolves *both* 2.2.2 and 3.0.0, two versions of the same native module. Inline comment on `package.json`.
2. **DEK creation race** — concurrent first-use calls to `getOrCreateScreenshotDek()` can generate two different DEKs; files encrypted under the loser become permanently undecryptable. Memoizing the in-flight promise fixes it. Inline comment.
3. **Transient Keychain read error silently rotates the DEK** — the catch-all fallback to `_generateAndStoreDek()` overwrites a valid stored key on any read failure, orphaning all existing screenshots. Inline comment.
4. **Concurrent read-modify-write of the metadata map** — `findTabScreenshot`'s stale-metadata cleanup rewrites the whole AsyncStorage map, and `loadScreenshots` runs it for every tab in parallel. Lost updates are possible. Inline comment.
5. **`resetScreenshotDek` is dead code and the PR description doesn't match the implementation** — the description says migration "rotates the DEK", but `migrateOldScreenshots` never calls it. Inline comment.
6. **Minor: storage location** — encrypted screenshots live in the iOS Documents directory, which is iCloud-backed; Caches is a better fit for regenerable, device-keyed data. Inline comment.

Also a doc nit, no code change needed: the PR description says `react-native-nitro-modules ^0.24.0` but `package.json` pins `^0.31.2`.

None of 2–4 leak plaintext — they only affect availability of thumbnails, which regenerate on the next browse. Item 1 is the one I'd treat as blocking since it likely breaks the build or autolinking; it should surface in the `yarn install` verification step already listed in the PR checklist.

package.json
JakeUrban: Fixed in cc51d600 — bumped to `^3.0.0`; the lockfile now resolves a single `react-native-quick-base64@3.0.0`. `^2.1.0` was not deliberate. Worth noting the Android/iOS build jobs failed on the previous commit — plausibly this exact dual-native-module issue, so the next CI run should tell us.

src/helpers/screenshotCrypto.ts
JakeUrban: Fixed in cc51d600 with the in-flight promise memoization as suggested. Failures are not cached (the promise slot is cleared on rejection), so a transient error does not permanently wedge encryption for the session.

src/helpers/screenshotCrypto.ts
JakeUrban: Fixed in cc51d600 — generation now only happens when the Keychain read succeeds and finds no entry. Read errors propagate out of `getOrCreateScreenshotDek`, so encrypt/decrypt fail soft: `findTabScreenshot` catches and falls back to favicon/home for that render, and the DEK is never rotated.

src/helpers/screenshotCrypto.ts
JakeUrban: Fixed in cc51d600 — wired `resetScreenshotDek()` into `migrateOldScreenshots()` to match the stated design, and restored the migration test assertions. You are right that the call was dropped unintentionally in the last commit.

src/helpers/screenshots.ts
JakeUrban: Fixed in cc51d600 — took the simplest option: `findTabScreenshot` no longer writes, it just returns null on stale metadata. Reconciliation is owned by `pruneScreenshots` and the next `saveScreenshot` for that tab, both of which already rewrite the map in a single pass.

src/helpers/screenshots.ts
JakeUrban: Fixed in cc51d600 — moved storage to `CachesDirectoryPath`. Regenerable, device-bound ciphertext now stays out of iCloud/iTunes backups and the OS can reclaim the space under disk pressure.

src/config/constants.ts
JakeUrban: Added a user-acceptance item to the Before merging checklist — thumbnails at 0.5 quality need a visual sign-off on real devices before this ships.

piyalbasu:
### Code review

Found 1 issue:

1. The screenshot DEK is wiped on non-wipe paths, silently losing all tab thumbnails. `clearScreenshotDek()` is documented as only being called on full wallet wipe ("Called on wallet wipe (logout with shouldWipeAllData)"), but it is wired into `clearTemporaryData()`, which also runs on two non-wipe paths: the repeated-decryption-failure branch in `getTemporaryStore` ([`auth.ts:680`](https://github.com/stellar/freighter-mobile/blob/71a6796f0e2029b779107a70604ad35ad24e16f2/src/ducks/auth.ts#L679-L681)) and `clearAllData` for signUp/importWallet ([`auth.ts:1031`](https://github.com/stellar/freighter-mobile/blob/71a6796f0e2029b779107a70604ad35ad24e16f2/src/ducks/auth.ts#L1029-L1034)). Only the actual logout-wipe path ([`auth.ts:2064-2070`](https://github.com/stellar/freighter-mobile/blob/71a6796f0e2029b779107a70604ad35ad24e16f2/src/ducks/auth.ts#L2063-L2070)) also calls `clearAllScreenshots`. On the two non-wipe paths the Keychain DEK is deleted while the `.enc` files and their metadata remain on disk; the next `loadScreenshots` generates a fresh random DEK, so every existing file fails AES-GCM auth-tag verification, `findTabScreenshot` returns `null`, and all tab previews silently disappear (orphaned `.enc` files for still-active tabs are never cleaned up on these paths). This contradicts the design intent that the DEK is independent of wallet material and "survives the session".

The DEK-clearing call site:

https://github.com/stellar/freighter-mobile/blob/71a6796f0e2029b779107a70604ad35ad24e16f2/src/services/storage/helpers.ts#L12-L19

And the comment it contradicts:

https://github.com/stellar/freighter-mobile/blob/71a6796f0e2029b779107a70604ad35ad24e16f2/src/helpers/screenshotCrypto.ts#L148-L152

JakeUrban:
Valid finding — confirmed all three `clearTemporaryData` call sites. Fixed in 5c3bf3d: moved `clearScreenshotDek()` out of `clearTemporaryData()` and call it explicitly on the logout-wipe path, right after `clearAllWebViewData()` deletes the files and metadata it protects. The DEK now survives the two non-wipe paths (signUp/importWallet via `clearAllData`, and the repeated-decryption-failure guard in `getTemporaryStore`), consistent with browser tabs, cookies, and screenshot files all surviving those paths — and with the documented design that the DEK is independent of wallet key material. Added regression assertions: DEK cleared on `logout(true)`, not cleared on the decryption-failure path.

CassioMG (approved):
LGTM - tested on both Android and iOS, the screenshot quality looks good and the migration also worked as expected

PR #898, opened 2026-06-10
fix: default feature flags on for all dev builds, not just __DEV__

JakeUrban:
## Summary

Feature flags (swap, discover, onramp) are defaulted to **enabled** in environments where Amplitude API keys aren't configured and real flags therefore can't be fetched. That gate previously used the React Native `__DEV__` constant, which is only `true` for the local Metro/debug bundle. This switches it to `isDev`, which keys off the `freighterDev` bundle ID.

## Functional changes to builds

| Build                                  | Before (`__DEV__`)        | After (`isDev`)        |
|----------------------------------------|---------------------------|------------------------|
| Local Metro / debug                    | flags default **on**      | flags default **on**   |
| CI preview build (`freighterDev`)      | flags default **off** ⚠️  | flags default **on** ✅ |
| Dev TestFlight build (`freighterDev`)  | flags default **off** ⚠️  | flags default **on** ✅ |
| Production / release                   | flags default **off**     | flags default **off**  |

CI preview builds and the dev TestFlight build use the `freighterDev` bundle ID but produce release-mode JS, so `__DEV__` is `false` for them. They don't ship Amplitude keys either, so feature flags couldn't be fetched **and** weren't defaulted on — leaving gated features (swap, discover, onramp) hidden on those dev distributions. Using `isDev` (true for any `freighterDev`-bundle build) means all dev builds now enable these features by default, while production behavior is unchanged.

This is the same `isDev`/`__DEV__` change made in #892, extracted on its own so it can land independently of that QR-scanner feature work.

## Changes

- `src/ducks/remoteConfig.ts`: import `isDev` from `helpers/isEnv`; gate the initial feature-flag state on `isDev || isE2ETest` instead of `__DEV__ || isE2ETest`; update the explanatory comment.

## Testing

- `__tests__/ducks/remoteConfig.test.ts` passes (16/16).
- Note: an unrelated, pre-existing failure in `SwapAmountScreen.test.tsx` (`saveFee is not a function`) also fails on clean `main` and is not caused by this change.

### DIFF
--- a/src/ducks/remoteConfig.ts
+++ b/src/ducks/remoteConfig.ts
@@ -1,7 +1,7 @@
 import { logger } from "config/logger";
 import { useNetworkStore } from "ducks/networkInfo";
 import { isAndroid } from "helpers/device";
-import { isE2ETest } from "helpers/isEnv";
+import { isDev, isE2ETest } from "helpers/isEnv";
 import { getBundleId, getVersion } from "react-native-device-info";
 import { ANALYTICS_CONFIG } from "services/analytics/constants";
 import { getExperimentClient } from "services/analytics/core";
@@ -60,10 +60,10 @@ interface RemoteConfigState extends FeatureFlags {
 // Get current app version for default values
 const currentAppVersion = getVersion();
 
-// While developing locally or during e2e tests we don't set the Amplitude API keys which prevents
-// us from fetching feature flags so let's set all "true" by default in __DEV__ or isE2ETest
+// In dev builds (local Metro, CI previews, TestFlight dev) or e2e tests, Amplitude keys
+// are not configured so feature flags can't be fetched — default all features to enabled.
 const INITIAL_REMOTE_CONFIG_STATE =
-  __DEV__ || isE2ETest
+  isDev || isE2ETest
     ? {
         swap_enabled: true,
         discover_enabled: true,
--- a/src/ducks/remoteConfig.ts
+++ b/src/ducks/remoteConfig.ts
@@ -1,7 +1,7 @@
 import { logger } from "config/logger";
 import { useNetworkStore } from "ducks/networkInfo";
 import { isAndroid } from "helpers/device";
-import { isDev, isE2ETest } from "helpers/isEnv";
+import { isDev } from "helpers/isEnv";
 import { getBundleId, getVersion } from "react-native-device-info";
 import { ANALYTICS_CONFIG } from "services/analytics/constants";
 import { getExperimentClient } from "services/analytics/core";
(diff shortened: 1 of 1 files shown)


piyalbasu:
### Optional: consider `isDev || __DEV__` to keep the dev-default a true superset

**Polish / suggestion — Not a merge blocker.** Purely optional consistency nit. It does not affect CI, PR previews, normal local development, or production; it only changes one niche local-only build configuration.

**TL;DR:** This change does exactly what it should — iOS PR-preview builds now start with onramp/swap/discover enabled. The only thing worth a second look is that it *replaces* the old "is this a debug build?" check with a "is this the dev app?" check, instead of combining them. As a result, one uncommon local setup — running the **production** app variant under the debugger on iOS — would now start with those three features defaulted **off** rather than on. That's arguably fine (you may even want prod-like defaults when debugging the prod app), so this is just a heads-up, not a fix request.

---

<details>

**Root cause:** the two checks answer different questions and only diverge in one quadrant.

- `__DEV__` — true for any Debug build, false for any Release build (build *configuration*).
- `isDev` — true when the bundle id is `org.stellar.freighterdev` (which *target* was built), regardless of Debug/Release.

See their definitions:
https://github.com/stellar/freighter-mobile/blob/fb9f193d0b52ecbc259ef1f6b975db029283e337/src/helpers/isEnv.ts#L9-L13

The condition changed from `__DEV__ || isE2ETest` to `isDev || isE2ETest`:
https://github.com/stellar/freighter-mobile/blob/fb9f193d0b52ecbc259ef1f6b975db029283e337/src/ducks/remoteConfig.ts#L65-L66

Truth table across real build contexts (the three boolean flags are the only difference between the two branches):

| Build context | `__DEV__` / `isDev` | Before | After |
|---|---|---|---|
| iOS PR preview (dev scheme, Release) | F / T | off on iOS | **on** ✅ the fix |
| Production (prod scheme, Release) | F / F | else-branch | else-branch (unchanged) |
| Local dev — `yarn ios` (dev scheme, Debug) | T / T | on | on (unchanged) |
| Local **prod** scheme in Debug (`yarn ios-prod`) | T / F | on | **off on iOS** ← only delta |

So the fix is correctly targeted (the preview build uses the `freighter-mobile-dev` scheme in Release config → `isDev` true, `__DEV__` false), and production is provably untouched. The sole behavior change outside the goal is the bottom row, which is a deliberate, non-automated action.

**Two minor notes:**

1. `isE2ETest` is now redundant in this condition. It's defined as `isDev && Config.IS_E2E_TEST === "true"`, so `isE2ETest` implies `isDev` — meaning `isDev || isE2ETest` is logically just `isDev`. Harmless, but the `|| isE2ETest` no longer adds coverage.

2. The sibling store uses `isDev || __DEV__` as the "any dev context" idiom:
https://github.com/stellar/freighter-mobile/blob/fb9f193d0b52ecbc259ef1f6b975db029283e337/src/ducks/debug.ts#L43

1. *Leave as-is* (recommended if "prod scheme should always behave like prod" is intended) — optionally drop the now-redundant `|| isE2ETest` for clarity.
2. *Match the title's "all dev builds" wording* — use `isDev || __DEV__`, a true superset that also keeps the local-prod-debug case enabled and matches the `debug.ts` idiom. (`isE2ETest` is subsumed by `isDev` either way.)

</details>

piyalbasu (approved):
Approving but leaving one comment about something I noted during review

leofelix077 (approved):

PR #920, opened 2026-07-07
feat(auth): derive anonymous backend auth keypair (#864)

### LINKED ISSUE #864: [Mobile] Derive auth keypair from seed for backend authentication
Derive the Ed25519 auth keypair on mobile via `HMAC-SHA256(seedBytes, "freighter-auth-v1")`. The hex-encoded pubkey is the user's anonymous backend user ID — separate from any Stellar G keypair, never used for wallet signing.

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — Auth Flow + Key properties.

Blocks: [Mobile] Generate per-request backend JWT in `apiFactory.ts`.

**Acceptance:**
- Same seed produces identical auth keypair (and therefore identical user ID) on extension and mobile.
- Auth pubkey is not a valid Stellar G address (cryptographically independent from the wallet keypair).
- No wallet-signing prompt is triggered by derivation or use.

piyalbasu:
## TL;DR

Closes #864. Derives an anonymous, per-user auth keypair from the wallet seed so Freighter Mobile can (in a follow-up) authenticate its calls to `freighter-backend-v2`. The derived key is cryptographically independent from the wallet signing key, never triggers a wallet-signing prompt, and produces the **same user id as the Freighter extension** for the same mnemonic (locked by shared cross-platform test vectors). The keypair is memoized in memory for the unlocked session and dropped on every session-ending path — logout, lock, soft-lock, hash-key hard-expiry, and wallet import — never persisted.

This is PR 1 of 2. PR 2 (#865) adds the per-request JWT and wires it into the backend-v2 axios client.

<details>

**What changed:**

- `src/services/auth/deriveAuthKeypair.ts` (new) — pure, **async** (bip39's `mnemonicToSeed` keeps the 2048-round PBKDF2 off the JS thread). `HMAC-SHA256(key = bip39 64-byte seed, msg = "freighter-auth-v1")` → `Keypair.fromRawEd25519Seed` → lowercase-hex `userId`. Uses `bip39` (direct dep on purpose — extension parity + its async seed/validate APIs) + the `crypto` import (react-native-crypto on device, Node's built-in under jest) + `@stellar/stellar-sdk` `Keypair` (signs via `@noble/curves`, unaffected by the `tweetnacl` jest mock). `validateMnemonic` first; rejects on invalid input.
- `src/services/auth/authKeypairVectors.ts` (new) — the **same** `mnemonic → authSeedHex → userId` vectors the extension committed, asserted in tests to guarantee cross-platform parity.
- `src/ducks/auth.ts` — exports `getActiveMnemonicPhrase(): Promise<string | null>` (thin wrapper over the existing private `getTemporaryStore`, gated on AUTHENTICATED with a re-check after the decrypt await); calls `clearAuthKeypairCache()` on both logout paths, in `clearAllData()`, in `softLock()`, and on the `HASH_KEY_EXPIRED` transitions in `getAuthStatus()` — so no session-ending path leaves key material resident.
- `src/services/auth/getAuthKeypair.ts` (new) — `getAuthKeypair(): Promise<Keypair | null>`, total (never rejects: errors are logged and surfaced as `null`), with an in-memory, module-scoped cache (never persisted) plus in-flight-promise memoization so the unlock request burst runs one derivation; `clearAuthKeypairCache(): void`. `selectAccount` intentionally does not clear — sub-accounts share one mnemonic and derivation is mnemonic-keyed, so the cached keypair stays correct across sub-account switches; only session-ending paths (logout/lock/soft-lock/hard-expiry/import) clear it.
- `package.json` — adds `bip39` pinned exact (`3.1.0`).

**Why this mirrors the extension:** HMAC-SHA256 and Ed25519 (RFC 8032) are standardized, so the RN crypto libraries produce byte-identical output for the same mnemonic. The shared vectors pin that output under jest (Node's `crypto`); the on-device react-native-crypto path is asserted by PR #865's gated E2E.

**Verification (post-review):** `deriveAuthKeypair` 7/7, `getAuthKeypair` 7/7, `ducks/auth` 137/137; `tsc --noEmit` clean; `madge` no circular deps; full pre-commit suite green on commit. Reviewed task-by-task (spec + quality) including the import-cycle, cache-eviction, and async-derivation paths.

**Follow-ups / out of scope:** PR 2 (#865) — `buildAuthJwt` + axios interceptor + gated E2E. Contacts UI and the backend permissive→required flip are separate.

</details>

### DIFF
--- a/__tests__/services/auth/getAuthKeypair.test.ts
+++ b/__tests__/services/auth/getAuthKeypair.test.ts
@@ -1,3 +1,4 @@
+import { AUTH_STATUS } from "config/types";
 import * as authDuck from "ducks/auth";
 import { AUTH_KEYPAIR_VECTORS } from "services/auth/authKeypairVectors";
 import {
@@ -5,14 +6,28 @@ import {
   clearAuthKeypairCache,
 } from "services/auth/getAuthKeypair";
 
-jest.mock("ducks/auth", () => ({ getActiveMnemonicPhrase: jest.fn() }));
+// Mock ducks/auth: expose both getActiveMnemonicPhrase and useAuthenticationStore
+// with a controllable getState so we can drive the auth-status gate directly.
+let mockAuthStatus: string = AUTH_STATUS.AUTHENTICATED;
+
+jest.mock("ducks/auth", () => ({
+  getActiveMnemonicPhrase: jest.fn(),
+  useAuthenticationStore: {
+    getState: jest.fn(() => ({ authStatus: mockAuthStatus })),
+  },
+}));
+
 const mockMnemonic = authDuck.getActiveMnemonicPhrase as jest.Mock;
+const mockGetState = authDuck.useAuthenticationStore.getState as jest.Mock;
 const M = AUTH_KEYPAIR_VECTORS[0];
 
 describe("getAuthKeypair", () => {
   beforeEach(() => {
     clearAuthKeypairCache();
     jest.clearAllMocks();
+    // Default: fully authenticated, so the status gate passes.
+    mockAuthStatus = AUTH_STATUS.AUTHENTICATED;
+    mockGetState.mockImplementation(() => ({ authStatus: mockAuthStatus }));
   });
 
   it("derives and memoizes (accessor called once for repeated calls)", async () => {
@@ -25,7 +40,52 @@ describe("getAuthKeypair", () => {
     expect(mockMnemonic).toHaveBeenCalledTimes(1);
   });
 
-  it("returns null when locked (no mnemonic)", async () => {
+  it("returns null when LOCKED even if getActiveMnemonicPhrase would return a mnemonic (regression guard)", async () => {
+    // This is the real LOCKED path: fast-unlock preserved session.
+    // getActiveMnemonicPhrase deliberately allows LOCKED, so without the
+    // auth-status gate it would still yield a keypair — this test proves it
+    // no longer does.
+    mockAuthStatus = AUTH_STATUS.LOCKED;
+    mockGetState.mockImplementation(() => ({ authStatus: mockAuthStatus }));
+    mockMnemonic.mockResolvedValue(M.mnemonic); // would succeed if gate absent
+
+    expect(await getAuthKeypair()).toBeNull();
+    // The mnemonic accessor must NOT have been called — gate fires first.
+    expect(mockMnemonic).not.toHaveBeenCalled();
+  });
+
+  it("returns null when HASH_KEY_EXPIRED", async () => {
+    mockAuthStatus = AUTH_STATUS.HASH_KEY_EXPIRED;
+    mockGetState.mockImplementation(() => ({ authStatus: mockAuthStatus }));
+    mockMnemonic.mockResolvedValue(M.mnemonic);
+
+    expect(await getAuthKeypair()).toBeNull();
+    expect(mockMnemonic).not.toHaveBeenCalled();
+  });
+
+  it("returns null when NOT_AUTHENTICATED", async () => {
+    mockAuthStatus = AUTH_STATUS.NOT_AUTHENTICATED;
+    mockGetState.mockImplementation(() => ({ authStatus: mockAuthStatus }));
+    mockMnemonic.mockResolvedValue(M.mnemonic);
+
+    expect(await getAuthKeypair()).toBeNull();
+    expect(mockMnemonic).not.toHaveBeenCalled();
+  });
+
+  it("warm cache is NOT returned after transitioning to LOCKED", async () => {
+    // Populate cache while AUTHENTICATED.
+    mockMnemonic.mockResolvedValue(M.mnemonic);
+    const a = await getAuthKeypair();
+    expect(a).not.toBeNull();
+
+    // Transition to LOCKED — warm cache must be bypassed.
+    mockAuthStatus = AUTH_STATUS.LOCKED;
+    mockGetState.mockImplementation(() => ({ authStatus: mockAuthStatus }));
+
+    expect(await getAuthKeypair()).toBeNull();
+  });
+
+  it("returns null when AUTHENTICATED but no mnemonic available", async () => {
     mockMnemonic.mockResolvedValue(null);
     expect(await getAuthKeypair()).toBeNull();
   });
--- a/__tests__/services/auth/getAuthKeypair.test.ts
+++ b/__tests__/services/auth/getAuthKeypair.test.ts
@@ -1,9 +1,7 @@
 import * as authDuck from "ducks/auth";
+import { clearAuthKeypairCache } from "services/auth/authKeypairCache";
 import { AUTH_KEYPAIR_VECTORS } from "services/auth/authKeypairVectors";
-import {
-  getAuthKeypair,
-  clearAuthKeypairCache,
-} from "services/auth/getAuthKeypair";
+import { getAuthKeypair } from "services/auth/getAuthKeypair";
 
 // Mock ducks/auth: expose both getActiveMnemonicPhrase and isSessionAuthValid
 // so we can drive the auth-status gate directly.
@@ -34,30 +32,12 @@ describe("getAuthKeypair", () => {
     expect(mockMnemonic).toHaveBeenCalledTimes(1);
   });
 
-  it("returns null when LOCKED even if getActiveMnemonicPhrase would return a mnemonic (regression guard)", async () => {
-    // This is the real LOCKED path: fast-unlock preserved session.
-    // getActiveMnemonicPhrase now also returns null when not AUTHENTICATED
-    // (Fix 2), but the gate in getAuthKeypair fires first — this test proves
-    // the accessor is never reached when LOCKED.
+  it("returns null when the session is invalid, before touching the mnemonic accessor", async () => {
+    // isSessionAuthValid folds LOCKED / HASH_KEY_EXPIRED / NOT_AUTHENTICATED
+    // into one boolean (the state→boolean mapping is covered in the duck tests).
+    // The gate must fire before getActiveMnemonicPhrase is ever reached.
     mockIsSessionAuthValid.mockResolvedValue(false);
     mockMnemonic.mockResolvedValue(M.mnemonic); // would succeed if gate absent
-
-    expect(await getAuthKeypair()).toBeNull();
-    // The mnemonic accessor must NOT have been called — gate fires first.
-    expect(mockMnemonic).not.toHaveBeenCalled();
-  });
-
-  it("returns null when HASH_KEY_EXPIRED", async () => {
-    mockIsSessionAuthValid.mockResolvedValue(false);
-    mockMnemonic.mockResolvedValue(M.mnemonic);
-
-    expect(await getAuthKeypair()).toBeNull();
-    expect(mockMnemonic).not.toHaveBeenCalled();
-  });
-
-  it("returns null when NOT_AUTHENTICATED", async () => {
-    mockIsSessionAuthValid.mockResolvedValue(false);
-    mockMnemonic.mockResolvedValue(M.mnemonic);
 
     expect(await getAuthKeypair()).toBeNull();
     expect(mockMnemonic).not.toHaveBeenCalled();
@@ -105,6 +85,21 @@ describe("getAuthKeypair", () => {
 
     expect(await getAuthKeypair()).toBeNull();
     // Mnemonic must not have been called on the second invocation.
+    expect(mockMnemonic).toHaveBeenCalledTimes(1);
+  });
+
+  it("dedupes concurrent first-derivations (accessor called once under a stampede)", async () => {
+    mockMnemonic.mockResolvedValue(M.mnemonic);
+    // Several callers race before the first derivation resolves (the
+    // balances/prices/history burst on unlock).
+    const [a, b, c] = await Promise.all([
+      getAuthKeypair(),
+      getAuthKeypair(),
+      getAuthKeypair(),
+    ]);
+    expect(a).not.toBeNull();
+    expect(b).toBe(a);
+    expect(c).toBe(a);
     expect(mockMnemonic).toHaveBeenCalledTimes(1);
   });
 });
--- a/__tests__/services/auth/getAuthKeypair.test.ts
+++ b/__tests__/services/auth/getAuthKeypair.test.ts
@@ -1,0 +1,44 @@
+import * as authDuck from "ducks/auth";
+import { AUTH_KEYPAIR_VECTORS } from "services/auth/authKeypairVectors";
+import {
+  getAuthKeypair,
+  clearAuthKeypairCache,
+} from "services/auth/getAuthKeypair";
+
+jest.mock("ducks/auth", () => ({ getActiveMnemonicPhrase: jest.fn() }));
+const mockMnemonic = authDuck.getActiveMnemonicPhrase as jest.Mock;
+const M = AUTH_KEYPAIR_VECTORS[0];
+
+describe("getAuthKeypair", () => {
+  beforeEach(() => {
+    clearAuthKeypairCache();
+    jest.clearAllMocks();
+  });
+
+  it("derives and memoizes (accessor called once for repeated calls)", async () => {
+    mockMnemonic.mockResolvedValue(M.mnemonic);
+    const a = await getAuthKeypair();
+    const b = await getAuthKeypair();
+    expect(a).not.toBeNull();
+    expect(a!.rawPublicKey().toString("hex")).toBe(M.userId);
+    expect(b).toBe(a); // same cached instance
+    expect(mockMnemonic).toHaveBeenCalledTimes(1);
+  });
+
+  it("returns null when locked (no mnemonic)", async () => {
+    mockMnemonic.mockResolvedValue(null);
+    expect(await getAuthKeypair()).toBeNull();
+  });
+
+  it("re-derives after cache clear, and on mnemonic change", async () => {
+    mockMnemonic.mockResolvedValue(M.mnemonic);
+    const a = await getAuthKeypair();
+    clearAuthKeypairCache();
+    mockMnemonic.mockResolvedValue(AUTH_KEYPAIR_VECTORS[1].mnemonic);
+    const c = await getAuthKeypair();
+    expect(c!.rawPublicKey().toString("hex")).toBe(
+      AUTH_KEYPAIR_VECTORS[1].userId,
+    );
+    expect(c).not.toBe(a);
+  });
+});
--- a/__tests__/services/auth/getAuthKeypair.test.ts
+++ b/__tests__/services/auth/getAuthKeypair.test.ts
@@ -1,4 +1,3 @@
-import { AUTH_STATUS } from "config/types";
 import * as authDuck from "ducks/auth";
 import { AUTH_KEYPAIR_VECTORS } from "services/auth/authKeypairVectors";
 import {
@@ -6,28 +5,23 @@ import {
   clearAuthKeypairCache,
 } from "services/auth/getAuthKeypair";
 
-// Mock ducks/auth: expose both getActiveMnemonicPhrase and useAuthenticationStore
-// with a controllable getState so we can drive the auth-status gate directly.
-let mockAuthStatus: string = AUTH_STATUS.AUTHENTICATED;
-
+// Mock ducks/auth: expose both getActiveMnemonicPhrase and isSessionAuthValid
+// so we can drive the auth-status gate directly.
 jest.mock("ducks/auth", () => ({
   getActiveMnemonicPhrase: jest.fn(),
-  useAuthenticationStore: {
-    getState: jest.fn(() => ({ authStatus: mockAuthStatus })),
-  },
+  isSessionAuthValid: jest.fn(),
 }));
 
 const mockMnemonic = authDuck.getActiveMnemonicPhrase as jest.Mock;
-const mockGetState = authDuck.useAuthenticationStore.getState as jest.Mock;
+const mockIsSessionAuthValid = authDuck.isSessionAuthValid as jest.Mock;
 const M = AUTH_KEYPAIR_VECTORS[0];
 
 describe("getAuthKeypair", () => {
   beforeEach(() => {
     clearAuthKeypairCache();
     jest.clearAllMocks();
-    // Default: fully authenticated, so the status gate passes.
-    mockAuthStatus = AUTH_STATUS.AUTHENTICATED;
-    mockGetState.mockImplementation(() => ({ authStatus: mockAuthStatus }));
+    // Default: session is fully valid.
+    mockIsSessionAuthValid.mockResolvedValue(true);
   });
 
   it("derives and memoizes (accessor called once for repeated calls)", async () => {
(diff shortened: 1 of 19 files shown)


piyalbasu:
**Extension parity:** this is the mobile port of stellar/freighter#2876 — *derive backend auth keypair from seed* (issue stellar/freighter#2769).

Same derivation (`HMAC-SHA256(bip39 seed, "freighter-auth-v1")` → Ed25519 `Keypair`) and the **same committed cross-platform test vectors** — the mobile-derived `userId` must match the extension's for the same mnemonic (asserted in `authKeypairVectors`).

src/services/auth/getAuthKeypair.ts
```diff
@@ -0,0 +1,36 @@
...
+export const getAuthKeypair = async (): Promise<Keypair | null> => {
+  // Only attach auth for a fully-unlocked (AUTHENTICATED) session. LOCKED
+  // (preserved fast-unlock session) and expired/unauthed states must not
+  // produce a keypair — and a warm cache must not be served in those states.
+  if (
+    useAuthenticationStore.getState().authStatus !== AUTH_STATUS.AUTHENTICATED
+  ) {
+    return null;
+  }
+  if (cache) return cache;
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 4d8e4810: the module-scoped cache moved to a new leaf module `services/auth/authKeypairCache.ts` that imports neither `ducks/auth` nor `getAuthKeypair`. `ducks/auth` now imports `clearAuthKeypairCache` from that leaf, removing the back-edge. `npx madge --circular` confirms: no circular dependency.

src/ducks/auth.ts
```diff
@@ -54,6 +54,7 @@ import { t } from "i18next";
 import ReactNativeBiometrics from "react-native-biometrics";
 import * as Keychain from "react-native-keychain";
 import { analytics } from "services/analytics";
+import { clearAuthKeypairCache } from "services/auth/getAuthKeypair";
 import { getAccount } from "services/stellar";
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 4d8e4810: `getActiveMnemonicPhrase` now returns `null` unless `authStatus === AUTH_STATUS.AUTHENTICATED` (explicit guard added before `getTemporaryStore`), so it no longer returns the mnemonic during a LOCKED session. `getTemporaryStore` itself is unchanged (the unlock flow still relies on it allowing LOCKED). Doc updated to match.

src/ducks/auth.ts
```diff
@@ -2895,3 +2899,18 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => ({
...
+export const getActiveMnemonicPhrase = async (): Promise<string | null> => {
+  try {
+    const store = await getTemporaryStore(
+      useAuthenticationStore.getState().authStatus,
+    );
+    return store?.mnemonicPhrase ?? null;
+  } catch {
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
piyalbasu: Reworded in 4d8e4810. After the accessor was gated on AUTHENTICATED, the comment now describes the intended regression scenario (the accessor returns `null` when not AUTHENTICATED, and the `getAuthKeypair` gate fires first) rather than asserting it "allows LOCKED".

__tests__/services/auth/getAuthKeypair.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    expect(b).toBe(a); // same cached instance
+    expect(mockMnemonic).toHaveBeenCalledTimes(1);
+  });
+
+  it("returns null when LOCKED even if getActiveMnemonicPhrase would return a mnemonic (regression guard)", async () => {
+    // This is the real LOCKED path: fast-unlock preserved session.
+    // getActiveMnemonicPhrase deliberately allows LOCKED, so without the
+    // auth-status gate it would still yield a keypair — this test proves it
+    // no longer does.
+    mockAuthStatus = AUTH_STATUS.LOCKED;
```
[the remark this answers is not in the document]
piyalbasu: Addressed in 4d8e4810: added a cheap `isSessionAuthValid()` (checks `authStatus` + reads the hash key via `getHashKey` + `isHashKeyExpired` — no temp-store decrypt / no PBKDF2) and gated `getAuthKeypair` on it **before** returning the warm cache. This closes the stale-`authStatus` window you described: an expired hash key now yields `null` even when the store still reports AUTHENTICATED. A regression test simulates that window and asserts the warm cache is not served.

src/services/auth/getAuthKeypair.ts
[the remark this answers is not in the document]
piyalbasu: Fixed in 4d8e4810: added an explicit `AUTHENTICATED` guard in `getActiveMnemonicPhrase` before decrypting the temporary store, so the helper can no longer return the mnemonic while the app is locked.

src/ducks/auth.ts
```diff
@@ -2895,3 +2899,18 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => ({
...
+
+/**
+ * Returns the mnemonic phrase from the unlocked temporary store,
+ * or null when the wallet is locked or the store is unavailable.
+ */
+export const getActiveMnemonicPhrase = async (): Promise<string | null> => {
+  try {
+    const store = await getTemporaryStore(
+      useAuthenticationStore.getState().authStatus,
+    );
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 4d8e4810: `softLock()` now calls `clearAuthKeypairCache()` alongside the account/session clear, so the idle/background auto-lock path evicts the derived signing key as the cache contract requires.

src/ducks/auth.ts
```diff
@@ -2151,6 +2153,7 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => ({
           if (hasAccountList && !shouldWipeAllData) {
             // Don't expire hash key - preserve temporary store accessibility
             // Security comes from app being locked, not key expiration
+            clearAuthKeypairCache();
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 4d8e4810: `softLock()` now calls `clearAuthKeypairCache()` alongside the account/session clear, so the idle/background auto-lock path evicts the derived signing key as the cache contract requires.

src/ducks/auth.ts
```diff
@@ -3106,3 +3111,40 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => ({
...
+  if (
+    useAuthenticationStore.getState().authStatus !== AUTH_STATUS.AUTHENTICATED
+  ) {
+    return null;
+  }
+  try {
+    const store = await getTemporaryStore(
+      useAuthenticationStore.getState().authStatus,
+    );
+    return store?.mnemonicPhrase ?? null;
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 994bb735: `getActiveMnemonicPhrase` now re-checks `authStatus === AUTHENTICATED` **after** the `await getTemporaryStore(...)` decrypt and returns `null` if a lock landed during the await — so the pre-lock mnemonic is never returned and `getAuthKeypair` can no longer re-cache a signing key after `softLock()` cleared it. Added a TOCTOU regression test that flips the store to LOCKED mid-decrypt and asserts `null`, plus a happy-path test proving the mnemonic is returned when the session stays AUTHENTICATED throughout.

src/services/auth/getAuthKeypair.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getActiveMnemonicPhrase, isSessionAuthValid } from "ducks/auth";
+import {
+  getCachedAuthKeypair,
+  setCachedAuthKeypair,
+} from "services/auth/authKeypairCache";
+import { deriveAuthKeypair } from "services/auth/deriveAuthKeypair";
+
+export { clearAuthKeypairCache } from "services/auth/authKeypairCache";
+
+export const getAuthKeypair = async (): Promise<Keypair | null> => {
```
aristidesstaffieri, line 21: This can reject despite the `Promise<Keypair | null>` contract: `isSessionAuthValid` → `getHashKey` throws on a keychain read failure or on `JSON.parse` of a corrupted HASH_KEY value, and `deriveAuthKeypair` throws when `validateMnemonic` rejects. Every other `getHashKey` consumer (`getAuthStatus`, `getTemporaryStore`) wraps the read in try/catch — this is the only unguarded path, so a caller in the request layer that treats the accessor as total gets an unhandled rejection instead of a graceful unauthenticated fallback. Suggest wrapping the body in a try/catch that logs a breadcrumb and returns null.

src/services/auth/deriveAuthKeypair.ts
```diff
@@ -0,0 +1,27 @@
...
+import { createHmac } from "crypto";
+
+export const AUTH_SALT = "freighter-auth-v1";
+
+/** 32-byte auth seed: HMAC-SHA256(key = 64-byte BIP39 seed, msg = AUTH_SALT). */
+export const deriveAuthSeed = (mnemonic: string): Buffer => {
+  if (!validateMnemonic(mnemonic)) {
+    throw new Error("Invalid mnemonic (see bip39)");
+  }
+  const seed = mnemonicToSeedSync(mnemonic); // 64 bytes, empty passphrase
```
aristidesstaffieri: `mnemonicToSeedSync` runs PBKDF2-HMAC-SHA512 (2048 iterations) in pure JS on the JS thread. The only caller is already async, so bip39's async `mnemonicToSeed` is a byte-identical drop-in. Related: `getAuthKeypair` caches only the finished keypair, so several concurrent first requests after unlock (balances + prices + history) each run a full derivation — memoizing the in-flight promise fixes the multiplier.

src/services/auth/deriveAuthKeypair.ts
```diff
@@ -0,0 +1,27 @@
+// alias → react-native-crypto (pure JS; works in jest & RN)
+
+import { Keypair } from "@stellar/stellar-sdk";
+import { mnemonicToSeedSync, validateMnemonic } from "bip39";
+import { createHmac } from "crypto";
```
aristidesstaffieri, line 14: The `crypto` import resolves to `react-native-crypto` only via package.json's `browser`/`react-native` field maps, which metro honors but Jest doesn't — jest.config.js has no `moduleNameMapper` entry for `crypto`, so under Jest this is Node's built-in crypto. The cross-platform vectors therefore validate Node's `createHmac`, not the RN runtime path: if `react-native-crypto` ever diverges on device (e.g. Buffer-key handling under buffer@4.9.2), users derive wrong userIds/keypairs while CI stays green. The line-1 comment ("works in jest & RN") overstates what the tests prove. Suggest a Jest mapping of `^crypto$` → `react-native-crypto`, or a device-side/E2E assertion of one vector.

src/services/auth/getAuthKeypair.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (!(await isSessionAuthValid())) return null; // AUTHENTICATED + hash key present & not expired (cheap)
```
aristidesstaffieri: On the cold path this gate runs twice: `isSessionAuthValid` reads the hash key + checks expiry, then `getActiveMnemonicPhrase → getTemporaryStore` re-reads the same hash key and re-runs the same AUTHENTICATED + expiry checks. The cheap pre-gate is only needed to protect the warm-cache return — e.g. `const cached = getCachedAuthKeypair(); if (cached) return (await isSessionAuthValid()) ? cached : null;` and let `getActiveMnemonicPhrase`'s internal gating cover the derive path. Saves a keychain round-trip per cold call and leaves one authoritative gate. (Take or leave — the current layering also reads as intentional defense-in-depth.)

src/services/auth/getAuthKeypair.ts
```diff
@@ -0,0 +1,20 @@
+import { Keypair } from "@stellar/stellar-sdk";
+import { getActiveMnemonicPhrase, isSessionAuthValid } from "ducks/auth";
+import {
+  getCachedAuthKeypair,
+  setCachedAuthKeypair,
+} from "services/auth/authKeypairCache";
+import { deriveAuthKeypair } from "services/auth/deriveAuthKeypair";
+
+export { clearAuthKeypairCache } from "services/auth/authKeypairCache";
```
aristidesstaffieri: `clearAuthKeypairCache` now has two import paths — this re-export, and the direct `services/auth/authKeypairCache` import in ducks/auth.ts. The tests already import from here while production imports from the cache module. Suggest dropping the re-export so there's one canonical path.

__tests__/services/auth/getAuthKeypair.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    // the accessor is never reached when LOCKED.
+    mockIsSessionAuthValid.mockResolvedValue(false);
+    mockMnemonic.mockResolvedValue(M.mnemonic); // would succeed if gate absent
+
+    expect(await getAuthKeypair()).toBeNull();
+    // The mnemonic accessor must NOT have been called — gate fires first.
+    expect(mockMnemonic).not.toHaveBeenCalled();
+  });
+
+  it("returns null when HASH_KEY_EXPIRED", async () => {
```
aristidesstaffieri: This test and "returns null when NOT_AUTHENTICATED" below are byte-identical to the LOCKED test above — all three set `mockIsSessionAuthValid.mockResolvedValue(false)` and assert the same thing, so they exercise the boolean gate once, not three states (the state→boolean mapping is covered in the duck tests). A single "returns null when session invalid" case says the same with a third of the code.

package.json
```diff
@@ -96,6 +96,7 @@
     "asyncstorage-down": "4.2.0",
     "axios": "1.13.2",
     "bignumber.js": "9.1.2",
+    "bip39": "3.1.0",
```
aristidesstaffieri, line 99: This adds bip39 as a direct dependency while stellar-hd-wallet already pins the same bip39 3.1.0 transitively, and `StellarHDWallet.fromMnemonic(...).seedHex` computes the same seed with the identical validation error. If the direct import is deliberate (extension parity — and bip39 is also where the async `mnemonicToSeed` lives if the sync-PBKDF2 comment is taken), a one-line comment saying so would prevent future "why two BIP39 stacks" questions; otherwise reusing stellar-hd-wallet avoids keeping the two pins in lockstep.

src/services/auth/authKeypairVectors.ts
```diff
@@ -0,0 +1,24 @@
+// Canonical cross-platform auth-keypair vectors — MUST match the extension's
```
aristidesstaffieri, line 1: Test-only fixture under `src/` — its only importers are the two new test files, and it's well-known mnemonics sitting next to real derivation code. Repo convention keeps fixtures under `__tests__/`; moving it there keeps the cross-platform parity contract without shipping it as app source. If `src/` placement deliberately mirrors the extension's shared-helpers layout, a header note would settle it.

src/ducks/auth.ts
```diff
@@ -3106,3 +3111,49 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => ({
...
+    // landed while awaiting SecureStorage. If so, do NOT return the pre-lock
+    // mnemonic — otherwise getAuthKeypair could derive and re-cache a signing
+    // key after softLock() already cleared the cache (TOCTOU race).
+    if (
+      useAuthenticationStore.getState().authStatus !== AUTH_STATUS.AUTHENTICATED
+    ) {
+      return null;
+    }
+    return store?.mnemonicPhrase ?? null;
+  } catch {
```
aristidesstaffieri: This catch is unreachable today — `getTemporaryStore` wraps its entire body in its own try/catch and returns null — and if a future edit adds a throwing call between the awaits, the bare `catch { return null; }` swallows it with no logging: an auth-keypair outage would look like a normal locked session with nothing in Sentry. Either drop it or log before returning null.

aristidesstaffieri:
One eviction gap in the new keypair cache: it's cleared on `clearAllData`, both `logout` paths, and `lock` — but not on the `AUTHENTICATED → HASH_KEY_EXPIRED` transition in `getAuthStatus` (`src/ducks/auth.ts:2827`, `set({ authStatus })` then navigate to lock). A session that hard-expires while the app is in memory leaves the derived Ed25519 private key in the JS heap indefinitely while the device sits on the re-auth screen — the `isSessionAuthValid` gate blocks *use* of the cache but not *retention*, which is the same rationale the softLock clear cites. Suggest clearing the cache where `HASH_KEY_EXPIRED` is set. (Top-level comment because that line predates this PR and isn't in the diff — the gap is in the PR's eviction coverage.)

src/services/auth/getAuthKeypair.ts
piyalbasu: Fixed in 02693b1c: the whole body is wrapped in try/catch, so the accessor is total. `isSessionAuthValid`→`getHashKey` throwing (keychain read failure or `JSON.parse` of a corrupt HASH_KEY) and `deriveAuthKeypair` rejecting on `validateMnemonic` are now logged via `logger.error` and surfaced as `null` — the request layer gets a graceful unauthenticated fallback instead of an unhandled rejection.

src/services/auth/deriveAuthKeypair.ts
piyalbasu: Fixed in 02693b1c on both counts:
- `deriveAuthSeed`/`deriveAuthKeypair` now use bip39's async `mnemonicToSeed` (byte-identical output, off the JS thread).
- `getAuthKeypair` memoizes the in-flight derivation promise, so the unlock burst (balances + prices + history) runs one PBKDF2 rather than one per caller.

Added a concurrency regression test asserting the mnemonic accessor is hit exactly once under a 3-caller stampede.

src/services/auth/deriveAuthKeypair.ts
piyalbasu: Corrected the overstated comment in 02693b1c — it now states plainly that `crypto` resolves to Node's built-in under Jest (no `moduleNameMapper` override) and to react-native-crypto only on device, so the vectors pin the HMAC-SHA256 *algorithm*, not the RN code path.

I did **not** add the `^crypto$`→react-native-crypto Jest mapping: it's a global remap that reroutes every `crypto` consumer in the suite (SDK internals included), and react-native-crypto pulls in the `react-native-randombytes` native module, which won't load under Jest. The honest coverage for the on-device library is a device-side vector assertion, which belongs in #865's gated E2E — I'll add it there. Leaving this open until that lands; say the word if you'd rather gate it differently.

src/services/auth/getAuthKeypair.ts
piyalbasu: Kept the two-stage gate deliberately. `isSessionAuthValid` is the only guard on the warm-cache return (your adjacent thread), and the existing regression tests assert the mnemonic accessor is never reached when the session is invalid — folding the cold-path gate into `getActiveMnemonicPhrase` would let the accessor be called (returning null via its own gate) and weaken that guarantee. The cost is one extra `getHashKey` read on the cold, once-per-session derive path, which reads as acceptable defense-in-depth for key material. Flagging it as a conscious keep — happy to revisit if you feel strongly. Leaving open for your call.

src/services/auth/getAuthKeypair.ts
piyalbasu: Fixed in 02693b1c: dropped the `clearAuthKeypairCache` re-export from `getAuthKeypair`. Both the duck and the test now import it directly from `services/auth/authKeypairCache` — one canonical path.

__tests__/services/auth/getAuthKeypair.test.ts
piyalbasu: Fixed in 02693b1c: collapsed the three into a single `returns null when the session is invalid, before touching the mnemonic accessor`. The state→boolean mapping is covered in the duck tests, so this just exercises the gate once.

package.json
piyalbasu: Documented in 02693b1c: added a comment at the import site stating bip39 is a deliberate direct dependency — extension parity (#2876 derives from bip39 directly) plus we use its async `mnemonicToSeed`/`validateMnemonic` — pinned to the same exact version `stellar-hd-wallet` resolves transitively. (package.json can't carry the note, so it lives on the import.)

src/services/auth/authKeypairVectors.ts
piyalbasu: The parity intent is already documented in the file header (lines 1–3: "MUST match the extension's committed values … SOURCE OF TRUTH: …"), which mirrors the extension's shared-helpers placement — that's the "header note settles it" path you mentioned. Happy to move it under `__tests__/` instead if you'd prefer that over the parity-mirroring layout; leaving open for your call.

src/ducks/auth.ts
piyalbasu: Fixed in 02693b1c: the catch now logs via `logger.error` before returning null (with a comment noting it's unreachable today but guards future throwing edits), so an auth-keypair outage surfaces in Sentry instead of masquerading as a normal locked session.

piyalbasu:
Addressed in 02693b1c: `getAuthStatus` now calls `clearAuthKeypairCache()` on both `HASH_KEY_EXPIRED` transitions (the soft-locked branch at 2820 and the general `set` at 2827), so a hard-expiry evicts the derived Ed25519 key rather than leaving it resident on the re-auth screen — closing the retention gap alongside the existing softLock/logout/wipe eviction. Added a duck regression test asserting eviction fires on the expiry transition.

src/services/auth/authKeypairVectors.ts
aristidesstaffieri: oh I see, I think its fine to leave it then

src/services/auth/getAuthKeypair.ts
aristidesstaffieri: got it yeah the cost seems minimal and I see why it exists now so I'm good here 👍

aristidesstaffieri (approved):

src/ducks/auth.ts
```diff
@@ -3106,3 +3111,49 @@ export const useAuthenticationStore = create<AuthStore>()((set, get) => ({
...
+ */
+export const isSessionAuthValid = async (): Promise<boolean> => {
+  if (
+    useAuthenticationStore.getState().authStatus !== AUTH_STATUS.AUTHENTICATED
+  ) {
+    return false;
+  }
+  const hashKey = await getHashKey();
+  if (!hashKey) return false;
+  return !isHashKeyExpired(hashKey);
```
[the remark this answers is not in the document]
piyalbasu: Fixed in e0b74af3: `isSessionAuthValid` now delegates to the authoritative `getAuthStatus()` funnel — `return (await getAuthStatus()) === AUTH_STATUS.AUTHENTICATED`. That funnel evaluates the auto-lock timer (`backgroundedAt` + `autoLockTimer`) alongside account existence, persisted `LOCKED`, and hash-key hard-expiry, so the stale foreground-before-funnel window you described now resolves to `LOCKED` and `getAuthKeypair` returns `null` instead of deriving/returning the signing key.

The previous check was a partial re-implementation that omitted exactly the auto-lock-timer branch — delegating removes that drift and keeps one source of truth. Added a regression test asserting `isSessionAuthValid` is `false` once the timer has elapsed even while the hash key is still within its TTL.

src/services/auth/deriveAuthKeypair.ts
piyalbasu: Follow-up on this one: I over-credited the E2E earlier. On closer look the gated `authJwt.integration` E2E also runs under **Node** crypto (jest environment), so it doesn't exercise `react-native-crypto` either — the on-device derivation/signing path is genuinely unasserted, exactly as you flagged here.

Filed #931 to track adding a real device-runtime assertion covering both the #864 derivation vector and #865 JWT signing (the two paths that share this crypto-fidelity gap). The RN-`URL` half of the fidelity concern was separately fixed in #925 (signed path no longer depends on any environment-specific primitive). Leaving this thread to #931.

PR #925, opened 2026-07-07
feat(auth): per-request backend JWT + apiFactory wiring (#865)

### LINKED ISSUE #865: [Mobile] Generate per-request backend JWT in apiFactory.ts
Generate a fresh Ed25519 JWT per request to `freighter-backend-v2` in `apiFactory.ts`. Sign with the auth private key.

Claims per the design doc:
- `sub` = hex auth pubkey (= user ID)
- `iss` = `"freighter-mobile"`
- `iat` = now
- `exp` = `iat` + 15s
- `bodyHash` = SHA256 of raw request body (empty-byte hash `e3b0c442...` for GET)
- `methodAndPath` = e.g., `"PUT /api/v1/contacts"`

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — Auth Flow + JWT claims table.

Depends on: #864 (auth keypair derivation), and freighter-backend-v2#88 (JWT middleware) for the end-to-end round-trip test.

**Acceptance:**
- Authenticated round-trip works end-to-end against `freighter-backend-v2` (200 on valid request, 401 on tampered body / wrong key / expired token).
- On 401: regenerate JWT and retry once before surfacing the error.
- JWT is generated fresh per request, not cached.

piyalbasu:
## TL;DR

Closes #865. Builds on #864: attaches a short-lived, per-request auth token to every Freighter Mobile call to `freighter-backend-v2`, so the backend can attribute requests to a user. When the wallet is unlocked the request carries the token; when locked it goes out anonymously (the backend is permissive today, so nothing breaks). On a 401 the request is retried once with a fresh token, and the second response (success or 401) surfaces to the caller — matching the extension's `authedFetch` (retry-once-then-return; no anonymous fallback). This mirrors the extension's per-request JWT work and produces the same token contract (only the issuer differs).

**Stacked on #920 (#864)** — review/merge that first. This PR's base is the `feat/864` branch.

<details>

**What changed:**

- `src/services/auth/buildAuthJwt.ts` (new) — pure, synchronous EdDSA (`alg:"EdDSA"`) compact JWS. Claims: `sub`=hex pubkey, `iss`="freighter-mobile", `iat`, `exp`=iat+15, `bodyHash`=hex SHA-256 of the raw body string (empty-input hash for GET), `methodAndPath`=`"METHOD /full/path?query"`. base64url, unpadded; the constant header segment is precomputed once at module load. Parity port of the extension's builder; only `iss` differs.
- `src/services/auth/attachAuth.ts` (new) — `attachAuthInterceptors(instance)`. **Request interceptor:** if `getAuthKeypair()` (from #864) returns a keypair, derive the signed path from `instance.getUri(config)` — axios's own `baseURL`-join + param serializer — by string-stripping its origin (NOT `new URL().pathname`: RN's built-in URL is non-WHATWG and would append a trailing `/` to query-less URLs → signed ≠ wire → 401). So the signed `methodAndPath` equals the wire request-target (incl. `/api/v1` + query) by construction and `config.params` is left for axios; serialize an object `config.data` to a string before hashing so `bodyHash` == the wire bytes; set `Authorization: Bearer …`. Locked → strip any stale `Authorization` and go anonymous. **Response interceptor:** on 401, retry once with a fresh token (`__isAuthRetry`) and return that response — success or the second 401. No anonymous fallback (matches the extension's `authedFetch`).
- `src/services/apiFactory.ts` — new optional `configureInstance` hook, invoked on the raw axios instance **before** the error-normalizer interceptor is registered. This is load-bearing: axios's normalizer converts errors to a plain `ApiError` (dropping `config`/`response`), so the auth interceptors must register first to see a real 401 and retry. Optional + no-op for other consumers (`freighterBackendV1` unaffected). Also dropped the now-unused `TWriteBody` generic — `post`/`put` bodies are typed `unknown`.
- `src/services/backend.ts` — `freighterBackendV2` created with `configureInstance: attachAuthInterceptors`. The two V2 **POST** call sites (`fetchTokenPrices`, `fetchCollectibles`) pass **plain object** bodies; the request interceptor serializes them centrally (before axios's `transformRequest`) so the hashed `bodyHash` equals the wire bytes. No per-call `JSON.stringify`, no `Content-Type` override (defaulted by `createApiService`), and no string-body contract (`createApiService<string>` / the `backendV2TypeContract` test are gone).
- `src/ducks/auth.ts` — short-TTL memo + in-flight dedup on `isSessionAuthValid`, which the interceptor calls on every request: the in-flight promise collapses the near-simultaneous unlock burst (balances + prices + history) to one funnel run, and a 1s value memo covers sequential calls; `signIn` clears the memo on unlock so the post-unlock burst isn't served a stale `false`. (The `isSessionAuthValid` → `getAuthStatus` funnel delegation itself lands in #920/#864.)
- `__tests__/services/auth/authJwt.integration.test.ts` (new) — gated E2E vs staging `whoami` (skipped unless `BACKEND_V2_URL`/`IS_INTEGRATION_MODE` set): valid→200+userId, tampered-body→401, flipped-sig→401, expired→401.

**Verification:** `tsc --noEmit` clean; auth + backend + api suites green (95 passed / 4 gated-e2e skipped), `ducks/auth` 75/75 incl. the `isSessionAuthValid` memo/in-flight tests. `attachAuth` covers the `getUri`-derived signing, central object-body serialization, and the retry-once-then-return 401 path through the full apiFactory chain (proving the retry re-signs with a fresh `iat` and that a second 401 surfaces without a third request). Reviewed task-by-task plus a whole-system (device-vs-jest) review that caught: the retry-ordering bug, the empty-`bodyHash` gap, and a device-only blocker where RN's non-spec `URL` trailing-slashed the signed path for query-less GETs (all fixed here).

**Known gap (tracked in #931):** `buildAuthJwt`'s `sha256` runs under Node `crypto` in jest but `react-native-crypto` on device, and even the gated E2E signs under Node — so no automated test asserts the on-device signing path (same class as #864). #931 tracks adding a device-runtime assertion for both.

**Follow-ups / out of scope:** Device-clock skew — a device clock off beyond the backend's tolerance (~+5s / −20s) 401s every authenticated V2 request while unlocked, same as the extension. Handling it — most likely widening the backend `ClockSkewLeeway` (5s today), else client `Date`-sync — is tracked in #930. Contacts UI/CRUD (separate ticket); the backend permissive→required flip (`freighter-backend-v2#114`), gated on extension + mobile adoption.

</details>


piyalbasu:
**Extension parity:** this ports two extension PRs:

- stellar/freighter#2877 — *per-request backend JWT + E2E round-trip* (issue stellar/freighter#2770): the `buildAuthJwt` builder and the authed-fetch interceptor/retry. The JWT contract is identical; only `iss` differs (`freighter-mobile` vs `freighter-extension`).
- stellar/freighter#2880 — *route all backend-v2 calls through the authed fetch chokepoint* (issue stellar/freighter#2879): the analogue of this PR's migration of the v2 POST call sites (`fetchTokenPrices`, `fetchCollectibles`) to the string-body/query-in-url contract so the signed `bodyHash`/`methodAndPath` match what's sent.

src/services/auth/attachAuth.ts
```diff
@@ -0,0 +1,210 @@
...
+  instance.interceptors.request.use(
+    async (config: InternalAxiosRequestConfig) => {
+      /* eslint-disable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires, global-require */
+      const { getAuthKeypair } =
+        require("services/auth/getAuthKeypair") as typeof import("services/auth/getAuthKeypair");
+      /* eslint-enable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires, global-require */
+      const keypair = await getAuthKeypair();
+      if (!keypair) {
+        // Wallet is locked — pass through without auth.
+        return config;
```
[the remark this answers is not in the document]
piyalbasu: Fixed in ff64a1a8: the request interceptor's locked/anonymous branch now strips any existing `Authorization` (and lowercase `authorization`) header before returning the config — so a 401-retry that copied the pre-lock Bearer token cannot re-send it after the wallet locks. Regression test added (anonymous request carrying a stale `Authorization` header → stripped, no JWT built).

src/services/auth/attachAuth.ts
```diff
@@ -0,0 +1,210 @@
...
+      }
+
+      /* eslint-disable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires, global-require */
+      const { buildAuthJwt } =
+        require("services/auth/buildAuthJwt") as typeof import("services/auth/buildAuthJwt");
+      /* eslint-enable @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires, global-require */
+
+      const serverPath = deriveServerPath(config.baseURL, config.url);
```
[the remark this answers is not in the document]
piyalbasu: Fixed in b0e1e3fb: authenticated requests now fold `config.params` into `config.url` before signing (then write the merged URL back and delete `config.params`), so the signed `methodAndPath` equals the wire request-target incl. the query. Tests assert the decoded JWT path and the outgoing URL match, plus the existing-query `?a=1&b=2` merge. Followed up in 0c5e5851 to also cover `URLSearchParams` params.

src/services/auth/attachAuth.ts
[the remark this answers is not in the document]
piyalbasu: Fixed in ff64a1a8: the request interceptor's locked/anonymous branch now strips any existing `Authorization` (and lowercase `authorization`) header before returning the config — so a 401-retry that copied the pre-lock Bearer token cannot re-send it after the wallet locks. Regression test added (anonymous request carrying a stale `Authorization` header → stripped, no JWT built).

src/services/auth/attachAuth.ts
```diff
@@ -0,0 +1,286 @@
...
+        body = config.data;
+      } else if (config.data !== undefined && config.data !== null) {
+        if (__DEV__) {
+          logger.warn(
+            "attachAuth",
+            "config.data is not a string — body will not be hashed. " +
+              "Pre-serialize the request body before passing it to axios.",
+            { method, path: serverPath, dataType: typeof config.data },
+          );
+        }
```
aristidesstaffieri: **Non-string bodies are silently signed with an empty bodyHash in production.** This warning is `__DEV__`-only, so a release build signs the empty-string hash while still sending the real body — a guaranteed double-401 with zero diagnostics. The compile-time guard (`TWriteBody`) only constrains the `data` argument of `post`/`put`; `config.data` is still reachable (e.g. `delete("/x", { data: {...} })`), and `FormData`/`Blob` bodies type-check fine. Worth either throwing here or logging unconditionally (not just in dev) so this failure mode is diagnosable in production.

src/services/backend.ts
```diff
@@ -408,9 +416,9 @@ export const fetchTokenPrices = async ({
           tokenId === NATIVE_TOKEN_CODE ? V2_NATIVE_PRICE_ID : tokenId,
         );
         ({ data } = await freighterBackendV2.post<TokenPricesResponse>(
-          "/token-prices",
-          { tokens: v2Tokens },
-          { params: { network: priceNetwork } },
+          `/token-prices?network=${priceNetwork}`,
+          JSON.stringify({ tokens: v2Tokens }),
```
aristidesstaffieri: **The string-only body contract pushes serialization into every call site — the interceptor could do it centrally.** Request interceptors run before axios's `transformRequest`, so a central `if (typeof config.data === "object" && config.data !== null) config.data = JSON.stringify(config.data)` in the auth interceptor would make the hashed string exactly what goes on the wire. Call sites could keep passing plain objects, and the `TWriteBody` generic, the `backendV2TypeContract` test file, the dev-only object-body warning, and the per-call `JSON.stringify` boilerplate would all become unnecessary — it also fixes the silent empty-bodyHash failure mode at the root.

src/services/backend.ts
```diff
@@ -408,9 +416,9 @@ export const fetchTokenPrices = async ({
...
+          { headers: { "Content-Type": "application/json" } },
```
aristidesstaffieri: Redundant: `createApiService` already defaults instance headers to `Content-Type: application/json` (see `apiFactory.ts`), so this per-call override is boilerplate (same at line 1214). If the instance default ever changes, these copies silently override it at exactly the signed call sites.

src/services/auth/attachAuth.ts
```diff
@@ -0,0 +1,286 @@
...
+        body = undefined;
```
aristidesstaffieri: Dead assignment — `body` is declared without an initializer at line 216, so it is already `undefined`; this branch only needs the dev warning.

src/services/auth/buildAuthJwt.ts
```diff
@@ -0,0 +1,45 @@
...
+    .digest("hex");
+  const payload = {
+    sub: keypair.rawPublicKey().toString("hex"),
+    iss: ISS,
+    iat,
+    exp: iat + JWT_LIFETIME_SECONDS,
+    bodyHash,
+    methodAndPath: `${method.toUpperCase()} ${path}`,
+  };
+  const signingInput = `${seg({ alg: "EdDSA", typ: "JWT" })}.${seg(payload)}`;
```
aristidesstaffieri: `seg({ alg: "EdDSA", typ: "JWT" })` is a constant but is re-computed (JSON.stringify + base64 + three regex replaces) on every authenticated request. Hoist it once to module level: `const HEADER_SEG = seg({ alg: "EdDSA", typ: "JWT" });`

__tests__/services/backend.test.ts
```diff
@@ -705,6 +711,46 @@ describe("Backend Service - fetchCollectibles severity split", () => {
...
+
+    const [url, body, config] = mockV2Post.mock.calls[0];
+    // URL must embed the network query param — not in config.params
+    expect(url).toContain("?network=");
+    expect(config).not.toHaveProperty("params");
+    // Body must be a pre-serialized string so bodyHash matches wire bytes
+    expect(typeof body).toBe("string");
+    expect(body).toBe(
+      JSON.stringify({ owner: params.owner, contracts: params.contracts }),
+    );
```
aristidesstaffieri: These individual assertions re-check exactly what the `toHaveBeenCalledWith` above already locks (url, stringified body, config shape) — every expectation is asserted twice, so any contract change must be edited in two places per test. Same pattern in the fetchTokenPrices test (line 797). The `toHaveBeenCalledWith` alone fully pins the contract.

src/services/auth/buildAuthJwt.ts
```diff
@@ -0,0 +1,45 @@
...
+}: BuildAuthJwtParams): string => {
+  const iat = Math.floor((now ?? Date.now()) / 1000);
+  const bodyHash = createHash("sha256")
+    .update(body ?? "")
+    .digest("hex");
+  const payload = {
+    sub: keypair.rawPublicKey().toString("hex"),
+    iss: ISS,
+    iat,
+    exp: iat + JWT_LIFETIME_SECONDS,
```
aristidesstaffieri, line 44: **Device-clock skew breaks all authenticated V2 requests.** `iat` comes straight from `Date.now()` with a 15s lifetime and no leeway or server-time sync. If the device clock is off by more than ~15s (common on Android with automatic date/time disabled), every token is already expired — or has a future `iat` — at validation time. The 401-retry in `attachAuth.ts` re-signs from the same skewed clock, so it fails identically, and since this PR removed the unauthenticated V2 path there is no anonymous fallback. Net effect: token prices, collectibles, and discover permanently fail for that user while the wallet is *unlocked* (but work while locked, which makes this confusing to debug from reports), and every V2 call doubles backend traffic via the futile retry.

Options: backdate `iat` (and/or extend `exp`) by a small leeway if the server tolerates it, lengthen the lifetime, sync against the server `Date` response header, or fall back to an anonymous request after the retry also 401s.

src/services/auth/attachAuth.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  let qs: string;
+  if (params instanceof URLSearchParams) {
+    qs = params.toString();
+  } else if (typeof params === "object") {
+    const entries = Object.entries(params as Record<string, unknown>);
+    if (entries.length === 0) return url;
+    qs = new URLSearchParams(
+      entries.map(([k, v]) => [k, String(v)] as [string, string]),
+    ).toString();
```
aristidesstaffieri: **Param serialization diverges from axios semantics — and only on authenticated requests.** `String(v)` turns `undefined`/`null`/arrays into literal `"undefined"`/`"null"`/`"a,b"`, whereas axios's default serializer drops null/undefined and repeats array keys. Since the locked-wallet branch returns early and lets axios serialize normally, the same call produces a different wire query depending on lock state. No current V2 call site passes `params`, but the next GET-with-params caller hits a state-dependent bug.

Axios exposes its exact URL-building logic as `instance.getUri(config)` (already used in `src/services/verified-token-lists/api.ts`) — that could replace both `mergeParamsIntoUrl` and the param handling here, guaranteeing the signed path always matches what axios sends.

src/services/auth/attachAuth.ts
```diff
@@ -0,0 +1,286 @@
...
+  if (params === undefined || params === null) return url;
+
+  let qs: string;
+  if (params instanceof URLSearchParams) {
+    qs = params.toString();
+  } else if (typeof params === "object") {
+    const entries = Object.entries(params as Record<string, unknown>);
+    if (entries.length === 0) return url;
+    qs = new URLSearchParams(
+      entries.map(([k, v]) => [k, String(v)] as [string, string]),
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 53fc76ad: `mergeParamsIntoUrl` now filters `null`/`undefined` entries before building the query, matching axios's default serializer (which omits them). The authenticated path no longer signs `?cursor=undefined` while the anonymous path omits it.

src/services/auth/attachAuth.ts
piyalbasu: Fixed at the root in 53fc76ad by serializing object bodies centrally in the request interceptor (your suggestion in the adjacent thread) — the non-string branch and its `__DEV__`-only warning are gone, so there is no empty-`bodyHash`-in-prod path left. `config.data` is serialized (before axios's `transformRequest`) so the hashed bytes equal the wire bytes.

src/services/backend.ts
piyalbasu: Adopted in 53fc76ad. The interceptor serializes an object `config.data` before axios's `transformRequest`, so the hashed bytes equal the wire bytes. Removed the `createApiService<string>` constraint on the V2 client, the `backendV2TypeContract` test, the dev warning, and the per-call `JSON.stringify`; call sites pass plain objects again.

One deviation: I kept the `TWriteBody` generic on `createApiService` itself (it now just defaults to `unknown`, so V2 accepts objects) rather than ripping the type param out of the shared factory. It's harmless dead-default now — happy to strip it entirely if you'd prefer the full removal.

src/services/backend.ts
piyalbasu: Fixed in 53fc76ad: dropped the per-call `Content-Type: application/json` overrides at both sites — `createApiService` already defaults it (apiFactory.ts:174). Removed as part of the central-serialization cleanup.

src/services/auth/attachAuth.ts
piyalbasu: Gone in 53fc76ad — the whole non-string branch (and its dead `body = undefined`) was removed by the central-serialization refactor.

src/services/auth/buildAuthJwt.ts
piyalbasu: Fixed in 53fc76ad: `const HEADER_SEG = seg({ alg: "EdDSA", typ: "JWT" })` is hoisted to module scope and computed once; the per-request path just interpolates it.

__tests__/services/backend.test.ts
piyalbasu: Fixed in 53fc76ad: collapsed both tests (fetchCollectibles + fetchTokenPrices) down to the single `toHaveBeenCalledWith` — dropped the redundant `const [url, body, config] = ...` re-checks. (The call shape also changed to a plain-object body now that serialization is central.)

src/services/auth/buildAuthJwt.ts
piyalbasu: Good catch — and checking freighter-backend-v2, the tolerance is actually *tighter* than 15s: `internal/auth/claims.go` has `ClockSkewLeeway = 5s` + `MaxTokenLifetime = 15s`, giving ~**+5s ahead / −20s behind**. So a client-side backdate/widen doesn't help — the server rejects `lifetime > 15s`, and a future `iat` beyond 5s.

53fc76ad instead adds a client-side **anonymous fallback**: if the signed 401-retry also 401s, the interceptor sends one anonymous request (these endpoints were anonymous pre-PR) and logs a `logger.warn`. That prevents the hard-break-while-unlocked and produces the fallback-rate telemetry. Proper skew handling — most likely widening the backend `ClockSkewLeeway`, else client `Date`-sync — is tracked in #930, to be decided from that data. Leaving this open until #930 lands.

src/services/auth/attachAuth.ts
piyalbasu: Addressed the concrete divergence you flagged: 53fc76ad filters `null`/`undefined` in `mergeParamsIntoUrl` so it matches axios (the `String(undefined)` → `"undefined"` case), removing the state-dependent query.

I did not switch to `instance.getUri()` yet. It's the cleaner end state, but it reworks the signed-path derivation — which is correctness-critical (signed path must equal the wire request-target), and no V2 call site passes `params` today — so I'd rather do it deliberately with its own tests than fold it into this PR. Happy to leave this open as the tracked cleanup, or take it as a fast-follow.

src/services/backend.ts
piyalbasu: Followed up in b77874f1: removed the `TWriteBody` generic entirely (not just the `<string>` instantiation) — `createApiService` no longer takes a type param and `post`/`put` bodies are typed `unknown`. tsc clean, suites green.

src/services/auth/attachAuth.ts
piyalbasu: Done in 0ab214b0 — switched to `instance.getUri()` as you suggested. Dropped both `mergeParamsIntoUrl` and `deriveServerPath` (and the fold-then-delete of `config.params`); the signed `methodAndPath` is now `pathname + search` of `instance.getUri(config)`, i.e. exactly what axios's own builder puts on the wire. So signed path == wire path by construction (and stays true across axios upgrades since both route through the same builder), and `config.params` is left untouched.

Confirmed against the installed axios (1.13.2): `getUri` does `buildFullPath(baseURL, url)` + `buildURL(params, serializer)`, so it includes the `/api/v1` prefix and uses the wire serializer — no hand-rolled parity needed. Serialization specifics (&-join, encoding) are now covered by the integration tests against real axios; the unit tests assert the getUri-derived path + params-preserved.

src/services/auth/buildAuthJwt.ts
piyalbasu: Update — reversed the anonymous-fallback decision (46d15cd6). On a closer look the extension's `authedFetch` retries once on 401 and returns the second 401 (no anonymous fallback), and matching that won out: the fallback broke parity, and during the permissive window it would mask a genuine auth failure behind an anonymous success. So #925 now does exactly retry-once-then-return, same as the extension.

That means clock skew is handled entirely at the backend, as you implied — tracked in #930, most likely widening `ClockSkewLeeway` (5s today) once 401/timing-rejection rates show it's worth it, else client `Date`-sync. Same lever the extension would use, so it fixes both clients. Leaving this open until #930 lands.

aristidesstaffieri (approved):

PR #951, opened 2026-07-26
fix(analytics): emit screen.viewed for the Scan/Receive tabs

CassioMG:
### What

`ScanReceiveScreen` now emits its per-tab screen views as `screen.viewed` (via `getScreenViewedProps()`, same as `BottomSheet`) instead of passing the `VIEW_*` slug to `track()` as an event name. Also adds the missing `VIEW_SCAN_QR_CODE` entry to `SCREEN_CATALOG`.

### Why

Every view of the Receive tab logged:

```
[Analytics] Screen-view "view_public_key_generator" passed to track() directly; emit SCREEN_VIEWED via the navigation choke points instead. Event dropped.
```

Sample: https://stellarorg.sentry.io/issues/7633220935/

#944 added the unified screen firing `track(VIEW_ACCOUNT_QR_CODE)`; #936 then made screen views a single `screen.viewed` event with a guard that drops screen-view slugs passed as event names. The screen was never migrated, so the Receive view was dropped entirely.

The Scan tab had the same bug silently: `VIEW_SCAN_QR_CODE` wasn't catalogued, so the guard couldn't see it and it emitted a bare `scan_qr_code` event.

Both tabs now emit `screen.viewed` with `{ screen_name, flow: "assets" }`. Tests previously passed through both bugs because they mock `track()`; they now assert the event name and props.

`Refs` the Sentry issue rather than `Fixes` — resolving manually after prod verification.

### DIFF
--- a/__tests__/components/screens/ScanReceiveScreen/ScanReceiveScreen.test.tsx
+++ b/__tests__/components/screens/ScanReceiveScreen/ScanReceiveScreen.test.tsx
@@ -114,6 +114,11 @@ describe("ScanReceiveScreen", () => {
       <ScanReceiveScreen {...makeProps("scan")} />,
     );
     fireEvent.press(getByText("Receive"));
+
+    // Assert the calls happened before inspecting them: if both tabs stopped
+    // resolving to catalogued screens the screen would track nothing at all,
+    // and a loop over an empty call list would pass vacuously.
+    expect(mockTrack).toHaveBeenCalledTimes(2);
     mockTrack.mock.calls.forEach(([event]) => {
       expect(event).toBe(AnalyticsEvent.SCREEN_VIEWED);
     });
--- a/__tests__/components/screens/ScanReceiveScreen/ScanReceiveScreen.test.tsx
+++ b/__tests__/components/screens/ScanReceiveScreen/ScanReceiveScreen.test.tsx
@@ -2,7 +2,7 @@ import { useIsFocused } from "@react-navigation/native";
 import { NativeStackScreenProps } from "@react-navigation/native-stack";
 import { act, fireEvent } from "@testing-library/react-native";
 import ScanReceiveScreen from "components/screens/ScanReceiveScreen";
-import { AnalyticsEvent } from "config/analyticsConfig";
+import { AnalyticsEvent, AnalyticsFlow } from "config/analyticsConfig";
 import { ROOT_NAVIGATOR_ROUTES, RootStackParamList } from "config/routes";
 import { renderWithProviders } from "helpers/testUtils";
 import React from "react";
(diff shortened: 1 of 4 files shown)


CassioMG:
The new vacuous-pass guard test can go green even when analytics is completely dead.

https://github.com/stellar/freighter-mobile/blob/e22fa48c881cb6b10667d477fc4e27a812d446b7/__tests__/components/screens/ScanReceiveScreen/ScanReceiveScreen.test.tsx#L111-L121

`mockTrack.mock.calls.forEach(...)` only constrains the shape of calls that *did* happen — it never asserts that any happened. If `getScreenViewedProps()` ever returns `null` for both tabs (e.g. a future refactor drops `VIEW_SCAN_QR_CODE` / `VIEW_ACCOUNT_QR_CODE` from `SCREEN_CATALOG`, which is exactly the regression this PR fixes), the `if (screenViewedProps)` gate in [`ScanReceiveScreen.tsx#L110-L112`](https://github.com/stellar/freighter-mobile/blob/e22fa48c881cb6b10667d477fc4e27a812d446b7/src/components/screens/ScanReceiveScreen/ScanReceiveScreen.tsx#L109-L113) skips `track()` entirely, `mock.calls` is empty, `forEach` iterates zero times, and the test passes.

Adding a call-count assertion makes the test fail on the drop it is meant to catch. This render mounts on `scan` and then switches to `receive`, so exactly two screen views are expected:

```js
it("never passes a VIEW_* slug to track() as an event name", () => {
  const { getByText } = renderWithProviders(
    <ScanReceiveScreen {...makeProps("scan")} />,
  );
  fireEvent.press(getByText("Receive"));

  // Both the initial scan view and the switch to Receive must have fired --
  // otherwise the loop below passes vacuously over an empty call list.
  expect(mockTrack).toHaveBeenCalledTimes(2);
  mockTrack.mock.calls.forEach(([event]) => {
    expect(event).toBe(AnalyticsEvent.SCREEN_VIEWED);
  });
});
```

`expect(mockTrack).toHaveBeenCalled()` is the minimal fix if you'd rather not pin the count.

The count of 2 is safe because `ScanReceiveScreen.tsx:111` is the only `track()` call site in the screen and neither tab view tracks on mount — `ScanTabView`'s scan analytics run through `useQRCodeScreenScanner` on an actual scan.

CassioMG:
Valid — the loop constrained the shape of calls that happened without asserting any happened, so dropping both catalog entries would have made it pass over an empty list.

Fixed in e0dc9bc2 with the count assertion. Verified both halves of the claim rather than assuming:

- Count of 2 is correct — passes as written.
- The test is genuinely non-vacuous now: removing both `VIEW_ACCOUNT_QR_CODE` and `VIEW_SCAN_QR_CODE` from `SCREEN_CATALOG` locally fails it with `Expected number of calls: 2 / Received number of calls: 0`. On the previous revision that same regression passed.

piyalbasu (approved):

PR #976, opened 2026-08-17
Add an inline CTA to the Collectibles empty state

CassioMG:
## What

The Collectibles tab had no CTA in its empty state and showed its floating pill unconditionally, so an empty collectibles tab rendered a floating "Add collectible" pill on top of an empty state that offered nothing — while the Tokens tab put its CTA inside the empty state and only floated a pill once funded.

**The Tokens tab now decides the Add button style for both tabs.**

| Tokens tab state | Tokens tab | Collectibles tab |
|---|---|---|
| Empty state (unfunded) | "Add XLM" in the empty state *(unchanged)* | **new** "Add collectible" in the empty state |
| Floating pill (funded) | "Add token" pill *(unchanged)* | "Add collectible" pill *(unchanged)* |

Since the tokens pill already appears exactly when `isFunded`, that single flag drives both tabs — the in-empty-state CTA and the pill are exact complements, so there is no combination of state where both are on screen.

## Scope

**Nothing changes on the Tokens tab.** `BalancesList` is untouched; its empty state keeps its own CTA and its pill keeps its `isFunded` gate. The only behavioral change is on the Collectibles tab: a new CTA in its empty state, and its pill now waits for the Tokens tab to switch over.

### Empty Collectibles state when account is unfunded (new)
Screenshot 2026-08-18 at 06 40 27

### Empty Collectibles state when account is funded
Screenshot 2026-08-18 at 06 42 29

### Navigating between all possible tab variations

https://github.com/user-attachments/assets/c3842eef-8a59-4c88-b8b3-79a8f556e1e8


piyalbasu:
### Bug: unfunded account + collectibles fetch error leaves the tab with no Add affordance

**Bug** — reproduced in the iOS simulator; regresses existing behavior on an error path. **Not a merge blocker**, but worth fixing in this PR since it breaks the invariant the PR itself introduces.

**TL;DR:** When the collectibles request fails for an account that isn't funded, the tab shows the error view with the Add pill hidden and the new empty-state button never appearing — so there is no way to add a collectible at all until a refetch succeeds. Before this PR, the Add pill stayed visible and enabled after a failed fetch.

**Steps to reproduce (confirmed):**
1. Make the collectibles request fail while balances stay healthy (e.g. block `POST /api/v1/collectibles` on the backend-v2 host with a proxy — the balances call is on the v1 host, so funding state resolves normally).
2. Switch to a never-funded account and open the collectibles tab.
3. The error view renders with no Add pill and no empty-state button. On `main` under the same conditions, the pill renders and is enabled.

---

<details>

**Root cause:** the pill's stand-down condition assumes the empty-state CTA will render, but doesn't track the CTA's actual mounting conditions.

The pill is suppressed whenever the account is unfunded and the grid has zero visible collectibles:

https://github.com/stellar/freighter-mobile/blob/2ee85972af68c95e9a24d685be80b249abda3c17/src/components/screens/HomeScreen/HomeScreen.tsx#L136-L146

But the CTA only mounts inside `renderEmptyView()`, and `CollectiblesGrid` pre-empts that view in two cases: the loading early-return at L137, and the error branch, where `ListEmptyComponent` renders the error view instead:

https://github.com/stellar/freighter-mobile/blob/2ee85972af68c95e9a24d685be80b249abda3c17/src/components/CollectiblesGrid.tsx#L226-L232

So for unfunded + empty grid + `error` set in the collectibles store: pill hidden AND CTA never mounted. This contradicts the comment guarding the pill render ("the tab is never left without either"):

https://github.com/stellar/freighter-mobile/blob/2ee85972af68c95e9a24d685be80b249abda3c17/src/components/screens/HomeScreen/HomeScreen.tsx#L440-L451

**Why the error case is a regression:** pre-PR the pill always rendered on the collectibles tab with `disabled={isCollectiblesLoading}`, and a failed fetch leaves `isCollectiblesLoading` false — so an enabled Add pill survived fetch errors. The loading sub-case of the same hole is cosmetic only (the pill was already disabled during load).

**Note on store semantics:** the error view keys off the store's `error` field, which is set only when the request itself throws (`ducks/collectibles.ts` catch). Backend per-collection error payloads are filtered/logged and do not set `error`, so they don't trigger this path.

**Deterministic repro:** render the collectibles tab with the store in `{collections: [], isLoading: false, error: "…"}` and `isFunded: false`. Currently untested — the HomeScreen test mock hardcodes `isLoading: false, error: null`, so neither pre-empting branch is exercised.

1. *Align the conditions:* include the grid's pre-empting states in the pill stand-down, e.g. `showCollectiblesPill = !(showEmptyStateCta && isCollectiblesGridEmpty && !isCollectiblesLoading && !collectiblesError)` — the pill then stands down exactly when the CTA actually renders, restoring pre-PR behavior on error.
2. *Single source of truth:* derive one `emptyStateCtaVisible` value (mirroring `shouldShowEmptyStateCta` plus the loading/error gates) and use it for both the grid CTA and the pill suppression, so the two can't drift again.
3. *Test:* add a HomeScreen case with the collectibles store in the error state for an unfunded account, asserting an Add affordance is present.

</details>

piyalbasu (approved):
Approved with 1 nit bug i found while testing

CassioMG:
Good find — fixed in 920c48d2, though not quite the way you suggested.

Your diagnosis is exactly right: the pill stands down on unfunded + empty grid, but the CTA only mounts inside `renderEmptyView()`, and the error branch replaces that view. I had closed the "grid has content" pre-emption in an earlier commit and missed that `error` and the loading spinner pre-empt it too.

**One problem with fix 1.** Adding `&& !isCollectiblesLoading` opens the inverse hole: `isRefreshing` is grid-local state HomeScreen can't see, so during a pull-to-refresh the store reports `isLoading: true` while the grid skips its spinner (`isLoading && !isRefreshing`) and **does** mount the CTA. The `!isCollectiblesLoading` term would flip the pill back on at the same time — trading a "neither" bug for a "both" bug. Fix 2 has the right instinct but runs into the same wall: HomeScreen can't mirror conditions that depend on state private to the grid.

**What I did instead:** made the CTA a shared element used by *both* the empty and the error view, so the branch that swallows the empty state no longer swallows its only action. That keeps the pill's condition free of loading state entirely, so the whole class goes away:

- error + unfunded + empty → CTA in the error view, pill suppressed → exactly one
- pull-to-refresh → CTA in the empty view, pill suppressed → exactly one
- initial spinner → neither, but transient, and pre-PR the pill was disabled there anyway

Also took fix 3. You were right that the mock hardcoded `isLoading: false, error: null` so neither pre-empting branch was exercised — both are now mutable, with a HomeScreen case for unfunded + failed collectibles fetch, and the grid's error-view test (which previously asserted the CTA was *absent*) inverted plus a FlatList-path variant.

PR #977, opened 2026-08-17
fix(wallets): size the Add wallet chip to match the list avatars

CassioMG:
### What

Mobile counterpart of stellar/freighter#2964, which made the same change on the extension.

On the wallets list bottom sheet, the purple "+" chip in the `Add wallet` footer was 34px while every row's identicon avatar is 40px. Sized the chip to 40px so the two circles match down the same column.

Mobile needed two things the extension didn't:

- The account rows were direct children of the sheet's `gap-[24px]` container, so they sat 24px apart instead of 16px.
- The `+` glyph needed a heavier stroke than `Icon.Plus` gives at this size — see Scope.

### Why

The two circled elements sit in one vertical column and read as different sizes. As a bonus, the footer button already uses a 16px gap, so matching the chip width also lines the `Add wallet` label up with each row's account name.

The 16px row gap is what the design calls for and matches the extension's `.Wallets__list`.

### Scope

Two files, three presentational changes. No behavior change, no new assets, and the SDS icon registry is untouched.

**1. `AddWalletFooter` — chip 34px → 40px.** The avatar size itself was already correct and is untouched: `Avatar size="lg"` is 40×40 in `components/sds/Avatar`.

**2. `ManageAccountBottomSheet` — the account list gets its own `gap-[16px]` View.** Wrapping rather than changing the parent gap preserves the surrounding 24px rhythm between the hero, quick actions, divider and the list as a block.

**3. The chip's glyph 14px → 16px, switched from `Icon.Plus` to `Icon.PlusDouble`.**

Screenshot 2026-08-17 at 17 06 53

### DIFF
--- a/__tests__/components/CollectiblesGrid.test.tsx
+++ b/__tests__/components/CollectiblesGrid.test.tsx
@@ -1,0 +1,159 @@
+import { fireEvent, render } from "@testing-library/react-native";
+import {
+  CollectibleFilterType,
+  CollectiblesGrid,
+} from "components/CollectiblesGrid";
+import React from "react";
+
+jest.mock("react-i18next", () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+  }),
+}));
+
+jest.mock("hooks/useAppTranslation", () => ({
+  __esModule: true,
+  default: () => ({
+    t: (key: string) => key,
+  }),
+}));
+
+const mockFetchCollectibles = jest.fn();
+
+// Mutable so each test can shape what the grid sees.
+let mockCollections: unknown[] = [];
+let mockIsLoading = false;
+let mockError: string | null = null;
+
+jest.mock("ducks/collectibles", () => ({
+  useCollectiblesStore: jest.fn((selector?: (s: unknown) => unknown) => {
+    const state = {
+      collections: mockCollections,
+      isLoading: mockIsLoading,
+      error: mockError,
+      fetchCollectibles: mockFetchCollectibles,
+    };
+    return selector ? selector(state) : state;
+  }),
+}));
+
+jest.mock("ducks/auth", () => ({
+  useAuthenticationStore: jest.fn(() => ({ network: "PUBLIC" })),
+}));
+
+jest.mock("hooks/useGetActiveAccount", () => ({
+  __esModule: true,
+  default: () => ({ account: { publicKey: "GTESTPUBLICKEY" } }),
+}));
+
+describe("CollectiblesGrid empty-state CTA", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockCollections = [];
+    mockIsLoading = false;
+    mockError = null;
+  });
+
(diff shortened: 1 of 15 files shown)


piyalbasu:
### Nit: empty account list now leaves a blank gap in the sheet

**Nit** — reachable only via a load race, cosmetic-only impact. **Not a merge blocker.**

**TL;DR:** If the manage-accounts sheet is opened while the account list is still empty (e.g. before it finishes loading), the new grouping added in this PR leaves a visible band of blank space between the divider and the bottom of the sheet, where previously the layout collapsed cleanly.

**Steps to reproduce:**
1. Trigger the manage-accounts sheet while accounts are still loading (there's no loading guard on the press handler, and the list starts out empty).
2. Observe extra vertical blank space between the divider and the scroll-end spacer for the duration of that frame.

---

<details>

**Root cause:** the new rows wrapper always renders, even when `accounts` is `[]`. In React Native an empty zero-height `View` still counts as a flex child for `gap` purposes, so the parent `gap-[24px]` column renders divider —24px— empty wrapper —24px— footer (~24px of extra blank space). Pre-PR the rows were direct siblings of the column, so an empty list contributed zero children and no extra gap.

https://github.com/stellar/freighter-mobile/blob/ef7fcf9d92ca99a8011e5775d20dcf00b3440bef/src/components/screens/HomeScreen/ManageAccountBottomSheet.tsx#L193-L212

Reachability: `handleManageAccountsPress()` has no guard against presenting the sheet while `isLoadingAccounts` is true, and `allAccounts` starts as `[]` in `ducks/auth.ts`. The sheet's own defensive `accounts.length === 0 && activeAccount && !isLoadingAccounts` toast acknowledges the empty state exists.

**Deterministic repro:** render `ManageAccountBottomSheet` with `accounts={[]}` and measure/snapshot — the wrapper `View` occupies a gap slot between the divider and the footer spacer.

1. *Cheap guard (sufficient):* conditionally render the wrapper — `{accounts.length > 0 && (<View className="w-full gap-[16px]">…</View>)}` — which restores the pre-PR collapse regardless of why the list is empty.
2. *Kill the race too (optional hardening):* additionally guard `handleManageAccountsPress()` in HomeScreen on `isLoadingAccounts`, though that changes button behavior during load and is a bigger call than this nit warrants.

</details>

piyalbasu (approved):
Approved with just 1 NIT comment

CassioMG:
Fixed in 01702b13 — took fix 1, skipped fix 2.

Confirmed the mechanism: the wrapper is new in this PR (the rows were direct children of the `gap-[24px]` column before), and an empty `View` is still a flex child, so an empty list laid out divider —24px— empty wrapper —24px— footer. Guarding it on `accounts.length > 0` restores the previous collapse regardless of *why* the list is empty.

Skipped the press-handler guard: it changes the button's behaviour during load, which is a bigger call than this cosmetic gap warrants — agreed with your own read there.

Also added a `manage-accounts-list` testID so the wrapper's absence is assertable, with tests for the empty and populated cases.

PR #984, opened 2026-08-18
feat(send): verify assembled collectible transfer matches intent before signing

JiahuiWho:
### What

- Add `verifyFlatTransferPreparedTransaction` that verifies the assembled transaction:
  - has exactly one `invokeHostFunction` operation whose host function is byte-identical to the one the wallet built,
  - has every other envelope field (source, sequence, memo, preconditions) unchanged,
  - does not raise the fee beyond the simulated resource fee shown to the user,
  - and has authorization entries that all use source-account credentials with no sub-invocations.

### Why

The collectible send flow builds a Soroban transfer with no authorization entries and delegates simulation and assembly to the backend. The authorization entries, fee, and Soroban resource data on the returned transaction are therefore derived from simulating the target contract, not built by the wallet.

The wallet is the only party that knows the call it intended, so it should confirm the assembled transaction still matches that intent before signing, rather than adopting the returned bytes verbatim.

### Known limitations

To fix this we can't just copy paste auth-entry guard onto /simulate-tx in the freighter-backend, due to Soroswap flow, which also calls this endpoint and legitimately produces sub-invocations.

#### PR structure

#### Testing

#### Release

### DIFF
--- a/__tests__/services/transactionService.test.ts
+++ b/__tests__/services/transactionService.test.ts
@@ -48,6 +48,11 @@ jest.mock("services/backend", () => ({
   checkContractSupportsMuxed: jest.fn().mockResolvedValue(false),
 }));
 
+// Prepared-transaction verification parses real XDR; these suites exercise
+// fee/plumbing behaviour with placeholder XDR, so stub it out here. The
+// verification itself is covered in verifyPreparedTransaction.test.ts.
+jest.mock("services/verifyPreparedTransaction");
+
 describe("buildSendCollectibleTransaction", () => {
   const mockSenderKeypair = Keypair.random();
   const mockRecipientKeypair = Keypair.random();
--- a/__tests__/services/verifyPreparedTransaction.test.ts
+++ b/__tests__/services/verifyPreparedTransaction.test.ts
@@ -103,6 +103,23 @@ const addressCredentialsAuth = () =>
     rootInvocation: rootTransferInvocation(),
   });
 
+// A source-account auth entry whose root invocation is an unrelated contract
+// call (a token transfer to the attacker) with no sub-invocations.
+const unrelatedRootAuth = () =>
+  new xdr.SorobanAuthorizationEntry({
+    credentials: xdr.SorobanCredentials.sorobanCredentialsSourceAccount(),
+    rootInvocation: authorizedInvocation(SAC, "transfer", [
+      new Address(SOURCE).toScVal(),
+      new Address(ATTACKER).toScVal(),
+      xdr.ScVal.scvI128(
+        new xdr.Int128Parts({
+          hi: xdr.Int64.fromString("0"),
+          lo: xdr.Uint64.fromString("99884944477"),
+        }),
+      ),
+    ]),
+  });
+
 // Rebuild the transaction the way an assembling backend does: reuse the invoked
 // host function verbatim (or tamper with it), set a fee, and attach auth entries.
 const buildPreparedXdr = (
@@ -111,7 +128,8 @@ const buildPreparedXdr = (
   {
     tamperFunc = false,
     fee = BASE_FEE,
-  }: { tamperFunc?: boolean; fee?: string } = {},
+    opSource,
+  }: { tamperFunc?: boolean; fee?: string; opSource?: string } = {},
 ): string => {
   const built = TransactionBuilder.fromXDR(builtXdr, NETWORK) as Transaction;
   const builtOp = built.operations[0];
@@ -130,7 +148,9 @@ const buildPreparedXdr = (
     : builtOp.func;
 
   return newBuilder(fee)
-    .addOperation(Operation.invokeHostFunction({ func, auth }))
+    .addOperation(
+      Operation.invokeHostFunction({ func, auth, source: opSource }),
+    )
     .setTimeout(0)
     .build()
     .toXDR();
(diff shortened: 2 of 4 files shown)


CassioMG:
### Missing simulated resource fee collapses the fee ceiling and rejects legitimate transfers

**Possible functional regression** — not a definite merge blocker, but worth confirming against real backend responses before merge, since the failure mode is a hard-failed send rather than a degraded one.

**TL;DR:** The simulated resource fee that the new check uses as its fee ceiling is an optional field in the backend response. When it comes back absent, the code substitutes zero, which sets the ceiling at exactly the fee the wallet originally built. Any genuinely assembled Soroban transaction has a fee above that by definition — that is what simulation is for — so a missing resource fee turns every collectible send into a verification failure rather than a send that simply cannot display a fee breakdown. Treating "unknown" as "zero" makes the ceiling stricter than the wallet's own intent instead of more permissive.

**Steps to reproduce:**
1. Make `/simulate-tx` return a response with a `preparedTransaction` but no simulated resource fee in the simulation response (the shape this repo's own tests already treat as valid).
2. Start a collectible send and let the flow reach simulation.
3. The send fails with "fee exceeds the built fee plus the simulated resource fee" instead of proceeding with no fee breakdown shown.

---

<details>

**Root cause:** `minResourceFee` is declared optional on `SorobanSimulationResponse`:

https://github.com/stellar/freighter-mobile/blob/d42a5fd9bbf2ebb71b8325af9c6286aa7b28c214/src/services/backend.ts#L865-L868

The new call site coerces that optional value to `"0"`:

https://github.com/stellar/freighter-mobile/blob/d42a5fd9bbf2ebb71b8325af9c6286aa7b28c214/src/services/transactionService.ts#L860-L866

which feeds the ceiling computation, making `maxFee === built.fee` whenever the field is absent:

https://github.com/stellar/freighter-mobile/blob/d42a5fd9bbf2ebb71b8325af9c6286aa7b28c214/src/services/verifyPreparedTransaction.ts#L155-L161

Because the SDK's `assembleTransaction` sets `fee = builtFee + resourceFee`, any real invocation with a non-zero resource cost satisfies `prepared.fee > built.fee` and therefore trips this branch. The error propagates out of `simulateCollectibleTransfer` (the `catch` re-throws after `trackSimulationError`) and both call sites in `src/ducks/transactionBuilder.ts` surface it as a build failure, so the send is blocked entirely.

**Why "absent" is a real state, not a hypothetical:** commit `97bd011cb` (#774) introduced `minResourceFee?: string` together with the `result.simulationResponse?.minResourceFee` optional-chaining pattern, and `372f63bb` ("improve resource fee validation") deliberately changed handling of a missing/NaN value to a safe fallback — `sorobanResourceFeeXlm = null`, i.e. "don't display a resource fee" rather than "the resource fee is zero". This PR's ceiling inverts that established convention for the same field.

**Deterministic repro:** the existing suite already encodes this response shape as a success case:

https://github.com/stellar/freighter-mobile/blob/d42a5fd9bbf2ebb71b8325af9c6286aa7b28c214/__tests__/services/transactionService.test.ts#L291-L295

That test passes today only because the verifier is auto-mocked at line 54 of the same file. Removing that `jest.mock` and giving `simulateTransaction` a `preparedTransaction` whose fee exceeds the built fee reproduces the rejection directly, with no backend needed.

1. *Cheap guard:* when `minResourceFee` is absent, skip the fee-ceiling check and document why — the other four checks (host function, envelope fields, auth credentials, sub-invocations) still hold, and they are the ones carrying the security value.
2. *Explicit failure:* keep failing closed, but with a distinct message such as "resource fee unknown; cannot verify fee ceiling", so the cause is diagnosable rather than presenting as an inflated-fee attack.
3. *Root cause:* make the parameter `maxResourceFee?: string` and force the caller to decide, rather than hiding the distinction between "zero" and "unknown" behind `?? "0"`. This also removes the redundant `|| "0"` inside the helper, which currently makes an absent value indistinguishable from a legitimately zero one.

Worth noting the same `?? "0"` reasoning applies if `simulationResponse` itself is ever absent — the type says it is required, but the call site's optional chaining suggests otherwise.

</details>

CassioMG:
### Auto-mocking the verifier leaves the new signing gate's wiring untested

**Suggestion** — not a merge blocker. The verification logic itself is well covered; what's uncovered is that it is actually still connected.

**TL;DR:** The new verification module is stubbed out wholesale in the transaction-service suite, which is a reasonable way to keep those tests on placeholder XDR. The side effect is that nothing asserts the check is still wired into the collectible send flow, or that a verification failure actually stops the send rather than being swallowed. If someone later removed the call, reordered it after the value is returned, or wrapped it in a `try/catch`, the whole suite would stay green and the gate would be silently disabled — the one failure mode a security check cannot afford, because it fails open and invisibly. Two assertions on the call site would close that gap.

---

<details>

**Root cause:** the module-level auto-mock replaces every export with a no-op:

https://github.com/stellar/freighter-mobile/blob/d42a5fd9bbf2ebb71b8325af9c6286aa7b28c214/__tests__/services/transactionService.test.ts#L50-L55

Every test in the `simulateCollectibleTransfer` block then asserts only on `backend.simulateTransaction` and the returned `{ preparedTransaction, minResourceFee }` — never on the verifier. The new call site is therefore exercised but unobserved:

https://github.com/stellar/freighter-mobile/blob/d42a5fd9bbf2ebb71b8325af9c6286aa7b28c214/src/services/transactionService.ts#L860-L868

Note the interaction with the sibling comment on this PR: the auto-mock is also what hides the missing-`minResourceFee` rejection, so these two gaps compound — the mock conceals both that the gate could vanish and that it currently over-rejects.

1. *Wiring assertion:* keep the auto-mock and add `expect(verifyFlatTransferPreparedTransaction).toHaveBeenCalledWith({ builtTransactionXdr: mockTransactionXdr, preparedTransactionXdr: mockPreparedXdr, networkPassphrase: mockNetworkDetails.networkPassphrase, maxResourceFee: ... })`. This pins the argument mapping, which is where an XDR mix-up (verifying the prepared transaction against itself, for instance) would otherwise hide undetected.
2. *Fail-closed assertion:* one test where the mock is given `mockImplementation(() => { throw new PreparedTransactionMismatchError("test"); })`, asserting `simulateCollectibleTransfer` rejects and returns no prepared transaction. This is the assertion that would catch a future `try/catch` swallowing the error.

**Prior context:** this "test the real call-site wiring, not just the isolated helper" point has been raised twice before on this same file, both times on money-moving paths:
- https://github.com/stellar/freighter-mobile/pull/774#discussion_r3141000630 — "this test only covers the single-mount case; the bug is in the multi-mount case actually used in the app."
- https://github.com/stellar/freighter-mobile/pull/821#discussion_r3120388809 — "For money-moving code that depends on an external step, regression risk is high and a single live E2E flow isn't sufficient coverage."

</details>

PR #986, opened 2026-08-20
feat(swap): default the You receive token to USDC

### LINKED ISSUE #940: [Swap] Default the "You receive" token to USDC
## Summary

In the Swap flow, the **You receive** field currently defaults to an empty state with a `Select` button (and a `+` icon on the left). We should default the receive token to **USDC** instead, so users have one less tap to complete a common swap.

## Current behavior

- The **You receive** field shows a `Select` button (empty state).
- A `+` icon is shown on the left of the field.

## Desired behavior

- Default the **You receive** token to **USDC**, except when starting the swap flow from the USDC Asset Details screen in which case the **You receive** token should default to the native **XLM**.
- Apply this default **regardless of whether the user already holds a USDC balance**.
- Users can still open the selector to choose a different receive token.
- If we keep an empty/select state anywhere, remove the `+` icon on the left (see [thread note](https://stellarfoundation.slack.com/archives/C03DK78B43C/p1783960898557799))

## Source

Based on Slack discussion: https://stellarfoundation.slack.com/archives/C03DK78B43C/p1783960898557799

JiahuiWho:
### What

- `config/constants.ts`: `DEFAULT_SWAP_DEST_TOKEN_ID` per-network map + Circle's testnet issuer
- `SwapAmountScreen.tsx`: 
    - Sets the default destination token once per mount, after balances for the active account/network have loaded; trustline status comes from those same balances, so existing consumers need no changes.
    - For a default token the user doesn't hold, attaches a real Blockaid scan result when it resolves, so mainnet users see a proper verdict at Review instead of "unable to scan". Failed scans or non-mainnet leave it unscanned.
- `AmountCard.tsx`: removed the `+` icon from the empty picker chip per the design note 

### Why

Closes [#940](https://github.com/stellar/freighter-mobile/issues/940). Mobile port of the swap flow's **You receive** field defaults to the network's USDC, or native XLM when the swap starts from USDC itself (token-details entry). 

### Known limitations

N/A

#### PR structure

#### Testing

#### Release

### Simulation
1. Swap xlm will have usdc as default dest with trustline adding notice

https://github.com/user-attachments/assets/d1b0b983-9594-4473-afed-f9f2cb8dd8ee

2.  Swap usdc will have xlm as default dest

https://github.com/user-attachments/assets/86138160-5c83-4770-b55f-51b57049e0ab

3. "+" icon is removed when displaying "Select Text"
plus icon is removed

4. In android

https://github.com/user-attachments/assets/60898efd-de38-4abc-838d-6c72f56d8dfd

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (3.0.9)
+    CFPropertyList (3.0.8)
     abbrev (0.1.2)
     activesupport (7.2.3)
       base64
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (3.0.8)
+    CFPropertyList (3.0.9)
     abbrev (0.1.2)
     activesupport (7.2.3)
       base64
--- a/__tests__/components/screens/SwapScreen/SwapAmountScreen.test.tsx
+++ b/__tests__/components/screens/SwapScreen/SwapAmountScreen.test.tsx
@@ -368,6 +368,113 @@ describe("SwapAmountScreen", () => {
     expect(mockSetDestinationToken).toHaveBeenCalledWith(null);
   });
 
+  describe("default destination (USDC) seeding", () => {
+    // The auth store defaults to PUBLIC in tests, so the seeded default is
+    // mainnet USDC (Circle issuer).
+    const MAINNET_USDC_ID =
+      "USDC:GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN";
+
+    it("seeds USDC as the destination when none is set", () => {
+      setSwapStoreState({ destinationToken: null });
+
+      renderWithProviders(
+        <SwapAmountScreen navigation={makeNavigation()} route={makeRoute()} />,
+      );
+
+      expect(mockSetDestinationToken).toHaveBeenCalledWith(
+        expect.objectContaining({
+          id: MAINNET_USDC_ID,
+          tokenCode: "USDC",
+          // The fixture balances hold a different USDC issuer, so the
+          // default is unheld and needs a trustline.
+          requiresTrustline: true,
+        }),
+      );
+    });
+
+    it("seeds native XLM instead when the swap starts from the default USDC", () => {
+      setSwapStoreState({ destinationToken: null });
+      const route = {
+        key: "swap-amount",
+        name: SWAP_ROUTES.SWAP_AMOUNT_SCREEN,
+        params: { tokenId: MAINNET_USDC_ID, tokenSymbol: "USDC" },
+      } as unknown as Props["route"];
+
+      renderWithProviders(
+        <SwapAmountScreen navigation={makeNavigation()} route={route} />,
+      );
+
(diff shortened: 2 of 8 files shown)


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (3.0.9)
+    CFPropertyList (3.0.8)
```
CassioMG: is this change intentional?

CassioMG:
### Extract the default-destination seeding into a hook

**Suggestion (structure)** — **not a merge blocker**; the behavior is correct and well-tested, this is about keeping the screen maintainable.

**TL;DR:** The new default-receive-token feature lives inline in the swap Amount screen, which pushes the screen well past the ~1k-line ceiling we set when we split its logic into hooks. The seeding logic is fully self-contained — wait for balances to hydrate, seed USDC (or XLM when swapping from USDC), and kick off a one-off security scan when the default isn't held — so it fits the established hook pattern as-is, and extracting it would keep the screen navigable.

---

<details>

**Context:** on #879 the swap screen's logic was extracted into `src/components/screens/SwapScreen/hooks/` specifically to keep `SwapAmountScreen.tsx` at around 1k lines ([comment](https://github.com/stellar/freighter-mobile/pull/879#discussion_r3385881915)). That directory now holds 19 hooks (`useSwapForXlmReserve`, `useSwapSecurityAssessments`, `useSwapDirectionToggle`, …). This PR adds the seeding feature inline instead, growing the screen from 1086 to 1205 lines.

**The block is cleanly extractable.** The whole unit spans lines 407–513: three `useBalancesStore` selectors, the `destinationDescriptorRef` staleness ref + its sync effect, and the `hasSeededDefaultDestination` seeding effect (hydration gate → native-vs-USDC branch → async `scanToken().then()` stamping). None of those symbols are referenced anywhere else in the screen, so the block can move verbatim:

https://github.com/stellar/freighter-mobile/blob/3d91e7e9b39a6fd0f82cdd3a2cb1739117bcfa5c/src/components/screens/SwapScreen/screens/SwapAmountScreen.tsx#L407-L513

**Suggested extraction** — a self-contained effect hook, one-line call site in the screen:

```ts
// src/components/screens/SwapScreen/hooks/useSwapDefaultDestination.ts
useSwapDefaultDestination({
  network,
  publicKey: account?.publicKey,
  swapFromTokenId,
  destinationTokenDescriptor,
  setDestinationToken,
});
```

The hook can read `useBalancesStore` itself (same as `useSwapBalances` does) and own both refs. The new tests in `__tests__/components/screens/SwapScreen/SwapAmountScreen.test.tsx` exercise the behavior by rendering the screen, so they should pass unchanged — the refactor is low-risk.

**Related one-liner while you're in there:** the seeded-default scan introduces a third source of `securityLevel` on the descriptor; the JSDoc in `src/components/screens/SwapScreen/helpers/types.ts` (L23–L29) still says it's only populated "when the descriptor comes from a scanned search record" — worth amending to mention the seeded-default path.

</details>

Gemfile.lock
JiahuiWho: Good catch, it's a stale local bundle install snuck that downgrade in. Reverted to main's 3.0.9.

CassioMG (approved):
This LGTM - left a suggestion to [extract the new logic into a hook](https://github.com/stellar/freighter-mobile/pull/986#issuecomment-5484646392) so we keep the Swap screen clean

PR #988, opened 2026-08-26
chore(deps): migrate to @stellar/stellar-sdk 17.0.1

aristidesstaffieri:
### What

Bumps `@stellar/stellar-sdk` from 16.0.0 to 17.0.1 (Protocol 28 XDR via `@stellar/js-xdr` 5) and migrates every call site through both v17 breaking-change guides: the class-based `xdr` namespace (unions switch on `.type` with property arms, enums are singletons, `int64` is `bigint`, named byte aliases are wrappers read via `.toBytes()`) and the `Buffer` -> `Uint8Array` return type change. This is the mobile port of stellar/freighter#2977, including its CAP-85 follow-up commits.

Beyond the mechanical rename/accessor updates, the diff covers: explicit `xdr.encodeBytes` at every site that previously relied on `Buffer.toString("hex"|"base64")` on an SDK result (SEP-53 `signMessage`, SEP-43 `signAuthEntry`, auth-JWT `sub`/`userId`/signature, analytics account-id hash, wasm hash/salt/bytes rendering), since `Uint8Array.toString` silently yields comma-joined decimals; auth-entry network-id comparison via `xdr.Hash.equals` (a wrapper's `equals` returns `false` for raw bytes); signer hashes rendered as the same uppercase hex whether the SDK hands back a hex string (`revokeSponsorship`) or bytes (`setOptions`); `@exodus/bytes` added to the Jest transform allowlist because v17's CJS build `require()`s it and Jest cannot load ESM through its CJS registry.

Protocol 28 / CAP-85: `getInvocationArgs` and the create-contract KeyVal handle `CONTRACT_EXECUTABLE_EXTERNAL_REF` (executable owner + tag, a note that the owner can change the code after signing, deliberately no wasm hash, `address`/`salt` only read from the address preimage arm); `scValByType` decodes `SCV_EXECUTABLE_TAG`; `getInvocationDetails` logs and renders an "unrecognized invocation" warning instead of letting one undecodable invocation throw from a render body and take down the signing view; invalid executable/preimage pairings throw an error naming the pairing.

### Why

v17 is the SDK release that carries the Protocol 28 XDR schema (CAP-71 `SOROBAN_CREDENTIALS_ADDRESS_V2` becomes the default, CAP-83, CAP-85) and Freighter must decode and review those entries on day one. The XDR rebuild also changed behaviors that do not surface as compile errors: `.toString("hex")` on byte results stops matching silently, `scValToNative` returns `Uint8Array` for non-UTF-8 strings, and previously undecodable CAP-85 arms now decode successfully and reach our own `switch` statements, so every one of those paths needed an explicit update and a test rather than relying on the deprecated aliases 17.0.1 restores.

17.0.1 is used over the extension's 17.0.0 because it adds the deprecated `toXDR`-style aliases as a runtime safety net for any third-party code path we do not control and contains no protocol-affecting changes. The lockfile pins the resolution, so the 7-day `npmMinimalAgeGate` in `.yarnrc.yml` does not affect `yarn install --immutable`.

### Known limitations

`yarn lint:check` reports the same pre-existing `import/order` conflict between ESLint and the Prettier sort-imports plugin for files importing Node builtins alongside `@stellar/stellar-sdk` (`parseTransaction.ts` on `main`, and the two `services/auth` files here); Prettier's order is what the pre-commit hook enforces, and `lint:check` is not part of CI. The sign-transaction and auth-entry review screens were verified via unit tests (including new CAP-85 fixtures) rather than a live dApp round-trip on device. `xdr.ScAddress.scAddressTypeContract(rawBytes)` in 17.0.1 produces an address whose `contractId` fails to encode (wrapping in `new xdr.ContractId(...)` works); only a test fixture was affected, but it is worth an upstream report.

#### PR structure

#### Testing

#### Release

#### Verification

- [x] `yarn tsc --noEmit` clean
- [x] `yarn jest --ci`: 223/223 suites, 2984 tests green
- [x] `yarn install --immutable` passes from the updated lockfile without touching `.yarnrc.yml`
- [x] Metro production bundle builds (`react-native bundle --platform ios --dev false`), confirming v17's ESM deps (`@exodus/bytes`, `uint8array-extras`, `smol-toml`) resolve under Metro
- [x] Dev build boots on Hermes (iPhone 17 simulator) with live Horizon balances and no JS errors
- [ ] CI green

### DIFF
--- a/docs/walletconnect-rpc-methods.md
+++ b/docs/walletconnect-rpc-methods.md
@@ -280,26 +280,27 @@ specific contract invocation without submitting a transaction.
 **Building the `entryXdr` on your dApp**
 
 ```typescript
-import { xdr, Networks, hash, Keypair } from "@stellar/stellar-sdk";
+import {
+  buildAuthorizationEntryPreimage,
+  Networks,
+} from "@stellar/stellar-sdk";
 
 // Obtain the SorobanAuthorizationEntry from your contract simulation
 const simulationResult = await server.simulateTransaction(tx);
 const authEntry = simulationResult.result.auth[0]; // SorobanAuthorizationEntry
 
-// Build the HashIdPreimage — this is what the wallet will hash and sign
-const preimage = xdr.HashIdPreimage.envelopeTypeSorobanAuthorization(
-  new xdr.HashIdPreimageSorobanAuthorization({
-    networkId: hash(Buffer.from(Networks.PUBLIC)),
-    nonce: authEntry.credentials().address().nonce(),
-    signatureExpirationLedger: authEntry
-      .credentials()
-      .address()
-      .signatureExpirationLedger(),
-    invocation: authEntry.rootInvocation(),
-  }),
+// Build the HashIdPreimage — this is what the wallet will hash and sign.
+// As of stellar-sdk v17 simulation records CAP-71 ADDRESS_V2 credentials by
+// default, whose preimage is the address-bound
+// `envelopeTypeSorobanAuthorizationWithAddress` arm; let the SDK pick the
+// right arm off the entry instead of hand-rolling the legacy one.
+const preimage = buildAuthorizationEntryPreimage(
+  authEntry,
+  validUntilLedgerSeq,
+  Networks.PUBLIC,
 );
 
-const entryXdr = preimage.toXDR("base64");
+const entryXdr = preimage.toXdr("base64");
 
 // Send via WalletConnect
 const result = await walletKit.request({
--- a/docs/walletconnect-rpc-methods.md
+++ b/docs/walletconnect-rpc-methods.md
@@ -289,6 +289,10 @@ import {
 const simulationResult = await server.simulateTransaction(tx);
 const authEntry = simulationResult.result.auth[0]; // SorobanAuthorizationEntry
 
+// The signature is only valid until this ledger; give the user time to review.
+const { sequence } = await server.getLatestLedger();
+const validUntilLedgerSeq = sequence + 100;
+
 // Build the HashIdPreimage — this is what the wallet will hash and sign.
 // As of stellar-sdk v17 simulation records CAP-71 ADDRESS_V2 credentials by
 // default, whose preimage is the address-bound
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "@shopify/react-native-skia": "2.2.21",
     "@stablelib/base64": "2.0.1",
     "@stablelib/utf8": "2.0.1",
-    "@stellar/stellar-sdk": "16.0.0",
+    "@stellar/stellar-sdk": "17.0.1",
     "@stellar/typescript-wallet-sdk-km": "3.0.0",
     "@tradle/react-native-http": "2.0.0",
     "@walletconnect/core": "2.23.1",
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
     "typescript": "5.8.3"
   },
   "engines": {
-    "node": ">=22"
+    "node": ">=22.12.0"
   },
   "react-native": {
     "crypto": "react-native-crypto",
--- a/src/components/screens/SignTransactionDetails/components/KeyVal.tsx
+++ b/src/components/screens/SignTransactionDetails/components/KeyVal.tsx
@@ -583,7 +583,8 @@ export const ExecutableDetails = ({
             operationKey={t(
               "signTransactionDetails.authorizations.executableTag",
             )}
-            operationValue={externalRef.tag.toString()}
+            // SEP-51 form: reversible for non-UTF-8 bytes, plain text otherwise.
+            operationValue={externalRef.tag.toJson()}
           />
           <ExternalExecutableNote />
         </>
--- a/src/components/screens/SignTransactionDetails/components/KeyVal.tsx
+++ b/src/components/screens/SignTransactionDetails/components/KeyVal.tsx
@@ -527,11 +527,11 @@ export const ExternalExecutableNote = () => (
  * reference, and deliberately no wasm hash, because the owner can change the
  * code the reference resolves to after this transaction is signed.
  */
-export const ExecutableDetails = ({
-  executable,
-}: {
+interface ExecutableDetailsProps {
   executable: xdr.ContractExecutable;
-}) => {
+}
+
+export const ExecutableDetails = ({ executable }: ExecutableDetailsProps) => {
   const { copyToClipboard } = useClipboard();
 
   // v17: `wasmHash` is an xdr.Hash wrapper on the wasm arm only; render it as
--- a/src/components/screens/SignTransactionDetails/components/KeyVal.tsx
+++ b/src/components/screens/SignTransactionDetails/components/KeyVal.tsx
@@ -13,11 +13,16 @@ import {
 } from "@stellar/stellar-sdk";
 import Spinner from "components/Spinner";
 import Avatar from "components/sds/Avatar";
+import { Banner } from "components/sds/Banner";
 import Icon from "components/sds/Icon";
 import { Text } from "components/sds/Typography";
 import { CLAIM_PREDICATES, mapNetworkToNetworkDetails } from "config/constants";
 import { useAuthenticationStore } from "ducks/auth";
-import { getCreateContractArgs, scValByType } from "helpers/soroban";
+import {
+  addressToString,
+  getCreateContractArgs,
+  scValByType,
+} from "helpers/soroban";
 import { truncateAddress } from "helpers/stellar";
 import { useClipboard } from "hooks/useClipboard";
 import useColors from "hooks/useColors";
@@ -36,12 +41,20 @@ type SignerKeyOptions = Extract<
   { type: "revokeSignerSponsorship" }
 >["signer"];
 
-// Signer hash fields (sha256Hash / preAuthTx) parse to a Buffer at runtime but
-// are typed `Buffer | string`; render either as an uppercase hex string.
-const signerKeyToHex = (value: Buffer | string): string =>
-  typeof value === "string"
+/**
+ * Renders a signer hash (sha256Hash / preAuthTx) as uppercase hex.
+ *
+ * Accepts a string as well as bytes because the SDK is inconsistent: the
+ * `revokeSponsorship` signer arm returns these fields as already-hex strings
+ * (`convertXdrSignerKeyToObject`), while `setOptions` returns real
+ * `Uint8Array`s. A string is passed through rather than re-encoded -- encoding
+ * it again would hex the ASCII of the hex.
+ */
+const signerKeyToHex = (value: Uint8Array | string): string =>
+  (typeof value === "string"
     ? value
-    : Buffer.from(value).toString("hex").toUpperCase();
+    : xdr.encodeBytes(value, "hex")
+  ).toUpperCase();
 
 interface KeyValueListItemProps {
   operationKey: string;
@@ -131,7 +144,7 @@ export const KeyValueInvokeHostFnArgs = ({
           </>
         )}
         {args.map((arg, index) => {
-          const xdrString = arg.toXDR().toString();
+          const xdrString = arg.toXdr("base64");
           const contextKey = `${contractId || "no-contract"}-${fnName || "no-fn"}`;
 
           return (
@@ -364,14 +377,14 @@ export const KeyValueClaimants = ({ claimants }: KeyValueClaimantsProps) => {
     predicate,
     hideKey = false,
   }: ClaimPredicateValueProps): React.ReactNode => {
-    switch (predicate.switch().name) {
+    switch (predicate.type) {
       case "claimPredicateUnconditional": {
         return (
           <KeyValueListItem
             operationKey={
               hideKey ? "" : t("signTransactionDetails.operations.predicate")
             }
-            operationValue={CLAIM_PREDICATES[predicate.switch().name]}
+            operationValue={CLAIM_PREDICATES[predicate.type]}
           />
         );
       }
@@ -383,11 +396,11 @@ export const KeyValueClaimants = ({ claimants }: KeyValueClaimantsProps) => {
               operationKey={
                 hideKey ? "" : t("signTransactionDetails.operations.predicate")
               }
-              operationValue={CLAIM_PREDICATES[predicate.switch().name]}
+              operationValue={CLAIM_PREDICATES[predicate.type]}
             />
-            {predicate
-              .andPredicates()
-              .map((p) => claimPredicateValue({ predicate: p, hideKey: true }))}
+            {predicate.andPredicates.map((p) =>
+              claimPredicateValue({ predicate: p, hideKey: true }),
+            )}
           </>
         );
       }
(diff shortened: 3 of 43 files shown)


src/helpers/soroban.ts
```diff
@@ -401,70 +438,112 @@ const isInvocationArg = (
...
+          const externalRefDetails = {
+            type: INVOCATION_TYPE_EXTERNAL_REF,
+            owner: Address.fromScAddress(executableOwner).toString(),
+            tag: tag.toString(),
+          } as FnArgsCreateExternalRef;
+
+          // CAP-85 refs are expected to deploy from an address, but the
+          // preimage arm is independent of the executable arm -- only read it
+          // when it is a shape we can describe.
+          if (preimage.type === "contractIdPreimageFromAddress") {
```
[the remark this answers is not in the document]
aristidesstaffieri: Verified against the SDK: `buildInvocationTree` only accepts `external ref + address` and throws for every other pairing. `getInvocationArgs` now does the same (`creation function appears invalid: an external executable reference is paired with contractIdPreimageFromAsset ...`), so the entry surfaces as an unrecognized invocation instead of a normal contract creation. `address`/`salt` are required on `FnArgsCreateExternalRef` again and the conditional rows in both auth-entry displays are gone. Test added for the invalid pairing.

src/helpers/soroban.ts
```diff
@@ -401,70 +438,112 @@ const isInvocationArg = (
...
         }
 
+        // CAP-85 (Protocol 28): the executable is a reference to a Wasm hash
+        // held by another contract, resolved on-chain at creation time.
+        case "contractExecutableExternalRef": {
+          const { executableOwner, tag } = exec.externalRef;
+          const externalRefDetails = {
+            type: INVOCATION_TYPE_EXTERNAL_REF,
+            owner: Address.fromScAddress(executableOwner).toString(),
+            tag: tag.toString(),
```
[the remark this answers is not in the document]
aristidesstaffieri: Switched `getInvocationArgs` and `scValByType` (`scvExecutableTag`) to `tag.toJson()`. Checked the behavior: identical to `toString()` for valid UTF-8 (`token-v2`), reversible escapes for binary (`\xff\xfe` instead of `��`). Tests added for the binary case in both helpers.

src/components/screens/SignTransactionDetails/components/KeyVal.tsx
```diff
@@ -493,6 +506,92 @@ export const KeyValueClaimants = ({ claimants }: KeyValueClaimantsProps) => {
...
+            operationKey={t(
+              "signTransactionDetails.authorizations.executableOwner",
+            )}
+            operationValue={externalRefOwner}
+          />
+          <KeyValueListItem
+            operationKey={t(
+              "signTransactionDetails.authorizations.executableTag",
+            )}
+            operationValue={externalRef.tag.toString()}
```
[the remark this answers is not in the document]
aristidesstaffieri: `ExecutableDetails` now renders `externalRef.tag.toJson()` as well, so the Operations view matches the auth-entry displays.

docs/walletconnect-rpc-methods.md
```diff
@@ -280,26 +280,27 @@ specific contract invocation without submitting a transaction.
...
+// Build the HashIdPreimage — this is what the wallet will hash and sign.
+// As of stellar-sdk v17 simulation records CAP-71 ADDRESS_V2 credentials by
+// default, whose preimage is the address-bound
+// `envelopeTypeSorobanAuthorizationWithAddress` arm; let the SDK pick the
+// right arm off the entry instead of hand-rolling the legacy one.
+const preimage = buildAuthorizationEntryPreimage(
+  authEntry,
+  validUntilLedgerSeq,
+  Networks.PUBLIC,
 );
```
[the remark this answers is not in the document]
aristidesstaffieri: Defined it in the snippet: `const { sequence } = await server.getLatestLedger(); const validUntilLedgerSeq = sequence + 100;`.

package.json
```diff
@@ -86,7 +86,7 @@
     "@shopify/react-native-skia": "2.2.21",
     "@stablelib/base64": "2.0.1",
     "@stablelib/utf8": "2.0.1",
-    "@stellar/stellar-sdk": "16.0.0",
+    "@stellar/stellar-sdk": "17.0.1",
```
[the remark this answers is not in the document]
aristidesstaffieri: `engines.node` raised to `>=22.12.0` and the README prerequisite updated to `>= 22.12` with the reason. CI workflows pin `node-version: "22"`, which resolves to the latest 22.x, so no workflow change was needed.

src/components/screens/WalletKit/DappAuthEntryDisplay.tsx
```diff
@@ -94,11 +100,16 @@ export const DappAuthEntryDisplay: React.FC<DappAuthEntryDisplayProps> = ({
     if (detail.type === INVOCATION_TYPE_INVOKE)
       return `invoke-${detail.contractId}-${detail.fnName}`;
     if (detail.type === INVOCATION_TYPE_WASM) return `wasm-${detail.hash}`;
+    if (detail.type === INVOCATION_TYPE_EXTERNAL_REF)
+      return `externalRef-${detail.owner}-${detail.tag}`;
+    if (detail.type === INVOCATION_TYPE_UNRECOGNIZED) return detail.type;
```
piyalbasu, line 105: ### nit: duplicate React keys when multiple invocations are unrecognized

**Nit** — not a merge blocker.

**TL;DR:** If one auth entry contains two or more invocations we can't decode, every one of them gets the same list key, so React will warn and could mis-reconcile the rows. The signing-details sibling screen already de-dupes its keys per item. Behavior today is unaffected since expand/collapse is index-based.

<details>

**Root cause:** `getDetailKey` returns the constant `INVOCATION_TYPE_UNRECOGNIZED` for every unrecognized invocation, and the result is used directly as the `key` in `details.map(...)`:

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/WalletKit/DappAuthEntryDisplay.tsx#L104-L106

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/WalletKit/DappAuthEntryDisplay.tsx#L327-L331

The sibling screen appends the item index and is immune:

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/SignTransactionDetails/components/SignTransactionAuthorizations.tsx#L281-L284

**Suggested fix:** match the sibling — `key={`${getDetailKey(detail)}-${index}`}` at the call site.

</details>

src/components/screens/WalletKit/DappAuthEntryDisplay.tsx
```diff
@@ -189,6 +200,78 @@ export const DappAuthEntryDisplay: React.FC<DappAuthEntryDisplayProps> = ({
       );
     }
 
+    // CAP-85 (Protocol 28): contract created from an external executable
```
piyalbasu: ### nit: CAP-85 comment sits above the wrong branch

**Nit** — not a merge blocker.

**TL;DR:** This comment describes contracts created from an external executable reference, but it sits above the unrecognized-invocation branch; the external-ref branch below it — the one it actually describes — is left uncommented. Looks like a copy/paste slip; the sibling screen has it in the right place.

<details>

The comment at L203-L204 precedes the `INVOCATION_TYPE_UNRECOGNIZED` branch (L205), while the `INVOCATION_TYPE_EXTERNAL_REF` branch it describes starts uncommented at L218:

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/WalletKit/DappAuthEntryDisplay.tsx#L202-L206

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/WalletKit/DappAuthEntryDisplay.tsx#L217-L219

Correct placement in the sibling file:

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/SignTransactionDetails/components/SignTransactionAuthorizations.tsx#L180-L184

**Suggested fix:** move the two comment lines down to sit above the `INVOCATION_TYPE_EXTERNAL_REF` branch.

</details>

src/components/screens/SignTransactionDetails/components/KeyVal.tsx
```diff
@@ -493,6 +506,93 @@ export const KeyValueClaimants = ({ claimants }: KeyValueClaimantsProps) => {
...
+);
+
+/**
+ * Renders the executable of a contract-creation host function. CAP-85
+ * (Protocol 28) adds a third arm whose code lives behind a reference into
+ * another contract's storage -- we show the owner and tag that identify the
+ * reference, and deliberately no wasm hash, because the owner can change the
+ * code the reference resolves to after this transaction is signed.
+ */
+export const ExecutableDetails = ({
```
piyalbasu: ### nit: ExecutableDetails types its props inline

**Nit** — not a merge blocker.

**TL;DR:** The new component declares its props inline instead of with a named props type, unlike the other twelve components in this file and the code-style guide's rule.

<details>

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/SignTransactionDetails/components/KeyVal.tsx#L529-L534

[docs/best-practices/code-style.md](https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/docs/best-practices/code-style.md#L190-L193) — "Use a separate `Props` interface (or type) for every component."

**Suggested fix:** extract `interface ExecutableDetailsProps { executable: xdr.ContractExecutable; }` to match the rest of the file.

</details>

src/helpers/soroban.ts
```diff
@@ -401,70 +439,112 @@ const isInvocationArg = (
 export const getInvocationArgs = (
   invocation: xdr.SorobanAuthorizedInvocation,
 ): InvocationArgs | undefined => {
-  const fn = invocation.function();
+  const fn = invocation.function;
```
piyalbasu, line 459: ### nit: getInvocationArgs behavior changed without JSDoc

**Nit** — not a merge blocker.

**TL;DR:** This function gained new validation throws, a new protocol-28 arm, and a wider return union in this PR, but carries no doc comment — the anti-patterns guide asks for JSDoc on any function whose signature or behavior changes, and the other functions this PR touched did get it.

<details>

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/helpers/soroban.ts#L438-L441

[docs/best-practices/anti-patterns.md](https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/docs/best-practices/anti-patterns.md#L67-L71) — "Include JSDoc on every new public function and on any function whose signature or behavior you change."

**Suggested fix:** add a short JSDoc covering the three validation throws, the CAP-85 `externalRef` arm, and the `FnArgsUnrecognized` return member.

</details>

src/helpers/soroban.ts
```diff
@@ -482,9 +562,16 @@ export const getInvocationDetails = (
...
-      invocations.push(args);
+    try {
+      const args = getInvocationArgs(inv);
+      if (args) {
+        invocations.push(args);
+      }
+    } catch (error) {
+      // An invocation we cannot decode must not take down the whole signing
+      // view -- surface it so the user sees that something was unreadable.
+      logger.error("soroban", "Failed to decode authorized invocation", error);
```
piyalbasu, line 597: ### nit: error vs warn for an undecodable dApp invocation

**Nit** — not a merge blocker; flagging so the level is a deliberate choice either way.

**TL;DR:** Failing to decode a dApp-supplied invocation logs at error level here, while the two adjacent call sites that fail to parse dApp-supplied XDR log at warn. The error-handling guide can be read either way (it lists malformed payloads under error), so this may well be intentional.

<details>

This call site:

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/helpers/soroban.ts#L570-L576

The two neighbors that handle the same failure class at `warn`:

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/WalletKit/DappAuthEntryDisplay.tsx#L60-L63

https://github.com/stellar/freighter-mobile/blob/1ae7516ca9e6592a28ddd3aded57965e0bed38a7/src/components/screens/SignTransactionDetails/hooks/useSignTransactionDetails.ts#L99-L103

**Suggested fix:** either downgrade to `logger.warn` for consistency with the neighbors, or leave as-is if undecodable invocations should page-worthy-log — a one-line rationale in the code would settle it.

</details>

piyalbasu (commented):
A few nits from a deeper pass — none blocking, each can be resolved independently.

src/components/screens/WalletKit/DappAuthEntryDisplay.tsx
aristidesstaffieri: Fixed in 0faf1a03: `key={`${getDetailKey(detail)}-${index}`}`, matching the sibling. The exact suggestion trips `react/no-array-index-key`; rather than the file-level disable the sibling uses, I added a targeted `eslint-disable-next-line` with a rationale (list is static per entry, `getDetailKey` can repeat, index disambiguates).

src/components/screens/WalletKit/DappAuthEntryDisplay.tsx
aristidesstaffieri: Fixed in 0faf1a03: moved the two comment lines onto the `INVOCATION_TYPE_EXTERNAL_REF` branch.

src/components/screens/SignTransactionDetails/components/KeyVal.tsx
aristidesstaffieri: Fixed in 0faf1a03: extracted `interface ExecutableDetailsProps { executable: xdr.ContractExecutable; }`.

src/helpers/soroban.ts
aristidesstaffieri: Fixed in 0faf1a03: added a JSDoc covering the `undefined` return, the CAP-85 `externalRef` arm (owner + tag, deliberately no wasm hash), the three preimage-pairing throws plus the unknown-executable throw, and the `FnArgsUnrecognized` substitution in `getInvocationDetails`.

src/helpers/soroban.ts
aristidesstaffieri: Kept `error`, and added the rationale inline in 0faf1a03. The distinction from the two `warn` neighbors: those handle XDR that failed to *parse* (a malformed dApp payload, not our bug). Here the XDR parsed fine and *our* decoder could not interpret it, which means either the network now accepts an executable/preimage arm we do not handle, or a dApp is asking the user to sign a creation the SDK itself considers invalid. Both warrant an engineer looking, and volume is bounded by user-initiated sign requests. Happy to flip to `warn` if you would rather keep Sentry quiet for this class.

piyalbasu (approved):

PR #996, opened 2026-08-31
Add USD volume telemetry for swaps and payments

JakeUrban:
### What

Adds comprehensive USD volume telemetry infrastructure for swap and payment transactions, including:
- USD value derivation for transaction legs with price snapshots
- Asset identity classification (native/classic/Soroban) with SAC-to-classic collapse
- Slippage computation (USD-based and execution-based)
- Failure categorization and reason code extraction
- Integration into swap and payment success/failure event tracking

### Why

This implements the cross-platform USD volume telemetry model (TR-9, TR-48) to track transaction economics at the Amplitude level. The telemetry captures:
- Source and destination leg USD values from one price snapshot (frozen before submission)
- Asset type classification for proper funnel analysis
- Slippage metrics (both USD-denominated and execution-based)
- Failure context (HTTP status, protocol vs. transport errors)

Key design decisions:
- **Price snapshots before submission**: Prices are captured once signing succeeds, immediately before submission, so the figure reflects the market as close as possible to execution
- **SAC collapse**: Classic assets moved via Soroban Asset Contracts are collapsed back to their classic identity for cleaner funnel analysis
- **Graceful degradation**: Missing prices fall back to cached display prices rather than blocking telemetry
- **Pre-submit failure handling**: Signing failures emit their terminal event with no volume data attached — nothing reached the network, so no snapshot is started and there is no attempted volume to report (diverges from TR-71, which emits nothing; see [stellar/freighter#2991](https://github.com/stellar/freighter/pull/2991))

### Known limitations

- SAC-to-classic collapse depends on the balance list being fresh; a SAC-wrapped classic asset the account doesn't hold will be misreported as Soroban-native
- Slippage computation requires both legs to price successfully; partial pricing results in `no_price` status rather than a partial slippage figure

#### PR structure

- [x] This PR does not mix refactoring changes with feature changes
- [x] This PR has reasonably narrow scope (focused on volume telemetry infrastructure)
- [x] This PR includes relevant before and after screenshots/videos highlighting these changes (N/A - backend telemetry)
- [x] I took the time to review my own PR

#### Testing

- [x] These changes have been tested and confirmed to work as intended on Android
- [x] These changes have been tested and confirmed to work as intended on iOS
- [x] These changes have been tested and confirmed to work as intended on small iOS screens
- [x] These changes have been tested and confirmed to work as intended on small Android screens
- [x] I have tried to break these changes while extensively testing them
- [x] This PR adds tests for the new functionality or fixes (comprehensive unit tests for all helpers)

#### Release

- [x] This is not a breaking change
- [x] This PR updates existing JSDocs when applicable
- [x] This PR adds JSDocs to new functionalities (all exported helpers documented)
- [x] I've checked with the product team if we should add metrics to these changes (metrics approved)
- [x] I've shared relevant before and after screenshots/videos highlighting these changes with the design team and they've approved the changes (N/A - backend telemetry)

https://claude.ai/code/session_01AAGXpeUWhCfefxhVBZuMjK


src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -762,43 +783,111 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
         logger.error(
           "TransactionAmountScreen",
           "Transaction submission failed:",
           error,
         );
-
-        analytics.trackTransactionError({
-          error: error instanceof Error ? error.message : String(error),
-          operationType: TransactionOperationType.Payment,
-        });
```
JakeUrban, line 801: Check the extension implementation. We still emit events for pre-submission failure cases, we just don't include volume data.

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -195,15 +228,78 @@ export const useSwapTransaction = ({
...
+        setIsProcessing(false);
         const { error: signingError } = useTransactionBuilderStore.getState();
-        throw new Error(signingError || "Failed to sign transaction");
+        showToast({
+          variant: "error",
+          title: signingError || t("swapScreen.errors.swapTransactionFailed"),
+          toastId: "swap-transaction-failed",
+          duration: 0,
+        });
+        return;
```
JakeUrban: Why are we showing a toast here that wasn't present prior to our changes?

src/helpers/usdVolume.ts
```diff
@@ -0,0 +1,321 @@
...
+ * its only defect for telemetry is the string return type, which this fixes.
+ */
+export const roundHalfUp2dp = (value: BigNumber.Value): number =>
+  new BigNumber(value).decimalPlaces(2, BigNumber.ROUND_HALF_UP).toNumber();
+
+// ---------------------------------------------------------------------------
+// Per-leg USD derivation
+// ---------------------------------------------------------------------------
+
+export type LegUsdStatus = "ok" | "no_price" | "error";
```
JakeUrban: See the extension PR, no union types, use enums instead.

src/services/analytics/transactions.ts
```diff
@@ -12,6 +13,15 @@ import type {
   TransactionErrorEvent,
 } from "services/analytics/types";
 
+/** Flattens an `AssetIdentity` under the given property prefix (e.g. `asset_` -> `asset_issuer`/`asset_type`, `from_` -> `from_asset_issuer`/`from_asset_type`). Issuer omitted for native XLM (TR-48). */
```
JakeUrban: remove all references to specific requirement numbers (ex. TR-48).

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
JakeUrban: Fixed in a348ff9.

You're right — I checked the extension. A signing failure there leaves `signedXDR` as `""`, `submitFreighterTransaction` is still dispatched, it rejects, and the `rejected` branch emits `payment.failed`/`swap.failed`. So the extension does emit a terminal event for a pre-submission failure; this PR had narrowed mobile to emit nothing, which was a behavior change from `main` on top of everything else.

Now on both flows a pre-submit failure emits its terminal event with **no volume data** attached (no `amount`/`amount_usd*`/`failure_category`), since nothing reached the network:

- **Send** (`TransactionAmountScreen.tsx`): the `!signedXDR` guard cancels the snapshot and emits `payment.failed`; the `catch` emits too (it did before this PR — I'd dropped it). The `catch` emit is guarded by an `emittedTerminalEvent` flag so a throw raised *after* the terminal event already fired can't double-count.
- **Swap** (`useSwapTransaction.ts`): reverted to `throw` instead of the early return, so the `catch` stays the single failure path. A new `didSubmit` flag decides whether `volume` gets attached — set just before `submitTransaction`, so every real submit rejection (which throws) still carries volume, and only genuinely pre-submit throws don't.

One judgment call worth flagging: I also moved `asset_code` out of the `if (volume)` gate in `trackTransactionError`, so a pre-submit `payment.failed` still identifies the asset rather than emitting only `reason_code: "unknown"` + `payment_type`. Happy to drop that if you'd rather keep it strictly to the volume-gated shape.

Also added coverage: a send-screen test for the signing-failure emit (via a captured `onConfirm` on the mocked review footer), and updated the swap test that asserted the old "emits nothing" contract.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
JakeUrban: Removed in a348ff9.

It shouldn't have been there. Before this PR, `!signedXDR` did `throw new Error(signingError || "Failed to sign transaction")`, and the `catch` at the bottom of `executeSwap` already showed the error toast. When I swapped the throw for an early `return` (to suppress the terminal event), that toast stopped firing, and I re-added it inline instead of questioning the early return.

Both problems have the same fix, which is the one in the sibling thread: the throw is back, so the `catch` shows the same toast it always did and there's no second toast site.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/helpers/usdVolume.ts
JakeUrban: Done in a348ff9 — converted all five to enums, matching the extension's `usdVolume.ts`/`confirmationPriceSnapshot.ts` exactly:

- `LegUsdStatus` (`OK`/`NO_PRICE`/`ERROR`)
- `AssetKind` (`NATIVE`/`CLASSIC`/`SOROBAN`)
- `FailureCategory` (all ten)
- `PriceSource` / `PriceFreshness` in `confirmationPriceSnapshot.ts`

I also took the extension's `LegUsdResult` shape while I was in there — a discriminated union of `LegUsdOk | LegUsdUnpriced` rather than one interface with optional fields. That's what lets the slippage call site drop its non-null assertions:

```ts
sourceLeg.status === LegUsdStatus.OK && destLeg?.status === LegUsdStatus.OK
  ? computeUsdSlippagePct(sourceLeg.unrounded, destLeg.unrounded)  // was sourceLeg.unrounded!
```

Emitted wire values are unchanged (the enums are string-valued), so the Amplitude property values still match the extension's and the catalog's.

One deliberate omission: the catalog lists `not_observed` as a fourth value for `to_amount_usd_status`, but mobile can't produce it — the settled amount comes off the submit response, which is in hand at the emit site — so I left it out of the enum rather than define a value nothing emits (same as the extension). There's a doc comment on `SwapVolume.toAmountUsdStatus` noting why.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/analytics/transactions.ts
JakeUrban: Done in a348ff9 — all eight removed, not just this one. `grep -rn "TR-[0-9]" src/ __tests__/` is now clean, matching the extension (which has zero).

Where the reference was carrying real information I kept the reasoning and dropped the number — e.g. this one became "Issuer omitted for native XLM", and the swap hook's "ONE price request (TR-9)" became "ONE price request, so they're priced at the same instant". Two were in test names (`useSwapTransaction.test.ts`), so those got reworded too.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
```diff
@@ -762,43 +786,129 @@ const TransactionAmountScreen: React.FC<TransactionAmountScreenProps> = ({
...
+        snapshotHandle = startConfirmationPriceSnapshot({
+          canonicalIds: [sourceCanonicalId],
+          network,
+          useV2: useRemoteConfigStore.getState().use_token_prices_v2,
+          cachedDisplayPrices: {
+            [sourceCanonicalId]: {
+              currentPrice: selectedBalance.currentPrice ?? null,
+            },
+          },
+        });
```
JakeUrban: Why not start the snapshot after signing is successful? Same question applies to the swap case. If it was to fetch prices closer to the user selecting "confirm", its actually preferable that the snapshot is closer to the actual transaction's execution time.

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
JakeUrban: Check the extension to understand which events and properties they emit in the same cases. If there is a meaningful difference between implementations, describe the difference here so I can evaluate before implementing. Otherwise, update this PR to ensure the same events and properties are emitted.

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -276,6 +479,22 @@ export const useSwapTransaction = ({
...
+        // rejection, so no special case is needed beyond emitting here too.
+        // Only swap.failed carries volume, so the pair cannot double-count.
+        analytics.trackTransactionError({
+          error: error instanceof Error ? error.message : String(error),
+          errorCode: reasonCode,
+          isSwap: true,
+          sourceToken: sourceBalance?.tokenCode,
+          destToken: destinationTokenInput?.tokenCode,
+          volume,
+        });
```
JakeUrban, line 543: Check with the extension, does it also have this double-emit pattern? I figured the quote expired event should only be fired before the user confirms and submits the transaction. Report back on what the extension does before proceeding with changes.

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
JakeUrban: Agreed, and done in 0717d3c — moved in both flows.

You're right that "closer to execution" is the better goal. My reason for putting it before signing was extension parity (`useSubmitTxData.tsx` starts the snapshot at line 171 and signs at 199) plus a head-start argument: the fetch's deadline is the transaction's terminal status, so an earlier start shifts a few more events into `confirmation_fetch` rather than `cached_display`. But that head start is tiny — mobile's `signTransaction` is synchronous (`transactionBuilder.ts:474`), single-digit-to-tens of ms against a network round trip in the hundreds — and it buys a freshness-bucket shift at the cost of pricing accuracy, which is the wrong trade.

What moved: the identity classification and `startConfirmationPriceSnapshot` now sit after the signing guard, immediately before `submitTransaction`, in both `TransactionAmountScreen.tsx` and `useSwapTransaction.ts`. Nothing in the requirements pinned it — TR-8 says "on confirmation", and TR-10's "must not block or delay signing" holds either way since the fetch is never awaited.

Bonus, which was your other point: a signing failure now issues no price request at all. The send flow's pre-submit guard drops its `cancel()` entirely (no handle exists yet); the swap flow's catch keeps `cancel()` only for a throw raised after the snapshot started. Added a test asserting the call order is `sign → fetchPrices → submit`, and changed the old "cancels the fetch on signing failure" test to assert no fetch is issued.

**One coupling worth flagging before you decide the sibling thread** (the one asking me to match the extension's pre-submission events): if that resolves toward "emit volume data on pre-submit failures like the extension does", this change has to come back out — a signing failure would then need a snapshot to price with, which means starting it before signing again. The two are mutually exclusive. My read is that they resolve in the same direction (no volume on pre-submit failures → no snapshot needed until signing succeeds), but flagging it so the order of decisions doesn't surprise you.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
JakeUrban: There **is** a meaningful difference, so reporting rather than implementing. Short version: the extension emits the same *event*, but with volume data attached — and I think that's a bug in the extension, not the behavior to copy.

### What the extension actually does on a signing failure

`useSubmitTxData.tsx` doesn't guard on signing at all:

```ts
let signedXDR = transactionSimulation.preparedTransaction ?? "";   // line 198
if (!isHardwareWallet) {
  const res = await reduxDispatch(signFreighterTransaction({...}));
  if (signFreighterTransaction.fulfilled.match(res) && res.payload.signedTransaction) {
    signedXDR = res.payload.signedTransaction;
  }
  // no else — a failed sign just leaves signedXDR as ""
}
const submitResp = await reduxDispatch(submitFreighterTransaction({ publicKey, signedXDR, networkDetails }));
```

`signFreighterTransaction` is a `createAsyncThunk` dispatched without `.unwrap()`, so a rejection doesn't throw — execution falls straight through and submits an empty XDR. That rejects, lands in the `rejected` branch, and emits `payment.failed`/`swap.failed` with the **full** property set: `amount`/`from_amount`, the `amount_usd*` family off the confirmation snapshot, the `asset_*` identity, `reason_code: "unknown"`, and `failure_category`.

### Why I read that as unintended

The comment on that very branch says the opposite:

> `// A pre-submission failure (signing, simulation) never reaches here, since nothing above this point calls submitFreighterTransaction.`

That's wrong for the signing case specifically — `submitFreighterTransaction` *is* called after a failed sign. So the extension's documented intent matches what this PR does; only its actual behavior diverges, via a fall-through.

Two concrete consequences on the extension today:

1. **Attempted-volume inflation.** A local signing failure contributes `amount_usd` to `SUM(amount_usd)` over `*.failed`, even though nothing was ever attempted on the network.
2. **`failure_category: "transport"` on a local failure.** The submit error is a client-side XDR parse failure, so `isDefiniteProtocolAnswer(error)` is false and `getFailureCategory` returns `TRANSPORT`. Per TR-72, `transport` is read as *unresolved — may have settled*. A signing failure definitively never reached the network, so it's the one bucket it must not be in.

### Where mobile sits

Mobile's `submitTransaction` guards explicitly (`transactionBuilder.ts`: `if (!signedTransactionXDR) throw new Error("No signed transaction to submit")`), so mobile can't reproduce the fall-through even if we removed the guard in the screen. Current behavior after a348ff9 — event emitted, no volume data — matches your original instruction, matches mobile's pre-PR behavior, and matches the extension's *stated* intent.

### Options

1. **Keep mobile as-is; open a follow-up to fix the extension.** My recommendation — the events and property *names* already match; the only difference is that the extension over-reports on a path its own comment says is unreachable.
2. **Match the extension exactly** — attach volume + `failure_category: transport` to pre-submit failures on mobile. Cross-surface identical, but bakes in both consequences above, and (see the sibling thread) forces the price snapshot back to before signing, undoing 0717d3c.

Tell me which and I'll implement. I haven't touched anything here pending your call.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
JakeUrban: **Yes, the extension has the same double-emit, and it looks deliberate.** Reporting only — nothing changed.

A submit-time quote expiry fires *both* events in the extension:

1. **`swap.failed`** — `useSubmitTxData.tsx` `rejected` branch, via the ordinary reason-code mapping (`op_under_dest_min`/`op_too_few_offers` → `failure_category: "slippage"`). Its comment calls this "the sole emit point for a quote expiring at submit".
2. **`swap.quote_expired`** — `views/Swap/index.tsx:92`, in an effect gated on `submitStatus === ERROR && isSwapQuoteExpired`. That flag is set in the reducer on the very same rejection (`transactionSubmission.ts:764`, `isSwapQuoteExpired = isQuoteExpiredError(action.payload)`), and this emit carries `result_code`.

Both fire off the one submit rejection, so mobile matches the extension.

**On your intuition that quote-expired should only fire pre-confirm** — the extension has *three* quote-expiry surfaces, and only one is silent:

| Trigger | Emits `swap.quote_expired`? | Where |
| --- | --- | --- |
| In-screen expiry from the simulate hook (pre-submit) | ✅ yes, no `result_code` | `useSwapQuoteExpiry.tsx:61` |
| Submit-time expiry, on the recovery route back to the amount screen | ✅ yes, **with** `result_code` | `views/Swap/index.tsx:92` |
| Submit-time expiry, arriving at the amount screen (`isSwapQuoteExpired` flag) | ❌ toast only, no metric | `useSwapQuoteExpiry.tsx:69` |

That third one is the dedupe you'd expect: the recovery flag re-shows the toast but doesn't re-emit, because `Swap/index.tsx` already emitted for that same expiry.

It also matches the requirements doc. TR-70 says a submit-time expiry "**additionally** emits `swap.failed` with `failure_category: slippage`", and that `swap.quote_expired` "keeps its current name, properties, and **both emit points**". The stated reason: only `swap.failed` carries volume, so the pair can't double-count in a volume total, and `swap.quote_expired` stays the funnel event driving the retry prompt.

**So no change is needed for parity, and making one would diverge from both the extension and TR-70.** The caveat I'd flag: the double emit is real in the funnel sense — one expiry produces two events — so any funnel counting swap failures has to pick one. If you'd still rather drop the submit-time `swap.quote_expired` on mobile, that's a deliberate cross-surface divergence and I'd want TR-70 updated to match. Say the word.

(Apologies — I first posted this on the snapshot-ordering thread by mistake, before I had this thread's comment id. Ignore that copy; this is the one.)

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
JakeUrban: We're going with option 1, fixing in the extension.

JakeUrban:
**CI note: two iOS E2E failures on `0717d3c` that aren't this PR's — re-run queued.**

Failing: `E2E SendClassicToken` and `E2E ForgotPasswordWarning`, both on iOS run [33559942956](https://github.com/stellar/freighter-mobile/actions/runs/33559942956).

Neither reached the feature it's named after. Both died in a shared setup preamble:

- **`ForgotPasswordWarning`** failed inside `../onboarding/CreateWallet.yaml` — after `skip-confirm-button`, the assert on `home-screen|welcome-banner-dismiss-button` timed out. It never got near a forgot-password screen.
- **`SendClassicToken`** failed inside `../shared/SwitchToTestnet.yaml` — tapped `home-account-switcher`, then the assert on `manage-accounts-settings-button` timed out. It never got near a send screen.

Why this isn't the diff:

1. **`E2E CreateWallet` passed in the same run**, while `ForgotPasswordWarning` failed *inside `CreateWallet.yaml`* — the identical shared flow, same commit, same runner. One passed and one failed on the same YAML, which is flakiness by construction.
2. **The two flows that actually exercise the code I changed passed on iOS**: `E2E SwapClassicToken` ✅ and `E2E SendClassicTokenFromDetails` ✅, plus `E2E SendFederatedAddress` ✅. My change only moves the price-snapshot start inside the send/swap confirm handlers — code these failing flows never reached.
3. **`ForgotPasswordWarning` passed on Android** for this same commit ([run 33559942924](https://github.com/stellar/freighter-mobile/actions/runs/33559942924)).
4. Both iOS failures stalled within ~13s of each other (21:30:20 and 21:30:33), consistent with one contended simulator rather than two independent regressions.
5. Both passed on the previous head `a348ff9`, where the identical telemetry code paths were already in place.

Every failure is an app-shell readiness timeout — home screen / account switcher not rendering in time — with no assertion touching telemetry, pricing, or transaction submission.

I've re-run the failed jobs once. If they fail again I'll treat it as real and root-cause it rather than re-running further.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx
JakeUrban: Got it — option 1. Extension fix is up as a draft: [stellar/freighter#2991](https://github.com/stellar/freighter/pull/2991).

**This PR needs no further change** — mobile's behavior (terminal event, no volume data) is already what we're standardizing on, and `0717d3c` stands since the coupling I flagged resolves in this direction.

One thing I found while writing the fix that I'd missed in the writeup above: **the fall-through has two shapes, not one.**

- Classic payment: `signedXDR` is `""` → submit fails on a client-side XDR parse error → `transport`.
- **Soroban/token transfer: `signedXDR` still holds the *unsigned* prepared XDR, which is truthy** → it gets genuinely submitted and comes back `tx_bad_auth`.

So the obvious guard (`if (!signedXDR)`) would have silently fixed only half of it. #2991 tracks whether signing actually succeeded instead, with a dedicated test for the Soroban path.

I also captured the real payload by running the new tests against current `master` — one local "Incorrect password" failure emits:

```
amount: 100, amount_usd: 50, amount_usd_rate: 0.5, amount_usd_status: "ok",
amount_usd_source: "token_prices_v2", amount_usd_price_freshness: "confirmation_fetch",
asset_type: "native", failure_category: "transport"
```

All five new tests fail against current behavior and pass with the fix (verified by reverting the source file and re-running), so the regression is locked down. Existing extension suites are untouched and green.

#2991 is a draft pending this PR, since the two should land consistently.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/services/backend.ts
```diff
@@ -555,9 +562,17 @@ export const fetchTokenPrices = async ({
         ({ data } = await freighterBackendV2.post<TokenPricesResponse>(
           "/token-prices",
           { tokens: v2Tokens },
-          { params: { network: priceNetwork } },
+          { params: { network: priceNetwork }, signal },
```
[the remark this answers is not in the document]
JakeUrban: Good catch — real, and introduced by this PR. Fixed in `7c616a2`.

Adding `signal` to `fetchTokenPrices` routed aborts through its catch, which called `logApiError` unconditionally. And this isn't a rare path: the confirmation snapshot aborts its in-flight request on **every** `cached_display` fallback — whenever the fetch hasn't resolved by terminal status — so a routine, working flow was stamping a misleading "network unreachable while fetching token prices" warn breadcrumb into Sentry each time.

Fixed with the same `isRequestCanceled` carve-out `getTokenDetails` already makes a hundred lines further down in this file, so the convention is consistent:

```ts
if (!isRequestCanceled(error)) {
  logApiError(/* … */);
}
throw error;
```

Still rethrown either way, so the snapshot's `cached_display` fallback is unaffected — that path depends on the rejection, only not on the logging.

Two tests added: a cancellation logs nothing but still rejects, and a genuine connectivity failure still logs (so the carve-out can't swallow real failures). I verified the first one fails against the previous code — `Expected number of calls: 0, Received number of calls: 1` — rather than just asserting the new behavior passes.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -230,13 +328,81 @@ export const useSwapTransaction = ({
...
+              pathResult?.destinationAmount,
+              settledDestAmount,
+            )
+          : undefined;
+      const usdSlippagePct =
+        sourceLeg.status === LegUsdStatus.OK &&
+        destLeg?.status === LegUsdStatus.OK &&
+        sourceLeg.value !== 0
+          ? computeUsdSlippagePct(sourceLeg.unrounded, destLeg.unrounded)
+          : undefined;
```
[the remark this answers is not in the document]
JakeUrban: Not changing this one, because the rounded-value gate is the specified behavior rather than an oversight — but the observation is correct about what it costs, so flagging it for @JakeUrban rather than just closing it out.

**Why it's written this way.** TR-34 spells the gate out:

> `usd_slippage_pct` is emitted only when **both** legs priced successfully. It is omitted whenever either leg's status is not `ok`, **or when `amount_usd` is zero**.

`amount_usd` is by definition the rounded 2dp figure (TR-19), so "amount_usd is zero" *is* the rounded-value test. And TR-23 accepts the consequence explicitly: a transfer worth less than half a cent contributes `0.00`, with no additional precision emitted to recover it.

The extension gates identically — `useSubmitTxData.tsx`:

```ts
const usdSlippagePct =
  sourceUsd.leg.status === LegUsdStatus.OK &&
  destUsd?.status === LegUsdStatus.OK &&
  sourceUsd.leg.value !== 0
    ? computeUsdSlippagePct(/* … */)
```

So changing it here alone would make mobile emit `usd_slippage_pct` for sub-cent swaps while the extension doesn't — a real divergence in a property that TR-77 requires to mean the same thing on both surfaces, and which is meant to total into one chart.

**That said, the underlying point stands.** `computeUsdSlippagePct` does already guard `sourceUnrounded.isZero()`, the ratio is perfectly computable from the unrounded values, and gating on the rounded figure discards a valid data point for no arithmetic reason. If we want that data, it's a three-part change — both clients plus TR-34 — not a one-line edit here, and it's worth deciding on the merits rather than drifting into it.

@JakeUrban: worth doing, or leave as spec'd? Happy to take it on across both clients and the requirements doc if you want it; otherwise this thread can be resolved as working-as-intended.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/components/screens/SwapScreen/hooks/useSwapTransaction.ts
```diff
@@ -195,21 +233,87 @@ export const useSwapTransaction = ({
...
+
+      snapshotHandle = startConfirmationPriceSnapshot({
+        canonicalIds: [sourceCanonicalId, destCanonicalId],
+        network,
+        useV2: useRemoteConfigStore.getState().use_token_prices_v2,
+        cachedDisplayPrices: {
+          [sourceCanonicalId]: {
+            currentPrice: freshSource.currentPrice ?? null,
+          },
+          [destCanonicalId]: { currentPrice: freshDest.currentPrice ?? null },
```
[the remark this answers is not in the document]
JakeUrban: Confirmed and fixed in 9cad8c6.

Verified the chain: `destinationTokenInput` is `destinationForPath` (`SwapAmountScreen.tsx:238-244`), which is `destinationBalance` when held and `descriptorAsPathBalance(descriptor)` otherwise — and the shim returns only `id`/`tokenCode`/`tokenType`/`token`, no `currentPrice`. The receive card gets its fiat from `computeDestinationFiat`, which falls back to `prices[tokenId]` off `useSwapTokenPrices` exactly when the balance has no `currentPrice`. So `freshDest.currentPrice ?? null` was guaranteed null for that case, and any `cached_display` close (fetch pending or rejected at terminal status) reported the leg unpriced despite a price being on screen.

The fix reads the same map and mirrors `computeDestinationFiat`'s precedence — balance price first, then the store:

```ts
const displayPrices = usePricesStore.getState().pricesByNetwork[network] ?? {};
const displayPriceFor = (balance: PricedBalance, canonicalId: TokenIdentifier) =>
  balance.currentPrice ?? displayPrices[canonicalId]?.currentPrice ?? null;
```

The keys line up without translation: `recordTokenId` and `canonicalIdFromIdentity` both produce `NATIVE_TOKEN_CODE` or `CODE:ISSUER`.

I applied it to the source leg too. That one is always a held balance carrying its own price, so it's a no-op in practice — but leaving the two legs asymmetric would read as an oversight.

Added a regression test that builds a real `pathPaymentStrictSend` result XDR, uses a priceless shim destination with a price only in the store, and forces the confirmation fetch to reject. It fails against the previous code with precisely the symptom you described — `to_amount_usd_status: "no_price"`, no `to_amount_usd`/`to_amount_usd_rate` — and passes with the fix.

Not changed: the send flow. `selectedBalance` there is always a held balance, so it has no equivalent shim path.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/helpers/transactionResult.ts
```diff
@@ -0,0 +1,81 @@
...
+      innerResult.type === "txFeeBumpInnerFailed"
+        ? innerResult.innerResultPair.result.result
+        : innerResult;
+
+    if (
+      innerTxResult.type !== "txSuccess" &&
+      innerTxResult.type !== "txFailed"
+    ) {
+      return null;
+    }
```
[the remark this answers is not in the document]
JakeUrban: Correct — fixed in fdd534f. The helper's own docstring already promised this ("Returns null for anything that isn't a clean success read: the transaction/operation didn't succeed"); the code just didn't honour it.

Now requires `txSuccess`:

```ts
if (innerTxResult.type !== "txSuccess") {
  return null;
}
```

The fee-bump descent is unchanged — a `txFeeBumpInnerFailed` inner result falls out through the same check.

Added the regression case you asked for: a `txFailed` result whose op 0 is a `pathPaymentStrictSendSuccess` for 50000000 stroops and whose op 1 is `paymentUnderfunded`. Against the previous code it returns `"5"` — five whole units of phantom settled volume — and `null` with the fix. The existing `txFailed` test only covered the path payment itself failing, so this gap was real.

Scope note, since it affects how urgent this is: today the helper is only called on the success path, after `submitTransaction` returned a hash, and Horizon answers a failed transaction with HTTP 400 (so `submitTx` throws and we never reach the read). So this is a contract fix rather than a live mis-report — but it is exactly the kind of guard that should not depend on a caller's current control flow, and the failure mode is silent bad volume data rather than an error.

**The extension has the identical check** in `extension/src/helpers/transactionResult.ts` (same two-arm condition), already merged to `master`, so it needs the same one-line fix. Flagging rather than fixing here, same as we did with #2991.

---
_Generated by [Claude Code](https://claude.ai/code)_

CassioMG:
### Partial price response handling differs from the extension

**Consistency hardening (medium)** — cross-platform parity gap in the new volume telemetry. It only triggers when the confirmation-time price response omits one of the requested tokens, but when it does the leg is silently reported as unpriced and mobile diverges from the extension for the same trade. **Not a merge blocker**, though it would be good to land with this PR since it's a small port of what the extension already ships, and schema v3 data starts flowing as soon as this releases.

**TL;DR:** The extension only treats the confirmation price fetch as usable when every requested token actually came back priced; otherwise it falls back to the display cache wholesale. Mobile counts the fetch as a success whenever the request completes, even if the response has no price for one of the legs. In that case the leg is reported as unpriced with "fresh" provenance, and the on-screen price that was captured precisely as the fallback for this situation is never used. Suggesting we port the extension's check so both platforms degrade the same way and the same swap doesn't report a USD value on one and "no price" on the other.

**Steps to reproduce:**
1. Start a swap (or send) for a token whose fiat estimate is visible on screen, so a display price exists.
2. Arrange for the confirmation-time price request to return 200 without an entry for that token (the backend has no current price for it at that moment).
3. Sign and submit, and let the price request finish before the terminal event fires.
4. Inspect the emitted `swap.completed` / `swap.failed` (or `payment.*`) event: the leg's USD status is `no_price` with freshness `confirmation_fetch`. On the extension the same case yields the cached price with freshness `cached_display`.

---

<details>

**Root cause:** `startConfirmationPriceSnapshot` flips `succeeded = true` on any fulfilled promise, and `resolve()` only checks that flag:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/confirmationPriceSnapshot.ts#L77-L99

But `fetchTokenPrices` never rejects on partial coverage. It null-fills every requested id the endpoint omitted (its JSDoc says so as well):

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/services/backend.ts#L668-L675

So a partial 200 becomes a `confirmation_fetch` snapshot whose entry for the missing leg is `{ currentPrice: null }`, which `deriveLegUsd` maps to `LegUsdStatus.NO_PRICE`:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/usdVolume.ts#L59-L65

The `cachedDisplayPrices` the swap hook builds for exactly this case (including the prices-store fallback for a non-held destination) is only consulted when the fetch is pending, rejected, or cancelled, never when it is fulfilled but incomplete:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L300-L327

The extension's merged implementation guards this with an `isComplete` check and falls back wholesale rather than merging per id:

https://github.com/stellar/freighter/blob/5ee1226cc3298f230c7454037ff5e752d3b400c0/extension/src/helpers/confirmationPriceSnapshot.ts#L97-L111

**Deterministic repro:** in `__tests__/helpers/confirmationPriceSnapshot.test.ts`, mock `fetchTokenPrices` to resolve with `{ XLM: { currentPrice: new BigNumber(0.2) }, "USDC:G…": { currentPrice: null } }`, start a snapshot for both ids with `cachedDisplayPrices` holding a non-null USDC price, `await flushMicrotasks()`, then call `resolve()`. Today it returns `freshness: "confirmation_fetch"` with the null USDC entry; expected is `freshness: "cached_display"` with the cached map.

1. *Port the extension guard:* in `resolve()`, replace `if (succeeded)` with
   ```ts
   const isComplete =
     succeeded &&
     canonicalIds.every((id) => fetchedPrices?.[id]?.currentPrice != null);
   if (isComplete) { ... }
   ```
   The mobile map is null-filled, so the check has to look at `currentPrice`, not just key presence. Keep the wholesale fallback (not a per-id merge) so both platforms produce the same freshness for the same response.
2. *Cover it:* add the partial-response test above. The file currently has no case where the fetch fulfills with a null-priced id.

</details>

CassioMG:
### Native check in the telemetry asset classifier should reuse the shared predicate

**Suggestion (low)** — convention alignment with the asset-identity consolidation that landed last week; no live misclassification found with today's callers. **Not a merge blocker.**

**TL;DR:** The new asset classifier decides "native" purely from the absence of an issuer. #999 consolidated every native check in the app into shared predicates and documented which one to use for each input shape; for a raw code/issuer pair, the predicate requires both the native code and no issuer. Suggesting we route this branch through it so the telemetry helper keeps the convention implemented there. The only input that would currently be misclassified is an empty identifier (liquidity-pool shares or an unknown balance), which the send and swap flows can't turn into a transaction, so this is about consistency rather than a data bug today.

---

<details>

**Where:** the first branch of `classifyAssetIdentity` returns `AssetKind.NATIVE` for any falsy issuer, regardless of `code`:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/usdVolume.ts#L192-L204

The file already imports `isNativeContract` / `isNativeToken` from the shared module, and the predicate for this input shape lives next to them:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/assetIdentity.ts#L45-L53

The convention table added by https://github.com/stellar/freighter-mobile/pull/999 maps "a raw code and issuer, with nothing better available" to `isNativeAssetPair(code, issuer)`:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/docs/best-practices/anti-patterns.md#L166-L177

**Reachability:** callers pass `formatTokenIdentifier(getTokenIdentifier(balance))`. `getTokenIdentifier` returns `""` for liquidity-pool shares and unknown shapes, and `formatTokenIdentifier("")` yields `{ tokenCode: "", issuer: "" }`, which the call sites turn into `(code: "", issuer: undefined)` and the classifier labels `native`:

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/balances.ts#L148-L178

https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/balances.ts#L297-L311

Neither flow can build a payment or path payment for those inputs, so nothing reaches the classifier with them today. The ESLint rule from #999 only flags code comparisons, so it doesn't catch a bare truthiness check either, which is why this is worth fixing by hand.

**Suggested fix:**
```ts
import {
  isNativeAssetPair,
  isNativeContract,
  isNativeToken,
} from "helpers/assetIdentity";

if (isNativeAssetPair(code, issuer)) {
  return { code, type: AssetKind.NATIVE };
}
if (!issuer || !isContractId(issuer)) {
  return { code, issuer: issuer || undefined, type: AssetKind.CLASSIC };
}
```
An issuerless non-native code then falls through as classic-without-issuer instead of being pooled into lumen volume (the impersonation case the comment further down already guards against for held balances). Worth two extra `classifyAssetIdentity` cases in `__tests__/helpers/usdVolume.test.ts`: `("XLM", undefined)` is native, `("", undefined)` is not.

</details>

CassioMG:
### Comment and JSDoc text that no longer matches the code

**Nit** — documentation accuracy only, no behavior change. **Not a merge blocker.**

**TL;DR:** Four spots where comments describe behavior the code in this PR no longer has, or were left incomplete: the quote-expiry branch still says it fires its dedicated event *instead of* the generic failure event although it now deliberately emits both; one new comment in that block has a garbled sentence; a test comment claims a second snapshot resolution would "see it as settled" when the implementation guarantees it wouldn't; and the price-fetch function's JSDoc doesn't mention the new cancellation option. Cheap to fix now, and otherwise they'll mislead the next reader.

---

<details>

1. **Stale "instead of SWAP_FAIL" comment.** The pre-existing comment says the quote-expiry branch fires `SWAP_QUOTE_EXPIRED` *instead of* `SWAP_FAIL`, but the block added below it now calls `trackTransactionError({ isSwap: true, … })`, which emits `swap.failed` too (intentionally, as the only volume-bearing event). Update the first comment so it doesn't invite someone to "fix" the pair later.

   https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L501-L529

2. **Garbled sentence** in the new comment: "without this the failure the failure_category exists to measure never reaches a volume-bearing event". Suggest: "without this, the failure that `failure_category` exists to measure never reaches a volume-bearing event."

   https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L515-L521

3. **Test comment describes behavior the implementation prevents.** "Calling resolve() again would now see it as settled" is not what happens: the first `resolve()` aborts the controller, so the `.then` guard discards the late result and a second `resolve()` would still return `cached_display`. The assertion itself is right; reword the comment to what the test proves (the frozen snapshot object is unchanged after the late result lands).

   https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/__tests__/helpers/confirmationPriceSnapshot.test.ts#L171-L185

   https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/helpers/confirmationPriceSnapshot.ts#L77-L102

4. **`fetchTokenPrices` JSDoc not updated for `signal`.** The new field is documented on `FetchTokenPricesParams`, but the function's `@param params` line and `@example` still list only tokens, network, and the v2 flag. Suggest: "Tokens to price, the active network, the v2 flag, and an optional `AbortSignal` to cancel the request."

   https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/services/backend.ts#L510-L522

   https://github.com/stellar/freighter-mobile/blob/fdd534fbd362457526070d8d7616bb696417b3be/src/services/backend.ts#L474-L479

</details>

JakeUrban:
All three addressed in 2192654. Thanks — the first one is a genuine parity bug I'd previously talked myself out of.

**1. Partial price response.** Confirmed and ported. Worth recording why I got this wrong earlier: Copilot raised the same case a few hours ago and I declined it, on the basis that the extension behaved identically so a mobile-only change would *create* divergence. That was wrong — I'd checked the `refs/pull/2984/head` snapshot, and the `isComplete` guard landed in that PR's final merge (`193a8df`), after the ref I had. Reading `origin/master` shows the guard, exactly as you said.

Ported with the one adaptation you flagged: mobile's `fetchTokenPrices` null-fills omitted ids where the extension leaves the key absent, so completeness has to test `currentPrice != null`, not key presence. Kept the wholesale fallback rather than a per-id merge — a per-leg merge would need per-leg freshness, and the catalog deliberately keeps one `amount_usd_price_freshness` / `amount_usd_source` for the whole snapshot so the two legs can't disagree.

One consequence worth naming, since it's a real behavior change beyond the reported case: a genuinely unpriced leg (Soroban token, unlisted asset) is null-filled on *every* response, so those transfers now close as `cached_display` rather than `confirmation_fetch`. The priced leg's figure is unchanged — only its freshness label. That's what the extension already does, so it's parity rather than a new quirk, but it will shift the freshness distribution once schema v3 data starts flowing.

**2. Native check.** Routed through `isNativeAssetPair`, and an issuerless non-native code now falls through as classic.

`canonicalIdFromIdentity` needed the same treatment — it also keyed off issuer presence, so the degenerate identity would still have resolved to `"XLM"` and been *priced* as lumens even with the type corrected. It now keys off the classified type, so such an identity resolves to its own (unpriceable) id. Flagging it since it's one line past what you asked for, but fixing only the type label would have left the pooling you described in place.

Added the two cases you suggested plus one for the canonical id.

**3. Comments.** All four: the "instead of SWAP_FAIL" comment now says "alongside" and points at the block below; the garbled sentence reworded as you had it; the test comment now states what the test proves rather than the impossible alternative; and `fetchTokenPrices`'s `@param`/`@example` cover `signal`.

**Verification.** Both behavioral fixes have tests that fail against the previous code — the snapshot one returns `confirmation_fetch` with the null entry, the classifier one returns `native`. Full suite: 236 suites / 3241 tests passing, `tsc --noEmit` clean, lint clean on the touched files. The one failing suite (`__tests__/scripts/stellar-testnet.test.ts`, `stellar-hd-wallet` wordlists) is pre-existing and fails identically on an unmodified tree.

---
_Generated by [Claude Code](https://claude.ai/code)_

CassioMG:
@JakeUrban thanks for addressing all three comments — and good catch extending the native fix to `canonicalIdFromIdentity`, which is where the pooling would actually have happened. One more relevant thing surfaced while re-reading the terminal-event paths at the new head.

### Terminal-event telemetry reads store fields that a Close-during-submit resets

**Bug (medium)** — data-integrity gap in the new volume telemetry, reachable through a flow the UI explicitly invites (the processing screens tell the user they can close while the submission finishes). **Not a merge blocker**, but worth fixing before schema v3 events start flowing, since the mislabels land in exactly the buckets this telemetry exists to measure.

**TL;DR:** Closing the processing screen while the submit is still in flight resets the transaction store. The store's guard against stale responses then (correctly) refuses to write the outcome of that submit when Horizon answers. But the terminal event is built by reading those store fields after the await, so in this flow a successful swap reports its destination leg as a derivation error with no settled amount or slippage, and a rejected swap or payment reports an unknown reason with a "transport" failure category, even though Horizon's response carried the settled amount or the verdict. A quote-expiry rejection also loses its dedicated event. Suggest handing the submit outcome back to the awaiting caller (the way the extension's emit site consumes the submit result) instead of reading it from the store afterwards.

**Steps to reproduce:**
1. Start a swap (or send) and confirm. While the processing screen still shows the in-progress state, tap Close.
2. Let Horizon respond a few seconds later, either a success or a rejection such as `op_under_dest_min`.
3. Inspect the emitted event. Success: `swap.completed` carries `to_amount_usd_status: error` and no `to_amount` or slippage figures. Failure: `swap.failed` / `payment.failed` carries `reason_code: unknown` and `failure_category: transport`, and no `swap.quote_expired` fires for a quote-expiry rejection.

---

<details>

**Root cause:** `submitTransaction` writes the PR-added `submitResultXdr` (success) and `submitErrorResultCodes` / `submitErrorHttpStatus` / `submitErrorIsProtocolAnswer` (failure) only when the store's `requestId` still matches the attempt, but returns the hash / `null` regardless:

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/ducks/transactionBuilder.ts#L558-L568

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/ducks/transactionBuilder.ts#L640-L655

The guard predates this PR and is right for the UI (a late response must not repaint a newer transaction). The problem is that the new emit sites read the outcome from the store after the await instead of from the attempt itself.

**How the store gets reset mid-flight:** both processing screens render Close in their non-terminal state:

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SwapScreen/screens/SwapProcessingScreen.tsx#L259-L273

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SendScreen/screens/TransactionProcessingScreen.tsx#L304-L318

The send close handler calls `resetTransaction()` directly; the swap handler navigates away, and the swap screen's unmount effect calls it. `initialState.requestId` is `null`, so the guard fails for the still-pending attempt:

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx#L930-L936

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SwapScreen/screens/SwapAmountScreen.tsx#L714-L725

https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/ducks/transactionBuilder.ts#L128-L133

**Consumers that read the reset store:**

- Swap success: `submitResultXdr` is `null`, so `settledDestAmount` is `null` and the destination leg falls to `LegUsdStatus.ERROR`. The comment above the read says there is "no navigate-away window between submit and reading the response", which holds for the read itself but not for the write, which is the one the guard skips.

  https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L360-L373

  https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L423-L429

- Swap failure: `submitErrorResultCodes` is `null` (so `reasonCode` is `unknown` and the quote-expiry detection never matches) and `submitErrorHttpStatus` / `submitErrorIsProtocolAnswer` read as `null` / `false`, so `getFailureCategory` returns `transport`.

  https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L341-L353

  https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SwapScreen/hooks/useSwapTransaction.ts#L483-L497

- Send failure: same reads, same result.

  https://github.com/stellar/freighter-mobile/blob/21926542581831d3814f23d46fbe17944fcdad24/src/components/screens/SendScreen/screens/TransactionAmountScreen.tsx#L866-L877

**Extension reference:** the emit site consumes the submit thunk's own result (`submitResp.payload.result_xdr` on success, `submitResp.payload` on rejection) rather than store state, so a navigate-away can't strip it:

https://github.com/stellar/freighter/blob/5ee1226cc3298f230c7454037ff5e752d3b400c0/extension/src/popup/components/InternalTransaction/SubmitTransaction/hooks/useSubmitTxData.tsx#L288-L300

https://github.com/stellar/freighter/blob/5ee1226cc3298f230c7454037ff5e752d3b400c0/extension/src/popup/components/InternalTransaction/SubmitTransaction/hooks/useSubmitTxData.tsx#L340-L347

**Deterministic repro:** in `__tests__/components/screens/SwapScreen/useSwapTransaction.test.ts`, mock `submitTx` (or the store's `submitTransaction`) with a deferred promise, call `useTransactionBuilderStore.getState().resetTransaction()` while it is pending, then resolve it with a hash and a `result_xdr` containing a settled `pathPaymentStrictSend`. `trackSwapSuccess` is called with `to_amount_usd_status: "error"` and no `to_amount`. Same shape for the failure path with a Horizon 400 carrying `op_under_dest_min`: `swap.failed` gets `failure_category: "transport"` and `swap.quote_expired` is never tracked.

1. *Return the attempt's outcome:* have `submitTransaction` resolve to a per-attempt value, e.g. `{ hash, resultXdr }` on success and `{ hash: null, error, resultCodes, httpStatus, isProtocolAnswer }` on failure (or a typed thrown error carrying them), and have both emit sites build the volume payload from that value. Keep the guarded store writes as they are for the UI. The two callers and their tests are the only consumers of the return value today.
2. *Tidy the comment:* once the read comes from the returned value, drop or reword the "no navigate-away window" note above the settled-amount read so it doesn't imply the store read is safe.

</details>

JakeUrban:
Fixed in 97e21d9, taking suggestion 1 — plus one more instance of the same bug class you didn't hit.

**The fix.** `submitTransaction` now resolves with a `SubmitTransactionOutcome` — `hash` + `resultXdr` on success, `error` / `resultCodes` / `httpStatus` / `isProtocolAnswer` on failure — and both emit sites build the event from that value. The guarded store writes are untouched: they serve the UI, which genuinely shouldn't be repainted by a superseded attempt. The failure details ride to the catch on the thrown `submitFailure`, so the quote-expiry detection reads this attempt's codes too.

**The extra one: `allowedSlippage`.** The swap screen's unmount cleanup also calls `resetToDefaults()` on `useSwapSettingsStore` (`SwapAmountScreen.tsx:720`), and `freshSwapSlippage` was read *after* the await — so a Close made `swap.completed` report the default tolerance rather than the one the swap was built with. Same shape, different store, so scoping the fix to `submitTransaction`'s return value wouldn't have caught it. Now read before the await.

**Scope check.** The source-leg figures were never at risk: `amount` / `from_amount`, `amount_usd` and `to_amount_quoted` all come from closure values and the frozen snapshot, none of which the reset touches. So this recovers the destination leg, both slippage figures, and the failure classification.

**One residual, unfixable in this shape.** All of the above still depends on the async continuation running after the screen unmounts. Nothing cancels it — `submitTx` takes no abort signal and neither close handler touches the in-flight promise — so it does. But if the OS backgrounds or kills the app before Horizon answers, no terminal event is emitted at all. That's the mobile analogue of the whole-event loss the requirements already accept for the extension popup, and it would need a background emitter rather than a change here.

**Tests.** Two regression cases under "Close during submit (store reset mid-flight)", one per outcome, with the builder store mocked empty to represent the reset. Verified both fail against the store-reading version — the success one reports `to_amount_usd_status: "error"` with no `to_amount`, the failure one `transport`. The success case also asserts `allowedSlippage: "0.5"` against the mocked non-default.

Also updated: `SendCollectibleReview` for the new return type (no volume data involved), the swap hook's now-inaccurate "no navigate-away window" comment, and the store tests that asserted a bare hash.

Full suite 236 suites / 3243 tests passing, `tsc --noEmit` clean, lint clean on the touched files; the one failing suite is the pre-existing `stellar-hd-wallet` one.

---
_Generated by [Claude Code](https://claude.ai/code)_

CassioMG:
### Store fields left write-only after the refactor

**Nit** — dead state only, no behavior change. **Not a merge blocker.**

**TL;DR:** Three of the store fields this PR added were there for the telemetry to read. Now that the terminal events build from the returned submit outcome instead, nothing in the app reads them, though they are still written on both submit paths and their doc comments still name telemetry as the reader. Worth either dropping them or rewording the docs to say why they stay.

---

<details>

`submitResultXdr`, `submitErrorHttpStatus` and `submitErrorIsProtocolAnswer` are written on the success and failure paths of `submitTransaction`, but `git grep` finds no reader in `src/` — the only remaining references are the assertions in `__tests__/ducks/transactionBuilder.test.ts`. Their docstrings still describe the consumer that moved to `SubmitTransactionOutcome`:

https://github.com/stellar/freighter-mobile/blob/97e21d97d9b0e7f05a8602b77846e864d73ef66d/src/ducks/transactionBuilder.ts#L87-L99

Two ways to close it:

1. *Drop them* — remove the three fields along with their writes, their entries in `initialState`, and the assertions that cover them. The guarded writes stay meaningful for `transactionHash` and `error`, which the processing screens do read.
2. *Keep them* — reword the three docstrings so they no longer point at telemetry, and say what does justify them (debugging, or a reader you expect to add).

Note `error` and `submitErrorResultCodes` are not in this group: the processing screens read `error`, and `SendCollectibleReview.tsx` still reads `submitErrorResultCodes` for its failure telemetry.

</details>

CassioMG (approved):
LGTM - left a last NIT comment about removing a dead state

JakeUrban:
Took option 1 — dropped, in 1f059d1. Confirmed all three were write-only: `git grep` finds no reader in `src/` for `submitResultXdr`, `submitErrorHttpStatus` or `submitErrorIsProtocolAnswer`. Since this PR added them, removing them leaves the store no larger than before.

Gone: the three fields, their writes on both submit paths, their `initialState` entries, and their docstrings. `transactionHash`, `error` and `submitErrorResultCodes` stay untouched, exactly as you scoped it.

Two things worth noting about how the removal landed:

The **locals** behind the last two remain in the failure path — they feed the outcome's `httpStatus` / `isProtocolAnswer`, which is what `getFailureCategory` reads now. So rather than delete the store tests that covered them, I repointed them at the returned outcome. The four Horizon-error-split cases keep asserting the same 400/429/504/non-Horizon classification inputs, just off `outcome` instead of `store.getState()`.

I also reworded `SubmitTransactionOutcome`'s own docstring, which had the same staleness in reverse — it described the store as holding "the equivalent fields", which is no longer true for three of them. It now says the store keeps what the UI renders and why the attempt's own outcome exists alongside it.

Two swap-hook test fixtures were still mocking the removed fields on `mockGetBuilderState`; dropped from both, since mocking state the hook no longer consults just misleads the next reader.

**One thing I noticed but did not touch**, since it's outside this PR: `SendCollectibleReview.tsx:414` reads `error` and `submitErrorResultCodes` from the store after its own await, so `collectible_send.failed` has the same Close-during-submit gap — its `reason_code` degrades to `unknown`. That predates this PR (it read the store the same way before), and collectible sends carry no volume data, so I left it. It's now a two-line change if you want it, since the outcome is already returned — happy to take it here or in a follow-up.

Full suite 236 suites / 3243 tests passing, `tsc --noEmit` clean, lint unchanged (69 pre-existing problems, none in the touched files).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1004, opened 2026-09-10
fix(blockaid): attribute signing balance changes to the signing account

CassioMG:
## TL;DR

The balance-change rows at the top of the dApp signing sheet were taken from Blockaid's `account_summary`, which describes whichever account the scan was requested for. That account is derived from the transaction's source, so when the transaction's source isn't the account being asked to sign, the amounts shown belonged to a different account than the signer.

This switches the rows to the per-account diffs Blockaid keys by address, looked up under the signing account's public key, so the sheet shows what the transaction does to *your* balances. The browser extension already resolves them this way.

Only the WalletConnect/dApp signing path is affected. Send, swap, and collectible flows build their own XDR with the user's account as the source, so their attribution was already correct and their behaviour is unchanged.

<details>

**What changed:**

`getTransactionBalanceChanges` now takes the signing account and resolves diffs under it, instead of reading the summary block:

https://github.com/stellar/freighter-mobile/blob/3eff114f897efc4206badbcde4002d1d5335db09/src/services/blockaid/helper.ts#L572-L616

The mechanism: `freighter-backend`'s `scanTx` sets Blockaid's `account_address` from `tx.source` (or `innerTransaction.source` for fee bumps), and `simulation.account_summary` is Blockaid's projection of `assets_diffs[account_address]`. Since a Stellar operation can carry its own `source`, a transaction can be sourced by account A while an operation debits account B — and B is the account signing. `simulation.assets_diffs` is keyed by account address and carries an entry per account whose balances move, so looking up the signer's key gives the diffs that actually apply to them.

Two behavioural notes:

- `null` (rendered as "Unable to simulate transaction") is now also returned when no `publicKey` is supplied. Returning `[]` there would render "No balance changes detected", which is a stronger claim than the data supports.
- `[]` (rendered as "No balance changes detected") now means "`assets_diffs` has no entry for this account", which is the correct reading for a transaction that genuinely doesn't move the signer's balances — a config-only transaction, for instance.

The hook threads the account through and adds it to the memo deps:

https://github.com/stellar/freighter-mobile/blob/3eff114f897efc4206badbcde4002d1d5335db09/src/hooks/blockaid/useTransactionBalanceListItems.tsx#L37-L60

There is one call site, which already had the active account in scope:

https://github.com/stellar/freighter-mobile/blob/3eff114f897efc4206badbcde4002d1d5335db09/src/components/screens/WalletKit/DappSignTransactionBottomSheetContent.tsx#L71-L75

For reference, the equivalent resolution in the extension is `extension/src/popup/views/SignTransaction/index.tsx` (`assets_diffs?.[publicKey]`) and `useGetSignTxData.tsx` (`diffs[publicKey] || []`).

**Tests:** the existing fixture built `account_summary`; it now builds `assets_diffs` keyed by a signer constant, and the suite resolves through a `balanceChangesFor` wrapper. Five cases added:

- `null` when `publicKey` is undefined, and when it is empty
- `[]` when `assets_diffs` is absent, when it is `{}`, and when it has no entry for the signer
- diffs are attributed to the signer, not the scanned source account — the fixture carries both sides of a transfer plus a contradicting `account_summary`, and asserts the signer's debit wins
- `[]` when only another account's balances change, with a populated `account_summary` present to confirm it is not consulted

**Verification:**

- `yarn jest __tests__/services/blockaid.helper.test.ts` — 38 passed
- `yarn jest` (full suite) — 234 suites passed, 1 skipped; 3183 tests passed, 6 skipped
- `yarn lint:ts` — clean
- `eslint` on the four changed files — clean

**Follow-ups / out of scope:**

- **Simulation union discrimination.** Mobile tests `"error" in simulation` while the extension tests `status === "Success"`. These select the same case under today's schema, since the success arm types `status` as the literal `'Success'` with no sibling members. But if Blockaid adds a third simulation state carrying neither an `error` key nor `status: "Success"`, mobile falls through into the parsing path while the extension bails. It spans three call sites in `helper.ts` (lines 326, 447, 583), two of which drive the malicious/suspicious banner, so tightening it here would have pulled banner logic into this diff. Left for a separate change.
- **`assets_diffs` absent on a successful simulation.** Treated as `[]` here, preserving today's rendering. Whether Blockaid omits the map entirely or returns `{}` for a transaction with no balance movement isn't documented — the field is optional in the schema, and every fixture across the three repos is hand-written. Worth confirming against a real scan of a config-only transaction and a SEP-10 challenge, then pinning with a test.
- **Transaction source account is never displayed.** `buildSummary` collects operation count, fee, sequence, and memo, but not the source. Surfacing it when it differs from the signing account would make this class of mismatch legible directly, rather than only via the per-operation `source` in the collapsed details pane. Note that a source differing from the signer is legitimate and common — a SEP-10 challenge is built exactly that way — so this wants design thought rather than a warning banner.

</details>

### DIFF
--- a/__tests__/services/blockaid.helper.test.ts
+++ b/__tests__/services/blockaid.helper.test.ts
@@ -11,14 +11,23 @@ const CONTRACT_ADDRESS =
   "CAZXRTOKNUQ2JQQF3NCRU7GYMDJNZ2NMQN6IGN4FCT5DWPODMPVEXSND";
 const USDC_ISSUER = "GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN";
 
+const SIGNER = "GDF32CQINROD3E2LMCGZUDVMWTXCJFR5SBYVRJ7WAAIAS3P7DCVWZEFY";
+const OTHER_ACCOUNT =
+  "GBTYAFHGNZSTE4VBWZYAGB3SRGJEPTI5I4Y22KZ4JTVAN56LESB6JZOF";
+
+/** A successful simulation whose `assets_diffs` attributes `diffs` to SIGNER. */
 const makeScanResult = (diffs: unknown[]) =>
   ({
     simulation: {
-      account_summary: {
-        account_assets_diffs: diffs,
+      assets_diffs: {
+        [SIGNER]: diffs,
       },
     },
   }) as any;
+
+/** Every case below resolves diffs for SIGNER unless it says otherwise. */
+const balanceChangesFor = (scanResult: any) =>
+  getTransactionBalanceChanges(scanResult, SIGNER);
 
 describe("getTransactionBalanceChanges", () => {
   it.each([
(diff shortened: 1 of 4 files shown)


CassioMG:
### Code review

No issues found. Checked for bugs and AGENTS.md compliance.

JiahuiWho (approved):
LGTM

PR #1010, opened 2026-09-18
Add breadcrumb-time StrKey scrubbing for native crash paths

### LINKED ISSUE #855: Scrub publicKey from Sentry HTTP auto-breadcrumb URLs
## Context

PR #854 added `publicKey` (and `public_key`) to `PII_FIELDS_LOWER` in [`src/config/logger.ts`](src/config/logger.ts) so the codebase's existing analytics-opt-in gate (in `buildSentryContext` at [`src/config/sentryConfig.ts`](src/config/sentryConfig.ts)) extends to logger payloads — breadcrumb `data.args` and event `extra` fields.

That closes one leak path for opt-out users but **leaves another open**: Sentry React Native's auto-instrumentation creates an HTTP breadcrumb for every fetch / axios request, and many of our endpoints embed the publicKey directly in the URL path:

- Horizon: `https://horizon-testnet.stellar.org/accounts/${publicKey}`
- Horizon: `https://horizon.stellar.org/accounts/${publicKey}/operations`
- Freighter backend: `/account-history/${publicKey}` and similar

These breadcrumbs ship the URL as a string, so they bypass `PII_FIELDS_LOWER` (which only matches object keys, not URL contents) and bypass the `buildSentryContext` analytics gate.

## Impact

For users who have opted out of analytics, the publicKey is now consistently absent from:
- ✅ `Sentry.setContext("appContext", ...)` (gated)
- ✅ `Sentry.setTag("publicKey", ...)` (gated)
- ✅ `logger.error/warn` extras and breadcrumb data (gated, after PR #854)
- ❌ HTTP auto-breadcrumb URLs (NOT gated — still ships)

## Possible approaches

1. **`beforeBreadcrumb` hook** in `Sentry.init` config — intercept HTTP breadcrumbs and rewrite the URL with `${publicKey}` patterns replaced by a placeholder.
2. **Custom `httpClientIntegration` configuration** — Sentry's SDK exposes integration options for URL scrubbing.
3. **Skip HTTP breadcrumbs entirely** when analytics is disabled (heavy-handed but simple).

Approach (1) seems most surgical: a regex replace targeting Stellar's G-prefixed addresses (`/G[A-Z2-7]{55}/`) covers Horizon + freighter-backend without per-endpoint maintenance.

## Acceptance criteria

- [ ] HTTP breadcrumbs in Sentry events show `${PUBLIC_KEY}` (or similar placeholder) instead of the raw publicKey for opt-out users.
- [ ] Opt-in users continue to see the unredacted URL (consistent with their preference).
- [ ] Test coverage for the URL-scrubbing logic.

## Related

- PR #854 — context for why the existing opt-out gate exists and how this gap was discovered.

HarshPatel0x07:
Closes #855

### What

Adds a `beforeBreadcrumb` hook to `Sentry.init()` that scrubs Stellar StrKeys from
`breadcrumb.message` and (recursively) `breadcrumb.data` at creation time, reusing the
existing `scrubStrKeys` / `deepScrubStrKeys` helpers already in `sentryConfig.ts`.

### Why

`beforeSend` already deep-scrubs `event.breadcrumbs[].data`, which covers HTTP
auto-instrumentation URLs for anything travelling through the JS layer. Two gaps remain:

1. **Native events never reach `beforeSend`.** The existing comment in `beforeSend` says so
   directly — native crashes, iOS app hangs and Android ANRs are captured and transmitted by
   the Cocoa/Android SDKs. They attach the breadcrumb buffer *as stored*, so send-time
   scrubbing never runs for them. Several endpoints embed the account in the URL path
   (`/accounts/${publicKey}`, `/account-history/${publicKey}`), so those breadcrumbs sit in
   the buffer unscrubbed until a native crash ships them. Scrubbing on the way in means the
   stored buffer is already clean whichever layer sends it.
2. **`breadcrumb.message` is not covered by `beforeSend`**, which only touches `bc.data`.
   HTTP breadcrumbs carry the URL in `data.url`, but console and navigation breadcrumbs carry
   strings in `message`.

The change is additive — it only adds redaction, and cannot reduce what is scrubbed today.

### Known limitations

- **Not verified on a simulator or device.** Verified via TypeScript, ESLint, Prettier and the
  Jest suite. This change has no UI surface, so there is nothing to screenshot, but I could not
  run the app to observe a real breadcrumb buffer end to end. Happy to adjust if you'd like
  device verification before merge.
- Scope is deliberately limited to scrubbing. It does not change whether breadcrumbs are
  *recorded* while analytics consent is off — `beforeSend` already drops those events.
- `deepScrubStrKeys` keeps its existing `MAX_RECURSIVE_DEPTH` cap and sentinel behaviour.

### Checklist

- [x] Does not mix refactoring with feature changes
- [x] Narrow scope — 2 files, 145 insertions, 0 deletions
- [x] I reviewed my own diff
- [x] Adds tests for the new functionality (8 new tests, all confirmed failing without the fix)
- [x] Not a breaking change
- [x] JSDoc added on the new hook
- [ ] Android / iOS device testing — see Known limitations
- [ ] Screenshots — N/A, no UI surface


HarshPatel0x07:
Thanks — I don't think this changes opt-in behaviour, and I'd like to check my reasoning
against yours.

`beforeSend` already scrubs raw string surfaces unconditionally: `event.message`,
`exception.values[].value`, `event.extra` and `bc.data` all go through `scrubStrKeys` /
`deepScrubStrKeys` with no analytics check. The consent gate applies only to
`event.contexts.appContext`.

The existing test spells out why:

> `does NOT mutate event.contexts (opt-in users keep their unredacted publicKey for triage)`

So the file's convention is: interpolated strings are scrubbed for everyone as defence-in-depth,
while the deliberate `appContext.publicKey` is preserved for opt-in users.

`beforeBreadcrumb` follows exactly that split. It touches `breadcrumb.message` and
`breadcrumb.data` and never `event.contexts`, so opt-in users keep the publicKey where triage
expects it. Nothing that was previously visible to opt-in users becomes hidden.

Happy to gate it on `useAnalyticsStore` if you'd prefer breadcrumbs treated differently from
event surfaces — but that would diverge from `beforeSend`, so I've matched the existing pattern.

